PR #3, opened 2018-12-12
Updated CI config, dockerfile

drortirosh:
server's makefile: added "go-get" just to get geth sources
cleaned up dockerfile
config to use the above.


drortirosh:
broken source branch.

PR #5, opened 2018-12-17
Support fallback between realys

alex-forshtat-tbk:
In case the relay that responded to '/getaddr' first later fails to
respond to '/relay' request or returns an invalid transaction,
RelayClient will try all available relays first before giving up.

Note that this may require from client to click "Sign" button multiple
times, as relay address is part of hashed message.

Also, refactoring (in process):
1. Scanning blockcahin for Relays is handled by ServerHelper class
2. Pinging the relays to check availability and address - by ActiveRelayPinger
3. HTTP request is moved to HttpWrapper class

### DIFF
--- a/src/js/relayclient/ServerHelper.js
+++ b/src/js/relayclient/ServerHelper.js
@@ -1,0 +1,168 @@
+
+
+const addPastEvents = require('./addPastEvents')
+
+class ActiveRelayPinger {
+
+    // TODO: 'httpSend' should be on a network layer
+    constructor(filteredRelays, httpSend) {
+        this.filteredRelays = filteredRelays
+        this.remainingRelays = filteredRelays
+        this.httpSend = httpSend
+        this.pingedRelays = 0
+        this.relaysCount = filteredRelays.size
+    }
+
+    /**
+     * Ping those relays that were not returned yet. Remove the retuned relay (first to respond) from {@link remainingRelays}
+     * @returns the first relay to respond to a ping message. Note: will never return the same relay twice.
+     */
+    async nextRelay() {
+        if (this.filteredRelays.length === 0 || this.remainingRelays.length === 0) {
+            return null
+        }
+
+        let firstRelayToRespond = await this.raceToSuccess(
+            this.filteredRelays.map(relay => this.getRelayAddressPing(relay.relayUrl))
+        );
+        this.remainingRelays = this.remainingRelays.filter(a => a.relayUrl !== firstRelayToRespond.relayUrl)
+        this.pingedRelays++
+        return firstRelayToRespond
+    }
+
+    /**
+     * @returns JSON response from the relay server, but adds the requested URL to it:
+     * { relayUrl: url,
+     *   RelayServerAddress: address
+     * }
+     */
+    async getRelayAddressPing(relayUrl) {
+        let self = this
+        return new Promise(function (resolve, reject) {
+            let callback = function (error, body) {
+                if (error) {
+                    reject(error);
+                    return
+                }
+                try {
+                    body.relayUrl = relayUrl
+                    resolve(body);
+                }
+                catch (err) {
+                    reject(err);
+                }
+            }
+            self.httpSend.send(relayUrl + "/getaddr", {}, callback)
+        });
+    }
+
+    /**
+     * From https://stackoverflow.com/a/37235207 (modified to catch exceptions)
+     * Resolves once any promise resolves, ignores the rest, ignores rejections
+     */
+    async raceToSuccess(promises) {
+        let numRejected = 0;
+        return new Promise(
+            (resolve, reject) =>
+                promises.forEach(
+                    promise =>
+                        promise.then((res) =>
+                            resolve(res)
+                        )
+                            .catch(err => {
+                                // console.log(err)
+                                if (++numRejected === promises.length) {
+                                    reject("No response from any server.", err);
+                                }
+                            })
+                )
+        );
+    }
+}
+
+class ServerHelper {
+    /**
+     * 
+     * @param {*} minStake 
+     * @param {*} minDelay 
+     * @param {*} httpSend 
+     */
+    constructor(minStake, minDelay, httpSend) {
+        this.minStake = minStake
+        this.minDelay = minDelay
+        this.httpSend = httpSend
+
+        this.filteredRelays = []
+        this.isInitialized = false
+    }
+    /**
+     * @param {*} relayHub 
+     * @param {*} relayHubAddress 
+     */
+    setHub(relayHub) {
+        //todo: re-run only on hub
+        this.relayHub = relayHub
+        addPastEvents(this.relayHub)
+        this.relayHubAddress = this.relayHub.address
+    }
+
+    async newActiveRelayPinger() {
+        if (this.relayHub === 'undefined') {
+            throw new Error("Must call to setHub first!")
+        }
+        if (this.filteredRelays.length == 0)
+        {
+            await this.fetchRelaysAdded()
+        }
+        return new ActiveRelayPinger(this.filteredRelays, this.httpSend)
+    }
+
+    /**
+     * Iterates through all RelayAdded and RelayRemoved logs emitted by given hub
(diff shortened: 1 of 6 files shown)


src/js/relayclient/ServerHelper.js
```diff
@@ -0,0 +1,168 @@
+
+
+const addPastEvents = require('./addPastEvents')
+
+class ActiveRelayPinger {
+
+    // TODO: 'httpSend' should be on a network layer
+    constructor(filteredRelays, httpSend) {
+        this.filteredRelays = filteredRelays
```
drortirosh: you modify the referenced list you received.
instead, you should shallow-copy it, using `array.slice()`

src/js/relayclient/ServerHelper.js
```diff
@@ -0,0 +1,168 @@
...
+
+        this.filteredRelays = []
+        this.isInitialized = false
+    }
+    /**
+     * @param {*} relayHub 
+     * @param {*} relayHubAddress 
+     */
+    setHub(relayHub) {
+        //todo: re-run only on hub
```
drortirosh: actually, the comment is misplaced: the `getPastEvents()` result should be cached, and re-run when a new hub is set.
the `addPastEvents()` itself does nothing if called more than once, so its no problem

src/js/relayclient/ServerHelper.js
```diff
@@ -0,0 +1,168 @@
...
+        this.relayHub = relayHub
+        addPastEvents(this.relayHub)
+        this.relayHubAddress = this.relayHub.address
+    }
+
+    async newActiveRelayPinger() {
+        if (this.relayHub === 'undefined') {
```
drortirosh: if ( **typeof** this.relayHub == 'undefined'

src/js/relayclient/ServerHelper.js
```diff
@@ -0,0 +1,168 @@
...
+        return new ActiveRelayPinger(this.filteredRelays, this.httpSend)
+    }
+
+    /**
+     * Iterates through all RelayAdded and RelayRemoved logs emitted by given hub
+     * initializes an array {@link filteredRelays} of relays curently registered on given RelayHub contract
+     */
+    async fetchRelaysAdded() {
+        let activeRelays = {}
+        let addedAndRemovedEvents = await this.relayHub.getPastEvents({ address: this.relayHub.address, fromBlock: 1, topics: [["RelayAdded", "RelayRemoved"]] })
```
drortirosh: don't scan from block 1. That's good for testing only.
estimate start block based on 12-second per block, e.g: `max(1,currentBlock - 24*60*(60/12))`

src/js/relayclient/ServerHelper.js
drortirosh: why 2 lists? I think you're using only one

src/js/relayclient/ServerHelper.js
```diff
@@ -0,0 +1,168 @@
+
+
+const addPastEvents = require('./addPastEvents')
+
+class ActiveRelayPinger {
+
+    // TODO: 'httpSend' should be on a network layer
+    constructor(filteredRelays, httpSend) {
+        this.filteredRelays = filteredRelays
+        this.remainingRelays = filteredRelays
```
drortirosh: also, make a shallow copy, since its a reference, using `Arrays.prototype.slice()`

src/js/relayclient/ServerHelper.js
```diff
@@ -0,0 +1,168 @@
...
+     * Ping those relays that were not returned yet. Remove the retuned relay (first to respond) from {@link remainingRelays}
+     * @returns the first relay to respond to a ping message. Note: will never return the same relay twice.
+     */
+    async nextRelay() {
+        if (this.filteredRelays.length === 0 || this.remainingRelays.length === 0) {
+            return null
+        }
+
+        let firstRelayToRespond = await this.raceToSuccess(
+            this.filteredRelays.map(relay => this.getRelayAddressPing(relay.relayUrl))
```
drortirosh: you ping all relays. instead, slice to use up to 3 on each call

src/js/relayclient/ServerHelper.js
drortirosh: the comment doesn't belong here, but to the `getPastEvents()`: its output should be cached until a hub is set (or: the cache should be cleared, when a new hub is set

src/js/relayclient/ServerHelper.js
drortirosh: if ( **typeof** this.relayHub ===

src/js/relayclient/relayclient.js
```diff
@@ -325,47 +220,46 @@ RelayClient.prototype.balanceOf = async function (target) {
...
-    let relayHubAddress = await promisify(relayRecipient.get_relay_hub.call)()
+  var self = this
+  let relayRecipient = this.RelayRecipient.at(options.to)
 
-    let relayUrl = this.config.relayUrl;
-    let relayAddress = this.config.relayAddress;
+  let relayHubAddress = await promisify(relayRecipient.get_relay_hub.call)()
 
-    if (!relayUrl || !relayAddress) {
+  let relayHub = this.RelayHub.at(relayHubAddress)
```
drortirosh: you use the serverHAlper, without setting the new hub of this contract.

src/js/relayclient/relayclient.js
```diff
@@ -37,11 +38,13 @@ function RelayClient(web3, config) {
     // TODO: require sign() or privKey
     this.config = config || {}
     this.web3 = web3
+    this.httpSend = new HttpWrapper(this.web3)
+    this.serverHelper = this.config.serverHelper || new ServerHelper(this.config.minStake || 0, this.config.minDelay || 0, this.httpSend)
 
     this.RelayRecipient = web3.eth.contract(relayRecipientAbi)
     this.RelayHub = web3.eth.contract(relayHubAbi)
+    this.serverHelper.setHub(this.RelayHub)
```
drortirosh: you have a type mix: `RelayHub` is a class, but inside the serverHelper, its `relayHub` (instance), where you assume there's an `address`: there is an address, but its a static one (probably the initial deployed one), and you can't trust it. That's why we read the hub from the contract.

src/js/relayclient/relayclient.js
```diff
@@ -207,72 +117,57 @@ function parseHexString(str) {
...
+      try {
+        validTransaction = self.validateRelayResponse(
+          body, relayAddress, from, to, encodedFunction,
+          relayFee, gasprice, gaslimit, nonce, relayHubAddress, relayAddress, signature);
+      }
+      catch (error) {
+        console.log(error)
+      }
+
+      if (validTransaction === undefined) {
```
drortirosh: did you check that this is what gets returned (not null, or anything)?

test/clientlib.js
```diff
@@ -79,19 +80,19 @@ contract('RelayClient', function (accounts) {
         // Added, removed, added again - go figure.
         // 2 x will not ping
         await register_new_relay(rhub, 1000, 20, 15, "https://abcd.com", accounts[4]);
-        await rhub.remove_relay_by_owner(accounts[4], {from: accounts[4]});
+        await rhub.remove_relay_by_owner(accounts[4], { from: accounts[4] });
         await register_new_relay(rhub, 1000, 20, 15, "https://abcd.com", accounts[4]);
```
drortirosh: try one server with invalid dns name.

src/js/relayclient/ServerHelper.js
alex-forshtat-tbk: Currently, I do not technically modify this.filteredRelay, but re-assign it, but that is a lucky coincidence. Fixed.

src/js/relayclient/ServerHelper.js
alex-forshtat-tbk: Will actually do:
setHub(relayHub) {
        if (this.relayHub !== relayHub){
            this.filteredRelays = []
        }
...
async newActiveRelayPinger() {
 if (this.filteredRelays.length == 0)
        {
            await this.fetchRelaysAdded() // <--- a.k.a 'getPastEvents'
        }

PR #9, opened 2018-12-25
Penalize flow

shahafn:
Fixing penalize flow issues after changing stake/register mechanism

### DIFF
--- a/contracts/RelayHub.sol
+++ b/contracts/RelayHub.sol
@@ -19,7 +19,6 @@ contract RelayHub is RelayHubApi {
     mapping (address => uint) public nonces;    // Nonces of senders, since their ether address nonce may never change.
 
     struct Relay {
-        address owner;
         uint timestamp;
         uint transaction_fee;
     }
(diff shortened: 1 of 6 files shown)


contracts/RelayHub.sol
```diff
@@ -140,9 +142,10 @@ contract RelayHub is RelayHubApi {
     }
 
     function remove_relay_internal(address relay) internal {
-//        delete relays[relay];
+        delete relays[relay];
```
drortirosh: either delete or set fields to zero, not both
(and if setting fields to zero, set them all, not only timestamp..

contracts/RelayHub.sol
```diff
@@ -252,18 +255,18 @@ contract RelayHub is RelayHubApi {
...
         // compensating the sender with the stake of the relay
         uint amount = stakes[addr1].stake;
-        msg.sender.transfer(amount);
-        emit Penalized(addr1, msg.sender,amount);
-        // TODO: maybe change to unstake_internal
-        delete stakes[addr1];
-        delete relays[addr1];
-        // TODO: move ownership of relay
-//        relays[addr1].owner = msg.sender;
+//        msg.sender.transfer(amount);
```
drortirosh: remove comment

contracts/RelayHub.sol
shahafn: done

PR #13, opened 2018-12-27
Webtools to manage contracts and relays

drortirosh:
contractmanager: check the balance of the contract, and deposit more
funds

relaymanager: ping relay for its status, check its stake and balance,
and update them as needed.


src/js/webtools/utils.js
```diff
@@ -0,0 +1,121 @@
...
+
+    getCookie: function (name, def) {
+        let c = document.cookie.match("\\b" + name + "=([^;]*)")
+        return c ? c[1] : def
+    },
+
+    httpreq: function (url, options) {
+        return new Promise((resolve, reject) => {
+            module.exports.asyncHttpReq(url, options, (e, r) => {
+                if (e) return reject(e)
```
alex-forshtat-tbk: What is the meaning of 'return reject'?

drortirosh:
There's no real return value. It's a way to bail out from the method

PR #15, opened 2018-12-27
refactor RelayRecipient

drortirosh:
renamed `may_relay()` to `accept_relayed_call()`
added `post_relayed_call()`.


contracts/RelayRecipient.sol
```diff
@@ -65,20 +65,29 @@ contract RelayRecipient is RelayRecipientApi {
...
+	 *  @return "0" if the the contract is willing to accept the charges from this sender, for this function call.
+	 *  	any other value is a failure. actual value is for diagnostics only.
+	 *   values below 10 are reserved by can_relay
+	 *  @param relay the relay that attempts to relay this function call.
+	 * 			the contract may restrict some encoded functions to specific known relays.
+	 *  @param from the sender (signer) of this function call.
+	 *  @param encoded_function the encoded function call (without any ethereum signature).
+	 * 			the contract may check the method-id for valid methods
+	 */
+    function accept_relayed_call(address relay, address from, bytes encoded_function ) public view returns(uint32);
```
yoav-tabookey: I just realized that we only pass the encoded function call, not the entire transaction. So accept_relayed_call can't take all factors into account.  E.g. transaction_fee and gas_price.
It's likely that some dapps will want to restrict collect-calls to a certain price limit.  Time-critical apps will accept the charges, but games, for example, might prefer to become unavailable under extreme network conditions and not incur excessive costs.
So shouldn't we include at least gas_price and transaction_fee in the args here?  Anything else that might get factored into the decision?

yoav-tabookey: Just to clarify, the reason to add gas_price is for the view-function-calls.  During the actual transaction, the dapp can look at tx.gasPrice, but it's harder for the relay to know whether a transaction will be compensated this way.  And re transaction_fee, that's required anyway, as it's not part of tx.

contracts/RelayRecipient.sol
```diff
@@ -65,20 +65,29 @@ contract RelayRecipient is RelayRecipientApi {
...
+
+    /**
+     * This method is called after the relayed call.
+     * It may be used to record the transaction (e.g. charge the caller by some contract logic) for this call.
+     * the method is given all parameters of accept_relayed_call, and also the success/failure status and actual used gas.
+     * - success - true if the relayed call succeeded, false if it reverted
+     * - used_gas - gas used by the relayed call. Note that for gas compensation calculation, the gas used by
+     *   the post_relayed_call() is also taken into account.
+     */
+    function post_relayed_call(address /*relay*/ , address /* from */, bytes /* encoded_function */, bool /* success */, uint /* used_gas */ ) external {
```
yoav-tabookey: gas_used is useful, and tx.gasPrice is available at this point (since it's not a view function), but transaction_fee is missing.  If the caller used an expensive relay, the dapp might want to know it and charge the user accordingly.  Let's add transaction_fee as well.

contracts/RelayRecipient.sol
```diff
@@ -65,20 +65,29 @@ contract RelayRecipient is RelayRecipientApi {
...
+	 */
+    function accept_relayed_call(address relay, address from, bytes encoded_function ) public view returns(uint32);
+
+    /**
+     * This method is called after the relayed call.
+     * It may be used to record the transaction (e.g. charge the caller by some contract logic) for this call.
+     * the method is given all parameters of accept_relayed_call, and also the success/failure status and actual used gas.
+     * - success - true if the relayed call succeeded, false if it reverted
+     * - used_gas - gas used by the relayed call. Note that for gas compensation calculation, the gas used by
+     *   the post_relayed_call() is also taken into account.
```
yoav-tabookey: "is also taken into account" is a bit confusing because it sounds like it's taken into account by someone else. So maybe "the recipient's post_relayed_call() should also take the gas used by itself into account"?

contracts/RelayRecipient.sol
drortirosh: right. its missing. I just hope we didn't reach the # of args limit...
note, however, that gas-price is problematic to use.

but anyway, its a different commit.

contracts/RelayRecipient.sol
drortirosh: yes, my wording was bad: I meant, that the **RelayHub** takes post_relayed_call()'s gas usage into account when compensating the relay.
we couldn't care less if post_relayed_call itself takes its own gas usage: I think that dapp will use a simpler-to-comprehend model, like fixed-#-of-tokens-per-request

PR #21, opened 2019-01-02
Gasprice periodic update

shahafn:
Need to better handles delays when testing

### DIFF
--- a/server/src/RelayHttpServer.go
+++ b/server/src/RelayHttpServer.go
@@ -1,19 +1,16 @@
 package main
 
 import (
-	"./librelay"
-	"crypto/ecdsa"
 	"encoding/json"
 	"flag"
 	"fmt"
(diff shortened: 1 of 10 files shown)


server/src/RelayHttpServer.go
```diff
@@ -335,55 +332,40 @@ func prepareServerOnBlockchain() {
...
 
 func keepAlive() {
-	fmt.Println("keepAlive relay...")
-	err := relay.RegisterRelay(common.HexToAddress("0"))
+
+	waitForOwnerActions()
+	when, err := relay.WhenRegistered(relay.HubAddress())
 	if err != nil {
 		log.Println(err)
+	} else if time.Now().Unix()-when < 24*int64(time.Hour/time.Second) { // time.Duration is in nanosec - converting to sec like unix
```
drortirosh: add parameter (in test its going to be ~5sec)

PR #25, opened 2019-01-07
Refactor RelayRecipient - get_recipient_balance, etc

drortirosh:
- added get_recipient_balance() to check balance (and also validate the
existence of the hub)
- call it from `set_relay_hub_internal()`
- make `relay_hub` member private (to require setter and getter)
- added `get_relay_hub()` to return it as RelayHub, and separate
`get_hub_addr()`, which is used from RelayRecipientApi


drortirosh:
Yes, I tested that it gets rejected either if the target contract doesn't
support that method, or if there's no contact at all at that address

PR #34, opened 2019-01-21
support truffle5, web3@1.0

drortirosh:
incomplete (abigen needs some work)

- changed to truffle 5, solc 0.5, web3@1.0
- updated contracts, client code, server, tests.
- replaced hooking mechanism with the more-standard RelayProvider


src/js/relayclient/utils.js
```diff
@@ -105,7 +107,7 @@ module.exports = {
         } catch (e) {
 
             sig_ = await new Promise((resolve, reject) => {
-                web3.eth.sign(account, hash, (err, res) => {
```
alex-forshtat-tbk: change `web3.personal.sign` above to `web3.eth.personal.sign`

PR #44, opened 2019-02-04
Add 'xDai POA Mainnet' support

alex-forshtat-tbk:
Note: actual money is used in this network.
Mnemonic is git-ignored.

### DIFF
--- a/truffle.js
+++ b/truffle.js
@@ -1,8 +1,12 @@
 var HDWalletProvider = require("truffle-hdwallet-provider");
 var mnemonic = "digital unknown jealous mother legal hedgehog save glory december universe spread figure custom found six"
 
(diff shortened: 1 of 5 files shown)


truffle.js
```diff
@@ -1,6 +1,12 @@
 var HDWalletProvider = require("truffle-hdwallet-provider");
 var mnemonic = "digital unknown jealous mother legal hedgehog save glory december universe spread figure custom found six"
 
+const secret_mnemonic_file = "./secret_mnemonic"
+const fs=require('fs')
+let secret_mnemonic
+if (fs.existsSync(secret_mnemonic_file)) {
```
drortirosh: move this into the "xdai_poa_mainnet" block, below, so we don't need it unless we try to contact xdai

PR #46, opened 2019-02-05
Fix server xdai

shahafn:


server/src/librelay/tbk_ethclient.go
```diff
@@ -25,3 +26,14 @@ func (tbkClient *TbkClient) EstimateGas(ctx context.Context, msg ethereum.CallMs
 	return gas,err
 
 }
+
+// SuggestGasPrice retrieves the currently suggested gas price to allow a timely
+// execution of a transaction.
+func (tbkClient *TbkClient) SuggestGasPrice(ctx context.Context) (*big.Int, error) {
+	gasPrice,err := tbkClient.Client.SuggestGasPrice(ctx)
+	if (err == nil && gasPrice.Uint64() == 0) {
+		gasPrice = big.NewInt(params.GWei)
```
drortirosh: Even though its an edge case, you hard-code a price of a transaction as 1 GWei.

PR #50, opened 2019-02-08
GLS-11 Implement withdrawByOwner

alex-forshtat-tbk:
Improve the 'withdraw' example, forbid deposit higher than minimum_stake
Note: Set minimum stake to 0.1 ether! (Was 1 wei). Also, modified the
constants accordingly.

Added two tests:
1. For SampleRecipient's withdraw method
2. For RelayHub's deposit method

Fixed grammar.


contracts/Ownable.sol
```diff
@@ -0,0 +1,72 @@
+pragma solidity ^0.5.0;
```
drortirosh: don't copy in code from somewhere else

contracts/RelayHub.sol
```diff
@@ -70,6 +70,7 @@ contract RelayHub is RelayHubApi {
      * Unused deposited can be withdrawn with `withdraw()`
      */
     function depositFor(address target) public payable {
+        require(msg.value <= minimum_stake, "deposit too big");
```
drortirosh: remove requirement: deposit can't depend on stake

contracts/SampleRecipient.sol
```diff
@@ -15,9 +16,10 @@ contract SampleRecipient is RelayRecipient {
         get_relay_hub().deposit.value(msg.value)();
     }
 
-    function withdraw() public {
+    function withdraw() public onlyOwner {
```
drortirosh: for this `onlyOwner` we copy in contract from openZepellin. I don't like it.

contracts/RelayHub.sol
shahafn: We remove it in a separate commit

contracts/SampleRecipient.sol
shahafn: fixed

PR #51, opened 2019-02-11
GLS-50 In RelayHub, verify that relays are not contracts

alex-forshtat-tbk:
There is no legitimate reason for a relay address to be controlled by a
smart contract within our design of Relay Network.
Making sure that message sender is the transaction origin is the safest
way to validate the caller is externally-owned.

### DIFF
--- a/test/relay_contract.js
+++ b/test/relay_contract.js
@@ -3,6 +3,7 @@
 const Big = require( 'big.js' )
 
 const SampleRecipient = artifacts.require("./SampleRecipient.sol");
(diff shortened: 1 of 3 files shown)


test/relay_contract.js
```diff
@@ -94,7 +95,15 @@ contract("RelayHub", function (accounts) {
         assert.equal(expected_stake, stake[0] );
     })
 
-    it("test_register_relay", async function () {
+    it("Should allow externally owned addresses to register as relays", async function () {
```
drortirosh: `relay must be external address, not a contract`

PR #52, opened 2019-02-11
GLS-49 Add 'RelayMaxNonce' to relay request

alex-forshtat-tbk:
It is possible for a relay to delay the relayed transaction indefinitely
by giving it a very high nonce.
The client's responsibility is then to decide (wisely) what is the
maximum relay nonce he will accept.
Higher nonce tx is be considered invalid. The client keeps trying other
relays. That is expected to trigger a race between the discarded high
nonce transaction and another one, with only one of them able to make it
to the blockchain.

On the server side, getting a 'RelayMaxNonce' lower than known
'lastNonce' leads to immediate error.

### DIFF
--- a/test/clientlib.js
+++ b/test/clientlib.js
@@ -80,7 +80,8 @@ contract('RelayClient', function (accounts) {
         }
         let relay_client_config = {
             relayUrl: localhostOne,
-            relayAddress: relayAddress
+            relayAddress: relayAddress,
+            allowed_relay_nonce_gap: 0
         }
 
         let tbk = new RelayClient(web3, relay_client_config);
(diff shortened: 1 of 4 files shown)


test/clientlib.js
```diff
@@ -102,6 +103,39 @@ contract('RelayClient', function (accounts) {
...
+                jsonRequestData.RelayMaxNonce = 1000000
+            }
+            orig_send.bind(tbk.httpSend)(url, jsonRequestData, callback)
+        }
+        try {
+            await tbk.relayTransaction(encoded, options);
+            assert.fail()
+        }
+        catch(error) {
+            assert.equal(true, error.otherErrors[0].includes("Relay used a tx nonce higher than requested"))
```
drortirosh: doesn't not handle assert.fail good.
probably the fix is 
{{if (error.instanceOf(AssertionError) throw e;}}

PR #59, opened 2019-02-14
Server handle removed

shahafn:


server/src/RelayHttpServer.go
```diff
@@ -20,22 +20,24 @@ import (
...
 )
 
+const VERSION = "0.3.1"
+
 var KeystoreDir = filepath.Join(os.Getenv("PWD"), "build/server/keystore")
 var delayBetweenRegistrations = 24 * int64(time.Hour/time.Second) // time.Duration is in nanosec - converting to sec like unix
 var shortSleep bool                                               // Whether we wait after calls to blockchain or return (almost) immediately. Usually when testing...
 
 var ready = false
+var removed = true
```
drortirosh: why it starts with `removed=true`?

server/src/librelay/relay_server.go
```diff
@@ -331,8 +339,35 @@ func (relay *relayServer) awaitTransactionMined(tx *types.Transaction) (err erro
 	return nil
 }
 
-func (relay *relayServer) UnregisterRelay() error {
-	return relay.Unstake()
+func (relay *relayServer) RemoveRelay(ownerKey *ecdsa.PrivateKey) (err error) {
```
drortirosh: its relay owner function to call, and it calls directly RelayHub - it doesn't call the relay at all.

server/src/librelay/relay_server.go
```diff
@@ -331,8 +339,35 @@ func (relay *relayServer) awaitTransactionMined(tx *types.Transaction) (err erro
...
+	tx, err := relay.sendRemoveTransaction(ownerKey)
+	if err != nil {
+		return err
+	}
+	return relay.awaitTransactionMined(tx)
+}
+
+func (relay *relayServer) sendRemoveTransaction(ownerKey *ecdsa.PrivateKey) (tx *types.Transaction, err error) {
+	auth := bind.NewKeyedTransactor(ownerKey)
+	nonceMutex.Lock()
```
drortirosh: here's an example why the code doesn't belong here: the nonce is of the OWNER, not of the RELAY. no need to lock or increment.
In fact, its strange that it even pass tests with it.

server/src/RelayHttpServer.go
```diff
@@ -351,6 +367,18 @@ func keepAlive() {
...
 }
 
+func shutdownOnRelayRemoved() {
+	removed, err := relay.IsRemoved()
+	if err != nil {
+		log.Println(err)
+		return
+	}
+	if removed {
+		server.Close()
```
drortirosh: better force the `ready`-state to false in case `removed==true`.
This way, we can still query the relay's status.

server/src/RelayHttpServer.go
shahafn: fixed

server/src/RelayHttpServer.go
```diff
@@ -351,6 +367,18 @@ func keepAlive() {
 	log.Println("Done registering")
 }
 
+func shutdownOnRelayRemoved() {
```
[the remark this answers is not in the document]
shahafn: it shuts down the http server, the relay keeps trying to send its balance to the owner.

PR #90, opened 2019-04-08
Fix typo in dockprefix variable

spalladino:
In the docker run script, dockprefix (the base folder where others are later placed) uses DOCKNAME as part of the folder name, but a typo was causing the folder to be named `build/-folders` instead of `build/relaydock-folders`.

### DIFF
--- a/dock/run.sh
+++ b/dock/run.sh
@@ -10,7 +10,7 @@ folders="`pwd` /Users /home"
 #folder_maps="build node_modules"
 
 #absolute prefix, where following folders are placed.
-DOCKPREFIX=~/build/$DOCKANE-folders/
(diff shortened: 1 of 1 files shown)


dock/run.sh
```diff
@@ -10,7 +10,7 @@ folders="`pwd` /Users /home"
 #folder_maps="build node_modules"
 
 #absolute prefix, where following folders are placed.
-DOCKPREFIX=~/build/$DOCKANE-folders/
```
drortirosh: good catch... but the DOCKPREFIX is actually not in use, and should be removed

drortirosh: ok.
let me know what's wrong with the docker build. we got it to work on macs and linux machines here.

PR #98, opened 2019-04-26
Store pending transactions in relay

### LINKED ISSUE #77: relay: persistent transaction management
Relay should:

- Save all sent (pending) transactions.
- Remove transactions as they are mined.
- When a relay process is started, check if there are pending transaction (more than X minutes)
- re-send such pending transaction, with boosted gas price (e.g. eth.gasPrice*1.5)
Rationale: if a transaction blocks the relay (even if it agreed for a given gas price), it blocks the relay from processing new transactions, so it's better to boost it, even if it's on its own expense.

spalladino:
Adds a new `TxStore` component that stores all unconfirmed transactions, with two implementations: an in-memory one and a LevelDB-backed one. Transactions are stored ordered by increasing nonce. Whenever a transactions is sent, it is added to the store.

RelayServer has now a `UpdateUnconfirmedTransactions` method for updating the transactions on the store. This process removes all transactions from the store that have been confirmed (ie mined at least 12 blocks ago). If the transaction with the lowest nonce is pending (ie not mined) after 5 minutes, it is resent with an increased gas price (25% extra, with a max cap), and its timestamp is reset.

This method is now invoked every minute from the RelayHTTPServer.

Fixes #77

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,9 +3,11 @@
 /node_modules/
 /server/src/github.com/ethereum/go-ethereum/
 /metacoin/
-tabookey-gasless-0.1.0.tgz
+tabookey-gasless-*.tgz
 /.vscode/settings.json
 /.idea/
 /server/pkg/darwin_amd64/
+/server/pkg/linux_amd64/
+/server/bin/
 /serverdock/version
 /secret_mnemonic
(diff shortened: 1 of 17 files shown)


shahafn:
I added `netcat` to the docker image since circle failed on not finding it, when you run it in `run-ganache` script. Please rebase and push so we can rerun the tests.

shahafn:
As we said, see #112

PR #111, opened 2019-05-14
RelayHub singleton

nventuro:
Fixes https://github.com/tabookey/tabookey-gasless/issues/93

This adds scripts in `scripts/singleton`, which generate a single contract using [truffle-flattener](https://github.com/nomiclabs/truffle-flattener), compile it, craft a transaction with a synthetic signature (0x1613161316131613...) that deploys it, and output the contract's source code, ABI, and deployment details (the transaction itself, the address that must be funded in order to execute the transaction, and the contract address).

I originally made the script generate a vanity address (e.g. one that starts with 0x1613), but that adds quite a bit of complexity, takes a long while to compute, and feels like a gimmick overall: it creates false confidence in an address that starts with that prefix, when a) it's very easy to generate another contract with that same prefix, so it alone is no guarantee and the full address should be checked, and b) there will probably be multiple versions of RelayHub out there, sharing this prefix.

I also did some minor changes to the contracts themselves: there was a circular dependency between `RelayHub` and `RelayHubRecipient` because they were importing the other contract's implementations, as opposed to the API, which was incomplete: they now depend just on the API, which has some methods that used to be in the impl (though there's some stuff missing still, but I wanted to keep this PR as small as possible).

### DIFF
--- a/scripts/singleton/get-deploy-data.js
+++ b/scripts/singleton/get-deploy-data.js
@@ -1,0 +1,34 @@
+const { lstatSync, readdirSync } = require('fs')
+const { join } = require('path')
+
+const ethTx = require('ethereumjs-tx');
+const ethUtils = require('ethereumjs-util');
+
+function toHexString(buffer) {
+  return '0x' + buffer.toString('hex')
+}
+
+const bin = require(`../../singleton/RelayHub.flattened.json`)
+  .contracts['singleton/RelayHub.flattened.sol:RelayHub']
+  .bin;
+
+const tx = new ethTx({
+  nonce: 0,
+  data: '0x' + bin,
+  value: 0,
+  gasPrice: 100000000000, /// 100 gigawei
+  gasLimit: 8000000,
+  v: 27,
+  r: '0x1613161316131613161316131613161316131613161316131613161316131613',
+  s: '0x1613161316131613161316131613161316131613161316131613161316131613'
+});
+
+const deployer = tx.getSenderAddress();
+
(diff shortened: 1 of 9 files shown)


scripts/singleton/get-deploy-data.js
```diff
@@ -0,0 +1,34 @@
...
+  data: '0x' + bin,
+  value: 0,
+  gasPrice: 100000000000, /// 100 gigawei
+  gasLimit: 8000000,
+  v: 27,
+  r: '0x1613161316131613161316131613161316131613161316131613161316131613',
+  s: '0x1613161316131613161316131613161316131613161316131613161316131613'
+});
+
+const deployer = tx.getSenderAddress();
```
spalladino: Note that this may fail about half of the time, whenever the code of the contract changes, since not all synthetic signatures are valid. We may need to add a loop to either make a change in the contract or use a slightly different signature if it fails.

nventuro: I thought about this, but couldn't make it fail after running it over a thousand times with small changes to the source, so I figured there was something wrong with my understanding of ECDSA.

spalladino: Maybe it's related to choosing `v=27`? Anyway, if it doesn't fail, then it's good!

alex-forshtat-tbk: Googled it. The ECDSA signature represents a point on a curve, so if this value is valid, changing the signed data won't invalidate it.

PR #113, opened 2019-05-21
Refactoring

shahafn:
- Changing snake to camel case
- Adding version to hub
- Changing canRelay ret val from uint32 to uint


contracts/RelayHubApi.sol
```diff
@@ -1,6 +1,6 @@
 pragma solidity >=0.4.0 <0.6.0;
 
-contract  RelayHubApi {
+contract RelayHubApi {
```
drortirosh: better use the interface notation `IRelayHub`

contracts/RelayHubApi.sol
```diff
@@ -21,19 +21,23 @@ contract  RelayHubApi {
...
+    event TransactionRelayed(address indexed relay, address indexed from, address indexed to, uint status, uint charge);
     event Deposited(address src, uint amount);
     event Withdrawn(address dest, uint amount);
     event Penalized(address indexed relay, address sender, uint amount);
 
-    function get_nonce(address from) view external returns (uint);
-    function relay(address from, address to, bytes memory encoded_function, uint transaction_fee, uint gas_price, uint gas_limit, uint nonce, bytes memory approval) public;
+    function getNonce(address from) view external returns (uint);
+
+    function relay(address from, address to, bytes memory encodedFunction, uint transactionFee, uint gasPrice, uint gasLimit, uint nonce, bytes memory approval) public;
```
drortirosh: should be renamed `relayCall`
- its better to have experssive method names, with object-subject
- we use the terminology of acceptRelayedCall, postRelayedCall, so it makes sense to use Call here too.

dock/run.sh
```diff
@@ -34,6 +34,9 @@ for f in $folders ; do test -d $f && FOLDERS+=" -v $f:$f"; done
 FOLDERS+=" -v `pwd`/build/dock-builD:`pwd`/build"
 FOLDERS+=" -v `pwd`/build/dock-node_modules:`pwd`/node_modules"
 mkdir -p build/dock-builD
+mkdir -p build/dock-node_modules
```
drortirosh: what are these for? 
In any case, you have to do `./dock/run.sh npm install` to get all required modules. This copies only few. We _could_ copy a lot of modules (just to save time for npm install) - but must skip all those with native code.

server/src/RelayHttpServer.go
```diff
@@ -19,7 +19,7 @@ import (
 	"time"
 )
 
-const VERSION = "0.3.5"
+const VERSION = "0.3.6"
```
drortirosh: I think its more than a minor change - its a clear API change.

server/src/librelay/relay_server.go
```diff
@@ -473,12 +473,17 @@ func (relay *RelayServer) CreateRelayTransaction(request RelayTransactionRequest
...
+		log.Println(err)
+		return
+	}
 	maxCharge := request.GasPrice.Uint64() * gasEstimate * (100 + relay.GasPricePercent.Uint64()) / 100
 	if toBalance.Uint64() < maxCharge {
 		err = fmt.Errorf("Recipient balance too low: %d, gasEstimate*fee: %d", toBalance, maxCharge)
 		log.Println(err)
 		return
 	}
+	log.Println("Estimated max charge of relyed tx:",maxCharge)
```
drortirosh: relayed, not relyed

server/src/RelayHttpServer.go
shahafn: This one is the the server version, no API change in it.

contracts/RelayHubApi.sol
shahafn: This one was done in a previous pull request... 
It was called  RelayHubApi to be consistent with RelayRecipientApi. I can rename them but not sure it if it matters

drortirosh:
Yes, I meant renaming them both.

dock/run.sh
shahafn: We only need them for openzeppelin & @0x deps we have in the contracts currently

server/src/librelay/relay_server.go
shahafn: done

server/src/RelayHttpServer.go
shahafn: done

contracts/RelayHubApi.sol
shahafn: done

contracts/RelayHubApi.sol
shahafn: done

dock/run.sh
drortirosh: no. if you want to do test or whatever inside the dock, you need other modules too. don't start copy partial modules around..
we must do `./dock/run.sh npm install`.
(the `mkdir` you added above IS required, but the `cp` below are not)

contracts/RelayRecipient.sol
```diff
@@ -3,81 +3,91 @@ pragma solidity >=0.4.0 <0.6.0;
...
         }
-        return orig_msg_data;
+        return origMsgData;
     }
+    /*** Note :values below 10 are reserved by canRelay
+    *  @param encodedFunction the encoded function call (without any ethereum signature).
+    *  @param gasPrice - the gas price for this transaction
+    *  @param transactionFee - the relay compensation (in %) for this transaction
+    */
+    function acceptRelayedCall(address relay, address from, bytes memory encodedFunction, uint gasPrice, uint transactionFee, bytes memory approval) public view returns(uint);
```
drortirosh: fix indent, newline after method (I almost commented that its a comment for the wrong method

dock/run.sh
shahafn: `./dock/run.sh npm install` doesn't work. It probably needs to be run as root but I'm not going to run it with sudo.

contracts/RelayRecipient.sol
shahafn: done

dock/run.sh
drortirosh: it does work, but only if the dock-node_modules folder exists before (the docker uses existing folder, but if it doesn't exists, it creates one as *root*, which breaks the scripts.
that's why I said the mkdir is required, but the rest isn't.
I just did such "docker/run npm install" on a clean environment .

shahafn: Now it works, but because of the mappings of the directories in we give the docker, it installs the modules in `build/dock-node_modules` but updates `package.json` and `package-lock.json` outside of docker which is a big issue.

shahafn: done

PR #116, opened 2019-05-29
Add 'solidity-coverage' node command to the project

alex-forshtat-tbk:
Current 'solidity-coverage' does not support solidity 0.5.0 and up, so
it relies on patcking the 'parser.js' file in the meantime.

Also, the "instrumented" version of the RelayHub contract is too large,
so we need to add a large gas limit to it's migration sctipt, as well as
run the ganache for coverage with the "allowUnlimitedContractSize" flag.

Note: need to fix "skipFiles" behavior as it does not seem to work.
Note 2: Many tests fail in coverage mode, which must be related to a
change in the number of logs/amounts of gas consumed, etc. Needs fixing.


drortirosh:
@alex-forshtat-tbk : I think we should close this one. solidity-coverage is too problematic: by requiring modifying the contracts, it breaks our tests, which are gas-dependent.
sol-coverage seems much less intrusive coverage test (even though that currently  it can only run the unit-tests, and not the actual relay, since it runs the node in-process, and fails to process well connetions from another process (the "go" relay)

drortirosh:
not used: too much code modification and gas manipulation (and thus breaking tests). used `sol-coverage` instead.

PR #128, opened 2019-06-02
Fixes on audit comments.

drortirosh:
following Zepelin's audit added some fixes.

fixed pragma solidity
removed abstract methods from RelayRecipient (as they duplicate IRelayRecipient's)
added documentation of main methods.
cleaned up comments

removed initRelayHub() - to signify setRelayHub should be used (and support updated relay)

### DIFF
--- a/contracts/RelayHub.sol
+++ b/contracts/RelayHub.sol
@@ -1,4 +1,4 @@
-pragma solidity >=0.4.0 <0.6.0;
+pragma solidity ^0.5.5;
 
 import "./IRelayHub.sol";
 import "./IRelayRecipient.sol";
@@ -16,6 +16,7 @@ contract RelayHub is IRelayHub {
     uint constant public minimumUnstakeDelay = 0;
     uint constant public minimumRelayBalance = 0.1 ether;  // can't register/refresh below this amount.
     uint constant public gasReserve = 100000; // how much reserve we actually need, to complete the post-call part of relayCall().
+
     /**
     * the total gas overhead of relayCall(), before the first gasleft() and after the last gasleft().
     * Assume that relay has non-zero balance (costs 15'000 more otherwise).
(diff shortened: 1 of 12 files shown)


contracts/RelayHub.sol
```diff
@@ -143,7 +144,7 @@ contract RelayHub is IRelayHub {
         require(relays[relay].owner == msg.sender, "not owner");
         relays[relay].unstakeTime = relays[relay].unstakeDelay + now;
         // Start the unstake counter
-        if (relays[relay].state != State.PENALIZED) {
+        if (relays[relay].state != State.PENALIZED && relays[relay].state != State.REMOVED) {
```
shahafn: Not sure what the `relays[relay].state != State.REMOVED` check achieves.

drortirosh: Without it, calling remove again will re-trigger the unstake timeout...

shahafn: unstake timeout is set right before this `if`, this branch merely  sets `relays[relay].state = State.REMOVED`.

shahafn: make this a require in the beginning of function

contracts/RelayHub.sol
```diff
@@ -328,7 +328,12 @@ contract RelayHub is IRelayHub {
         require(relays[addr1].stake > 0, "Unstaked relay");
         // Checking that the relay wasn't penalized yet
         require(relays[addr1].state != State.PENALIZED, "Relay already penalized");
-        // compensating the sender with the stake of the relay
+
+        // compensating the sender with HALF the stake of the relay (the other half is burned)
+        uint to_burn = relays[addr1].stake/2;
```
shahafn: 1. Please keep camelCase
2. Use safemath

PR #130, opened 2019-06-05
Adding preRelayedCall functionality

shahafn:


contracts/IRelayRecipient.sol
```diff
@@ -30,6 +32,19 @@ contract IRelayRecipient {
      */
     function acceptRelayedCall(address relay, address from, bytes memory encodedFunction, uint gasPrice, uint transactionFee, bytes memory approval) public view returns (uint);
```
shahafn: @drortirosh  already fixed some comments in the PR #128, so I tried avoiding changing much where he already did . Once @drortirosh 's PR is merged I will rebase and fix some of the comments

contracts/IRelayRecipient.sol
```diff
@@ -30,6 +32,19 @@ contract IRelayRecipient {
...
+     *
+     * @param usedGas - gas used up to this point. The recipient may use this information to perform local booking and
+     *   charge the sender for this call (e.g. in tokens).
+     *   Note that the relay's compensation will also include gas used by preRelayedCall itself.
+     *
+     * Revert in this functions causes a revert of the client's relayed call but not in the entire transaction
+     * (that is, the relay will still get compensated)
+     */
+    function preRelayedCall(address relay, address from, bytes memory encodedFunction, uint usedGas, uint transactionFee) public;
+
```
drortirosh: Can't be given gas used...

contracts/RelayHub.sol
```diff
@@ -238,23 +239,27 @@ contract RelayHub is IRelayHub {
 
     function recipientCallsAtomic(address from, address to, address relayAddr, bytes calldata encodedFunction, uint transactionFee, uint gasLimit, uint initialGas) external returns (bool) {
         // This function can only be called by RelayHub.
-        // In order to Revert the client's relayedCall if postRelayedCall reverts, we wrap them in one function.
+        // In order to Revert the client's relayedCall if preRelayedCall/postRelayedCall reverts, we wrap them in one function.
```
drortirosh: This comment fix it not needed. If preRelayedCall reverts, the relayed call wasn't
Called yet, so there's nothing more to revert

shahafn: Yes, I meant not to run the relayed call in the first place

test/relay_hub_test.js
```diff
@@ -760,6 +761,48 @@ contract("RelayHub", function (accounts) {
 
     });
 
+    it("should revert the 'relayedCall' if 'preRelayedCall' reverts", async function () {
```
drortirosh: TODO: duplicate the test to check for postRelayed reverted

alex-forshtat-tbk: I think this test is actually a copy of a test that checks 'postRelayed', I added this test last week.

contracts/RelayHub.sol
```diff
@@ -251,18 +252,23 @@ contract RelayHub is IRelayHub {
 
         require(msg.sender == address(this), "Only RelayHub should call this function");
 
+        bool successPrePost;
+        bytes memory preRetVal;
+        bytes memory transaction = abi.encodeWithSelector(IRelayRecipient(to).preRelayedCall.selector, relayAddr, from, encodedFunction, (gasOverhead + initialGas - gasleft()), transactionFee);
+        (successPrePost,preRetVal) = to.call.gas(preRelayedCallMaxGas)(transaction);
+        require(successPrePost, "preRelayedCall reverted - Not calling the relayed transaction");
```
alex-forshtat-tbk: A question - would it be possible to make this function return an integer, just as 'acceptRelayCall' does, so that in 'relayCall' we would know which function reverted - 'pre', 'relay' or 'post'? If it is, is there a reason not to do that now, that we are changing it? It will be a small bit of useful debug info on the world of "something went wrong. change something and try again"-based debugging :-)

drortirosh: This is very tricky and adds a lot of complexity. The reason is that we have to revert all three methods (pre-, relayed call, post-) together so it's very hard to do that and still get a return value.

PR #136, opened 2019-06-14
RelayHub refactor

frangio:
We've refactored the critical parts of `RelayHub`: `canRelay`, `relayCall`, and `recipientCallsAtomic`.

The goal was to make the code more straightforward, and easier to read and audit, without changing at all the semantics of the contract.

Since the PR is relatively big, its broken up in small commits that can each be reviewed separately.


alex-forshtat-tbk:
I've created a PR on top of this branch to fix the failed test, please see: 
https://github.com/nventuro/tabookey-gasless/pull/1

contracts/RelayHub.sol
```diff
@@ -255,18 +255,13 @@ contract RelayHub is IRelayHub {
         }
         // Relay transactionFee is in %.  E.g. if transactionFee=40, payment will be 1.4*usedGas.
         uint256 charge = (gasOverhead + initialGas - gasleft()) * gasPrice * (100 + transactionFee) / 100;
-        emitTransactionRelayed(msg.sender, from, to, encodedFunction, uint256(status), charge);
+        emit TransactionRelayed(msg.sender, from, to, abi.decode(encodedFunction, (bytes4)), uint256(status), charge);
```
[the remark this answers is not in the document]
alex-forshtat-tbk: Should still be cheaper than logging the entire operation, IMHO.

frangio:
@alex-forshtat-tbk Do you understand why our PR made the test fail?

alex-forshtat-tbk:
@frangio not really, but it does not seem related to your changes, maybe it's truffle version or something like that. I checked the exception message before your pull request and it was a large stack trace (which gave a false positive), and after it was the exact error message. I thought it didn't require further investigation as a new code is what it was supposed to be - relays will attempt to execute a transaction, get reverted and return an error message to the customer, who will end up "rejected" by all relays. Could have a better test there, I admit.

frangio:
We found a small problem with this refactor.

`abi.decode(encodedFunction, (bytes4))` doesn't seem to be safe. It will revert when `encodedFunction` only contains the selector, as is the case with functions with no arguments. It's not clear to me if this is a bug in Solidity, but we will have to revert back to a custom function for now.

nventuro:
The `abi.decode` issue has been addressed, this should now be ready to be merged.

contracts/RelayHub.sol
```diff
@@ -205,85 +238,168 @@ contract RelayHub is IRelayHub {
...
+        // a sanity check to prevent overflows in case of bugs.msg.sender
+        require(balances[recipient] >= charge, "Should not get here");
+        balances[recipient] -= charge;
         balances[relays[msg.sender].owner] += charge;
-    }
 
-    // Bypassing "stack too deep"... in relayCall()
-    function emitTransactionRelayed(address sender, address from, address to, bytes memory encodedFunction, uint status, uint charge) internal {
-        emit TransactionRelayed(sender, from, to, LibBytes.readBytes4(encodedFunction, 0), status, charge);
+        emit TransactionRelayed(msg.sender, from, recipient, abi.decode(encodedFunction, (bytes4)), uint256(status), charge);
```
drortirosh: We said that `decode( ..bytes4)` is dangerous. So we do need `readBytes4`, and thus re-introduce `emitTransactionRelayed`

alex-forshtat-tbk:
Maybe you could rebase this branch onto the master so we can merge easier? I just added a test for a no parameters transaction.

truffle.js
```diff
@@ -50,7 +50,7 @@ module.exports = {
   },
   compilers: {
     solc: {
-      version: "0.5.5",
+      version: "0.5.9",
```
drortirosh: this change broke our coverage tool.
need to update `coverage-prov.js` to the same version, run `yarn coverage` to see the coverage is running.
(note that if you run coverage and then change the version, you need to `rm -rf .0x-artifacts` before running coverage again)

drortirosh:
@frangio , @nventuro : The branch is finally complete and can be merged. 
Do you think its better to leave all small commits separate, or clean it up?
I don't think it should be all squashed, but still, 18 small commits for a single target is alot...

nventuro:
I think squashing is fine, the separate commits can be viewed here anyway, but there's no need to keep them in the repo's history. I can later rebase the other open PRs on the squash merge commit.

PR #137, opened 2019-06-21
Recipient balance check bugfix

nventuro:
This PR fixes a bug where the recipient's balance is checked after `preRelayedCall` is called, meaning balance draining during that call is not detected. I also added new tests (in a separate test file - future PRs will also use this new test suite) to check for all scenarios.

### DIFF
--- a/test/RelayHub.test.js
+++ b/test/RelayHub.test.js
@@ -1,0 +1,89 @@
+const { BN, ether, expectEvent, time } = require('openzeppelin-test-helpers');
+
+const { getTransactionHash, getTransactionSignature } = require('../src/js/relayclient/utils')
+
+const RelayHub = artifacts.require('RelayHub');
+const SampleRecipient = artifacts.require('SampleRecipient');
+
+contract('RelayHub', function ([_, relayOwner, relay, sender, other]) {  // eslint-disable-line no-unused-vars
+  const RelayCallStatusCodes = {
+    'OK': new BN('0'),
+    'CanRelayFailed': new BN('1'),
+    'RelayedCallFailed': new BN('2'),
+    'PreRelayedFailed': new BN('3'),
+    'PostRelayedFailed': new BN('4'),
+    'RecipientBalanceChanged': new BN('5'),
+  };
+
+  let relayHub;
+  let recipient;
+
+  beforeEach(async function () {
+    relayHub = await RelayHub.new();
+    recipient = await SampleRecipient.new(relayHub.address);
+  });
+
+  context('with staked relay', async function () {
+    const unstakeDelay = time.duration.weeks(4);
+
+    beforeEach(async function () {
+      await relayHub.stake(relay, unstakeDelay, { value: ether('5'), from: relayOwner });
+    });
+
+    context('with registered relay', async function () {
+      const url = 'http://relay.com';
+      const fee = 10; // 10%
+
+      beforeEach(async function () {
+        await relayHub.registerRelay(fee, url, { from: relay });
+      });
+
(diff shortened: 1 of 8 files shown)


drortirosh:
I think the balance check can be simplified, with the pre/post pattern:
1. calculate gasLimit. make sure balance is above. subtract gasLimit.
2. make the pre/relayed/post calls
3. calculate actual gas usage. refund balance with unused gas.

nventuro:
That _would_ work, but we'd need at least one extra write to storage, so at a minimum 5k extra gas. Reading the balance only costs 400 gas.

drortirosh:
> That _would_ work, but we'd need at least one extra write to storage, so at a minimum 5k extra gas. Reading the balance only costs 400 gas.

No need to use storage: we're in the context of our method (`recipientCallsAtomic`). its an in-memory variable

nventuro:
Maybe I'm misunderstanding what you're proposing, could you share a short snippet of code describing your idea? I don't see how you could prevent the recipient from overwithdrawing without storing the subtracted value in store.

test/RelayHub.test.js
```diff
@@ -0,0 +1,89 @@
+const { BN, ether, expectEvent, time } = require('openzeppelin-test-helpers');
+
+const { getTransactionHash, getTransactionSignature } = require('../src/js/relayclient/utils')
+
+const RelayHub = artifacts.require('RelayHub');
+const SampleRecipient = artifacts.require('SampleRecipient');
+
+contract('RelayHub', function ([_, relayOwner, relay, sender, other]) {  // eslint-disable-line no-unused-vars
```
nventuro: We should be able to use the `varsIgnorePattern` option to avoid doing this, but I couldn't get it to work.

drortirosh:
> Maybe I'm misunderstanding what you're proposing, could you share a short snippet of code describing your idea? I don't see how you could prevent the recipient from overwithdrawing without storing the subtracted value in store.

few lines before calling `recipientCallsAtomic`, we do 
```
require(gasPrice * initialGas <= balances[recipient], "Recipient balance too low");
```
so it should be followed by 
```
balances[recipient] -= gasPrice*initialGas
```
later, instead of `balances[recipient] -= charge`, we restore this "pre-charge", and charge the accurate amount:
`balances[recipient] += gasPrice*initialGas - charge`
(Note that the code assumes anyway `charge <  gasPrice*initialGas`, but now I read it again, and I think it might not be true, since there's the transactionfee in the `charge` calculation - but that's a separate issue to check)

nventuro:
> few lines before calling recipientCallsAtomic, we do
>`require(gasPrice * initialGas <= balances[recipient], "Recipient balance too low");`

Wouldn't this require an additional `sstore`? That's what I meant in my initial reply. As far as I know, since [EIP 1087](https://eips.ethereum.org/EIPS/eip-1087) is not yet included, each `sstore` costs (at least) 5k gas.

>(Note that the code assumes anyway charge < gasPrice*initialGas, but now I read it again, and I think it might not be true, since there's the transactionfee in the charge calculation - but that's a separate issue to check)

Yes, we've already identified this bug and are currently finishing up a PR that tackles this.

drortirosh:
> Wouldn't this require an additional `sstore`? 

why would it? we don't set any field, only stack-local variable.

frangio:
`balances` is a mapping in storage, and `balances[recipient] -= gasPrice*initialGas` modifies an element in this mapping, so it results in an `SSTORE`.

spalladino:
Thanks @drortirosh!

PR #139, opened 2019-06-21
Invalid error codes in acceptRelayedCall

nventuro:
This PR adds a check to assert that `acceptRelayedCall` doesn't return a reserved error code, and emits a new code when this happens.

### DIFF
--- a/test/RelayHub.test.js
+++ b/test/RelayHub.test.js
@@ -7,12 +7,20 @@ const SampleRecipient = artifacts.require('SampleRecipient');
 
 contract('RelayHub', function ([_, relayOwner, relay, sender, other]) {  // eslint-disable-line no-unused-vars
   const RelayCallStatusCodes = {
-    'OK': new BN('0'),
-    'CanRelayFailed': new BN('1'),
-    'RelayedCallFailed': new BN('2'),
-    'PreRelayedFailed': new BN('3'),
-    'PostRelayedFailed': new BN('4'),
-    'RecipientBalanceChanged': new BN('5'),
+    OK: new BN('0'),
+    CanRelayFailed: new BN('1'),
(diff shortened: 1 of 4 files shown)


test/RelayHub.test.js
```diff
@@ -0,0 +1,89 @@
+const { BN, ether, expectEvent, time } = require('openzeppelin-test-helpers');
+
+const { getTransactionHash, getTransactionSignature } = require('../src/js/relayclient/utils')
+
+const RelayHub = artifacts.require('RelayHub');
+const SampleRecipient = artifacts.require('SampleRecipient');
+
+contract('RelayHub', function ([_, relayOwner, relay, sender, other]) {
```
drortirosh: please check why it fails.
also, please explain why you need to make such a deep context stack, where each one is used only once...

nventuro: I addressed the linter failing in #137, which this PR is built on (it builds on the tests added there). 

Regarding the nested `context`s, you're right in that they're not needed here. I find that by breaking up complex setups (like this one, with staking, registering, deposits, etc.) Solidity contract tests are easier to write and maintain, since different features will usually be tested at different stages (e.g. attempting to `relayCall` before register, after removal or penalization, etc.).

drortirosh:
notice that with this change you *REDUCED* our code-coverage from 100% to 99.37% - one error case is not reached anymore.
https://451-161199275-gh.circle-artifacts.com/0/root/gls/coverage/contracts/index.html

nventuro:
@drortirosh note that the testfile `RelayHub.test.js` _does_ test for this scenario, but it lacks coverage measurement (I only now noticed it is manually enabled in `relay_hub_test.js`).

PR #140, opened 2019-06-21
Use standard signature format (r, s, v)

nventuro:
This PR changes the signature format the sender uses (to sign over all relayCall parameters) from (v, r, s) to (r, s, v), the same format used for Ethereum transactions and by `web3.sign`. 

In the process, I replaced the direct call to `ecrecover`, with manual byte handling, for OpenZeppelin's `ECDSA` library.

I'm not sure if some other file or helper (e.g. in the server code) needs to be updated to reflect this change.


nventuro:
Tests are failing because [`ECDSA.recover`](https://docs.openzeppelin.org/v2.3.0/api/cryptography#ECDSA.recover(bytes32,bytes)) expects the signature to be exactly 65 bytes long, which doesn't hold when dapp-specific approval data is passed.

However, while reviewing this I spotted a bug in the current implementation: any data passed after the signature in `approval` _is not signed by the sender_. I'm therefore increasing the scope of this PR to split `approval` into a `signature` and `approvalData`, which _is_ signed and checked by `RelayHub`.

nventuro:
>However, while reviewing this I spotted a bug in the current implementation: any data passed after the signature in approval is not signed by the sender.

After thinking about this in more detail, having the sender sign this parameter would make the intended use-case of a backend signing the tx data very cumbersome, and a signature is 'self-verified', so we can leave this as-is. We should however make it very clear that this value is simply forwarded and untrusted (i.e. it cannot be used for much more than a signature).

nventuro:
I'd much rather be explicit about this value not being verified in any way (a relay can effectively send whatever it want) than it be an almost accidental side effect of being packed along with the signature. 

This will also make usage, debugging, etc. easier, since `approvals` won't be an array with an implicit schema, and we remove the possibility of errors in both the slicing to read and the concatenation when sending.

>when it won't be in use in most cases

It's probably too early to tell what will and won't stick. I personally think this feature is rather important in that it facilitates onboarding to the GSN by providing a dead simple approval strategy that requires no transactions (to e.g. whitelist a user on-chain, mint dapp-specific tokens to them, etc.).

src/js/relayclient/RelayClient.js
```diff
@@ -76,7 +76,7 @@ class RelayClient {
      */
     validateRelayResponse(returned_tx, address_relay,
                           from, to, transaction_orig, transaction_fee, gas_price, gas_limit, nonce,
-                          relay_hub_address, relay_address, sig) {
+                          relay_hub_address, relay_address, approvalData, sig) {
```
alex-forshtat-tbk: We need to reorder these parameters so that our code conforms to some rules. In solidity, 'approvalData' goes last, why add it to the middle of parameters list here? Also in 'sendViaRelay'. 
As you are on it, please, make the parameters' order in Solidity and in JavaScript aligned. Thanks.

src/js/relayclient/RelayClient.js
```diff
@@ -139,13 +140,14 @@ class RelayClient {
      * Performs a '/relay' HTTP request to the given url
      * @returns a Promise that resolves to an instance of {@link ethJsTx} signed by a relay
      */
-    sendViaRelay(relayUrl, signature, from, to, encodedFunction, gasprice, gaslimit, relayFee, recipientNonce, relayHubAddress, relayAddress, relayMaxNonce) {
+    sendViaRelay(relayUrl, approvalData, signature, from, to, encodedFunction, gasprice, gaslimit, relayFee, recipientNonce, relayHubAddress, relayAddress, relayMaxNonce) {
```
alex-forshtat-tbk: Also here.

test/relay_client_test.js
```diff
@@ -365,7 +365,7 @@ contract('RelayClient', function (accounts) {
         let fromAddr = ephemeralKeypair.address
         rc.useKeypairForSigning(ephemeralKeypair)
         var did_assert = false
-        rc.sendViaRelay = function(relayUrl, signature, from, to, encodedFunction, gasprice, gaslimit, relayFee, nonce, relayHubAddress, relayAddress){
+        rc.sendViaRelay = function(relayUrl, signature, approvalData, from, to, encodedFunction, gasprice, gaslimit, relayFee, nonce, relayHubAddress, relayAddress){
```
drortirosh: The idea was to move ALL occurrences of `signature` and  `approvalData` : they both should be at the end, after all signed parameters, and `approvalData` the last param.

nventuro: Right, I though Alex simply meant having the order of signature and approvalData be consistent. If you look at `sendViaRelay` for example, many other values are out of order: the relay fee should go before the gas price, but it is after the gas limit. Should I also fix those here?

drortirosh: Yes, these should be fixed too. (there's also a small issue of the CanRelay in the go code,  which uses different parameter ordering, and can be fixed too)

contracts/RelayHub.sol
```diff
@@ -23,7 +26,7 @@ contract RelayHub is IRelayHub {
     * Assume that relay has non-zero balance (costs 15'000 more otherwise).
     */
     uint256 constant public gasReserve = 100000; // how much reserve we actually need, to complete the post-call part of relayCall().
-    uint256 constant public gasOverhead = 47241;
```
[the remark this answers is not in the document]
nventuro: I'm not sure, since the only change is the addition of this array (which has length 0 in the test that measures this overhead). We're in the process of measuring the gas consumed by each block of code in `relayCall`, we can do a more in-depth analysis of gas expenditure once those measurements are finalized.

PR #141, opened 2019-06-21
Relay management refactor

nventuro:
This PR refactors a bit the internals of the relay management functions (`stake`, `registerRelay`, etc.), adds documentation to them on `IRelayHub`, and adds new tests for these.

This will serve as a base from which to change certain current behaviours (e.g. disallow re-register after unstake), all current behaviour is preserved here.


nventuro:
I ended up removing the import of `IRelayRecipient` since the `extract_abi` script doesn't seem to be able to handle imports.

nventuro:
The following test failed, but I'm not sure why, since it seems to be unrelated to my changes:

```
  1) HttpWrapper
       should timeout after specified time:
     AssertionError: expected promise to be rejected but it was fulfilled with { Object (id, jsonrpc, ...) }
```

PR #146, opened 2019-06-28
Upgrade to solc-0.5.10

nventuro:
This upgrades the compiler version to use v0.5.10, which includes a fix for https://github.com/ethereum/solidity/issues/6901, letting us remove `staticCallWithMaxGas` and its inline assembly.

### DIFF
--- a/contracts/RelayHub.sol
+++ b/contracts/RelayHub.sol
@@ -191,38 +191,18 @@ contract RelayHub is IRelayHub {
             return uint256(PreconditionCheck.WrongNonce);
         }
 
-        bytes memory rawTx = abi.encodeWithSelector(to.acceptRelayedCall.selector,
-            relay, from, encodedFunction, gasPrice, transactionFee, approval);
+        bytes memory encodedTx = abi.encodeWithSelector(to.acceptRelayedCall.selector,
+            relay, from, encodedFunction, gasPrice, transactionFee, approval
+        );
 
-        (bool success, uint256 accept) = staticCallWithMaxGas(address(to), acceptRelayedCallMaxGas, rawTx);
+        (bool success, bytes memory returndata) = address(to).staticcall.gas(acceptRelayedCallMaxGas)(encodedTx);
 
         if (!success) {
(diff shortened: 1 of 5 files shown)


nventuro:
I'm not sure why CI is failing, I'm able to compile and run tests on Docker after updating the Dockerfile. Is the solidity version specified somewhere else?

contracts/RelayHub.sol
```diff
@@ -191,40 +191,20 @@ contract RelayHub is IRelayHub {
...
         }
 
-        bytes memory rawTx = abi.encodeWithSelector(to.acceptRelayedCall.selector,
-            relay, from, encodedFunction, gasPrice, transactionFee, approval);
+        bytes memory encodedTx = abi.encodeWithSelector(to.acceptRelayedCall.selector,
+            relay, from, encodedFunction, gasPrice, transactionFee, approval
+        );
 
-        (bool success, uint256 accept) = staticCallWithMaxGas(address(to), acceptRelayedCallMaxGas, rawTx);
+        (bool success, bytes memory returndata) = address(to).staticcall.gas(acceptRelayedCallMaxGas)(encodedTx);
```
dtr2: see the circleCI build failure: it claims the method can't be  `view`....
Could be that we hit another solc problem.

nventuro: That is exactly the bug that was fixed in 0.5.10 (see the link in the first comment) - Docker doesn't seem to be running the Dockerfile on Circle, I think there's an issue with the cache not being invalidated.

nventuro: @drortirosh I'm not extremely familiar with CircleCI and Docker, but it seems to me like the CI build is not actually running the Dockerfile commands, but rather downloading the published image (`- image: tabookey/eth-tools:solc-5.5b`). Is this so, or is there a step I'm missing somewhere>

dtr2:
Yes, circleci needs the image took be pre-built. Including the solc binary
( since our golang relay relies on abigen, which doesn't use solcjs.)
I'll create a new solc-5.10 image, and let you know.

PR #149, opened 2019-07-02
Make penalization reporter reward immediate

nventuro:
This rewards the reporter of a penalization immediately, removing the need to wait until the `unstakeDelay` has elapsed.

It also adds some new penalization tests that check this functionality works as intended, as well as extract the data required to penalize a relay from a mined transaction (`getDataAndSignatureFromHash`).


package.json
```diff
@@ -33,7 +33,7 @@
   "scripts": {
     "test": "npm run test-server && npm run test-js",
     "test-server": "make test-server",
-    "test-js": "run-with-testrpc --port 8544 --gas-limit 8000000 --defaultBalanceEther 1000 'npx truffle --network npmtest test'",
+    "test-js": "run-with-testrpc --port 8544 --gas-limit 8000000 --defaultBalanceEther 1000 --deterministic 'npx truffle --network npmtest test'",
```
drortirosh: why did you add deterministic? we preferred not to rely on specific addresses..

test/RelayHub.test.js
```diff
@@ -335,6 +338,294 @@ contract('RelayHub', function ([_, relayOwner, relay, sender, other]) {  // esli
...
+        nonce: 0,
+        privateKey: '6370fd033278c143179d81c5526140625662b8daa446c22ee2d73db3707e620c',
+      };
+
+      describe('repeated relay nonce', async function () {
+        it('penalizes transactions with same nonce and different data', async function () {
+          const txDataSigA = getDataAndSignature(encodeRelayCall(encodedCallArgs, relayCallArgs));
+          const txDataSigB = getDataAndSignature(encodeRelayCall(Object.assign(encodedCallArgs, { data: "0xabcd" }), relayCallArgs));
+
+          await relayHub.penalizeRepeatedNonce(txDataSigA.data, txDataSigA.signature, txDataSigB.data, txDataSigB.signature);
```
drortirosh: you only check if the method reverts (or not), but not if it actually remove the relay, and reward the snitcher..

package.json
nventuro: Good catch - this is from a previous version of the test suite where I included the private key for one of the relays, but the current tests don't need the key to correspond to an actual funded address.

test/RelayHub.test.js
nventuro: Correct, these tests (the whole 'penalizable behaviors' `describe` block) test the different ways in which a relay can be penalized. The following describe ('relay state and reward', not the best name) tests in which states a relay can be penalized, and what happens if it is.

Looking at Mocha's output may help understand how the tests are laid out:
```
Conctract: RelayHub  
  penalizations
    penalizable behaviors
      repeated relay nonce
        ✓ penalizes transactions with same nonce and different data
        ✓ does not penalize transactions with same nonce and same data
        ✓ does not penalize transactions with different nonces
        ✓ does not penalize transactions with same nonce from different relays
      illegal call
        with pre-EIP155 signatures
          ✓ penalizes relay transactions to addresses other than RelayHub
          ✓ penalizes relay transactions to illegal RelayHub functions (stake)
          ✓ penalizes relay transactions to illegal RelayHub functions (penalize)
          ✓ does not penalize legal relay transactions
    relay state and reward
      with penalizable transaction
        with unstaked relay
          ✓ account cannot be penalized
          with staked relay
            ✓ relay can be penalized
            once penalized
              ✓ relay cannot be penalized again
            with registered relay
              ✓ relay can be penalized
              ✓ RelayRemoved event is emitted
              once penalized
                ✓ relay cannot be penalized again
              with removed relay
                ✓ relay can be penalized
                once penalized
                  ✓ relay cannot be penalized again
                with unstaked relay
                  ✓ relay cannot be penalized
```

package.json
nventuro: Some of the penalization tests require having access to a relay's private key in order to sign transactions manually (since ganache's RPC interface doesn't support doing so).

I added back a test in e22dbba80841a6f60dc24b51a86fe72f76eeb1b2 checking that this key matches the address. I don't see much of an issue in using deterministic mode in these tests to be honest, in fact, I'd rather the test run to be reproducible as opposed to based on random values.

test/RelayHub.test.js
shahafn: I'm not sure I see an added value in separating the penalization tests to different `describe` blocks. What's the meaning of having `penalizes transactions with same nonce and different data` pass yet `relay state and reward` -> `with penalizable transaction` -> `with staked relay` -> `relay can be penalized` fail? or vice versa? why not just check the state after each penalizable flow that doesn't revert?

nventuro: >What's the meaning of having penalizes transactions with same nonce and different data pass yet relay state and reward -> with penalizable transaction -> with staked relay -> relay can be penalized fail? 

That would mean that:
a) two transactions with same nonce and different data can be penalized
b) a staked relay cannot be penalized

This could happen if e.g. the `penalize` function disallowed penalizing unregistered relays. Each `describe` block is testing different things: _the behaviors_ that can trigger a penalization, and _in which states_ penalizations can happen.

shahafn: Ok, let me correct myself then:
What's the meaning of having `penalizes transactions with same nonce and different data` pass yet `relay state and reward` -> `with penalizable transaction` -> `with **registered** relay` -> `relay can be penalized` fail? or vice versa? why not just check the state after each penalizable flow that doesn't revert?

The point I was making is that checking that the penalize didn't revert alone, and then in a different test checking the state changes is unnecessary. In the above case, if  `penalizes transactions with same nonce and different data` passes yet  `with **registered** relay` -> `relay can be penalized` fails, the first test is misleading and actually doesn't have any significant meaning, and on the other hand, if the state changed per the second test, then `penalizes transactions with same nonce and different data` will always pass with no exception but a bug in evm implementation that would revert yet allow state changes...

nventuro: > the first test is misleading and actually doesn't have any significant meaning

That would simply mean that a registered relay cannot be penalized (but perhaps an unregistered or removed one can).

>then penalizes transactions with same nonce and different data will always pass with no exception

Yes, but there are multiple reasons why a relay may be penalized, and same nonce different data is only one of them.

To get something closer to 'complete' coverage, you'd need to check all possible penalizable behaviors, on all possible states. I'm not going for a full black-box testing approach, and since I know penalizable behaviours are independent of the state the relay is in, I chose to also split those in the tests.

That said, I don't really see a reason to block this PR over such a thing, considering these new tests _do_ cover scenarios which weren't covered before, and don't really add a maintenance burden.

dtr2: the only thing we're saying is that instead of checking just for revert or not, to add a `expectPenalized()` method, which takes a promise and verify that afterward, the relay indeed lost its stake and the reported got its share.
that method should be called on each of those tests (except, of course, the test that expectRevert)

dtr2: and BTW: don't "resolve" issues that fast. It took me some time to find this thread, as it disappeared

nventuro: Ah, I see what you mean, that's a good idea. I've added that to the tests, and reworked them a bit since there was some unnecessary nesting, it should read a bit easier now.

PR #151, opened 2019-07-02
Update stake, deposit, etc. min and max values

nventuro:
As discussed, we're updating some of the min and max values, as well as adding some new ones, in preparation for the testnet launch:

* `minimumStake`: increased from 0.1 ETH to 1 ETH. A relay can perform an attack (and be penalized) at a cost of half the minimumStake: 0.05 ETH (currently ~15 USD) is IMO unnecessarily low.
* `minimumUnstakeDelay`: was zero, set to a week. 
* `maximumUnstakeDelay`: new value, set to three months. We want to both prevent relays from mistakenly locking funds for unreasonable amounts of time (e.g. a century), and being too tied to the current `RelayHub` instance, since an updated version is on the roadmap.
* `maximumStakeDeposit`: new value, set to 5 ETH. Following the reasoning behind `maximumRecipientDeposit`, it makes sense to also protect relays from making unreasonably large stake deposits at once

Many tests (particularly `ServerHelper` tests) relied on the old values, so I had to go over them to reflect these new constants.


drortirosh:
I'd make these parameters more relaxed - but enforce them in the standard client library: why block anyone from bringing up a relay with 1-day unstakeDelay, when the **default client** ignores relays with unstakeTime lower than a week ?

nventuro:
I see your point, though I'd argue that letting someone bring up a relay that will be ignored by the default client is an even worse situation: I think we should prioritize user experience over flexibility.

That said, I don't have strong feelings about this.

contracts/RelayHub.sol
```diff
@@ -8,15 +8,23 @@ import "@0x/contracts-utils/contracts/src/LibBytes.sol";
...
+    // will not be able to immediatly start serving requests.
     uint256 constant public minimumRelayBalance = 0.1 ether;
 
-    // Maximum funds that can be deposited at once. Prevents user errors by disallowing large deposits.
-    uint256 constant public maximumDeposit = 2 ether;
+    // Maximum funds that can be deposited at once. Prevents user error by disallowing large deposits.
+    uint256 constant public maximumRecipientDeposit = 2 ether;
+
+    // Maximum amount that can be staked at once. Prevents user error by disallowing large deposits.
+    uint256 constant public maximumStakeDeposit = 5 ether;
```
drortirosh: why prevent large staking? 
for recipients, it makes sense since:
1. 2 ether allows huge # of transactions, 
2. security-wise, deposit means nothing, and 
3. RR Implementation might forget to add a "withdraw" implementation, making this deposit "non-refundable" for the foreseeable future..

But for relay,  staking means the trust a user can put on it: I might prefer a 100-ether-staked relay for high-valued transactions.  Also, there is always a way to unstake (even if with some delay) - so there is no real reason to have such a limit.

nventuro: To clarify, this doesn't prevent large stakes, but rather caps how much a stake can be increased by in a single transaction. Similarly, a user may deposit as much ether as they want in a contract, but they have to do it over multiple calls to `deposit`. 

Limiting this value (just like `unstakeDelay` is also capped) seems like a good idea from a usability perspective: it limits the severity of user mistakes (e.g. you won't stake all of your balance due to a typo).

Doing a couple transactions to achieve a stake higher than 5 ETH doesn't seem like a big issue imo (considering 5 ETH is already ~1.5k USD), but if we agree it makes sense to have this cap, we can increase this value is you think the limit is low.

drortirosh: This limit of a kind of saying that we don't think you should stake more.
Yes, it makes little sense to have a cap, but it should be big, e.g 100
UI tool to deposit it might add warning are 'are you sure' messages. It's not the contact's job.

nventuro: I removed `maximumStakeDeposit`: I agree with your assessment that there is a mechanism to retrieve it anyway, so this is more of an UI concern and not the contract's.

drortirosh:
I think that one is OK. just resolve the conflict.

nventuro:
@drortirosh waiting for your approval of the PR to merge

PR #158, opened 2019-07-04
Show server errors in RelayClient

spalladino:
If the server returned an error on body.error, since it was missing a body.nonce, the client would just output "empty server response", making debugging difficult. This change checks if body.error is present and shows it if so. Note that if the server returns body.error, it is assumed that the request failed.


spalladino:
Closing PR - it was already fixed on a commit I had missed.

PR #159, opened 2019-07-04
Support setting txfee in tx options and add default value

spalladino:
Instead of just accepting tx fee when initializing the relay provider, support specifying the txfee parameter along with from, gas, to, gasPrice, etc in each tx. Also defaults fee to 12, which is above the relayer default minFee of 11.


drortirosh:
I actually think this is wrong: 
There should not be a default fee on the client.
Instead, the client should default to the fee advertised by the relay.
The user may override it with a different value (either higher or lower )

Note that a static default is bad since:
- if the relay's required txfee is higher, then the transaction will be rejected by the relay.
- if the relay's required tx if lower, then the user is simply over-pay.

PR #161, opened 2019-07-05
Make removals and penalizations permanent

nventuro:
Once a relay has been removed or penalized, we should disallow it from being staked for and registered again, since a) relays should always be new accounts, and b) a penalized relay could be re-penalized if it was ever reused.

This PR adds these two new states, where removed (unstaked) and penalized relays remain forever. I'm not a fan of the names, and I'm open to suggestions (`removeRelay` causes it to go to `Removing`, and `unstake` causes it to become `Removed` - perhaps `deregisterRelay` and `Deregistering` work better?).

Many tests in `relay_hub_test.js` relied on the fact that removed and penalized relays could be re-used (since those tests are all interdependent), so this change broke them. #149 added these same tests in `RelayHub.test.js` so we won't be losing coverage, but I'd appreciate a second set of eyes to check that I didn't forget to cover for one of the test cases.


dtr2:
Why bother? Relays addresses should not reuse any address that already
issued any transaction, since it makes them penalizeable.
This is unavoidable, and can't be presented on chain.
Also, if there is a new RelayHub, relays should not be moved to the new
relay, as they are subject to attack.
Likewise, one should not reuse an address once a relay is stopped or
penalized. I don't see any merit for checking and explicitly preventing it.

nventuro:
True, and I was simply mentioning reasons why this change is specially desirable in this system as a whole, but the underlying reason why such a change is a good thing is general consistency. 

Being able to query a relay's status and getting as a response that it has been removed, or penalized, is a much better experience than these relays simply disappearing from the blockchain. Similarly, it is easier to reason about security and soundness with explicit states: a relay cannot be penalized if it is already in the penalized state, it can only be unstaked if it is being removed, etc.

We could write this program in a minimal way with zero redundancy, but in return we'd get reduced readability and single points of failure. Given the characteristics of smart contract development, I'd try to avoid such scenarios.

drortirosh:
> Being able to query a relay's status and getting as a response that it has been removed, or penalized, is a much better experience than these relays simply disappearing.

True - except that I disagree it is required off-chain.
A relay is **off-chain** entity. There's an easy way to know its state: simply search for relayAdded/removed/penalized with its address.

We don't think your quite esoteric use-case warrants adding all that code.

shahafn:
I don't think either that adding new states and complexity to the code just to prevent a rare edge-case is necessary. 
As Dror said, one can scan all RelayAdded events on the blockchain and look for his address to see if his relay was used before.

yoav-tabookey:
I think we discussed this on a call a while ago.  We can't test for it on-chain but it's trivial to check from a client (or a relay).  A relay should never attempt to register or stake itself, if the nonce of its address is non-zero.  No way to see it on chain, but the relay does see it.  As for clients checking the status of a certain relay, we have events for that.  If a relay has been unregistered or penalized, RelayHub events tell the story.  I did want to check for it in RelayHub, for completeness, but I don't know if it's worth storing in contract-readable storage just for that, when it's only useful for off-chain entities.

PR #165, opened 2019-07-08
Revert "Support setting txfee in tx options (#159)"

drortirosh:
This reverts commit 21a348d2f0ba8df8065b349e8d22ec09c535c14a.

@spalladino , @shahafn :
I actually think this is wrong:
There should not be a default fee on the client.
Instead, the client should default to the fee advertised by the relay.
The user may override it with a different value (either higher or lower )

Note that a static default is bad since:

if the relay's required txfee is higher, then the transaction will be rejected by the relay.
if the relay's required tx if lower, then the user is simply over-pay.


drortirosh:
instead of this, I created #170  to fix it.

shahafn:
Fixed in #170

spalladino:
Sounds good @drortirosh. I went with this quick solution using the relayer's default tx fee to get the web3 provider working right out of the box.

PR #169, opened 2019-07-10
Unify solc version 

drortirosh:
Make the package.json the source of current solc version to use.
- define explicit solc version in package.json (no "^")
- read this value from truffle.js and coverage-prov.js
- (server code alreayd uses current solcjs)


server/scripts/test.sh
```diff
@@ -10,6 +12,16 @@ onexit() {
   fi
 }
 
+solcjs_ver=`cd ..;npx solcjs --version|perl -ne 'print $1 if /(\d+\.\d+\.\d+)/'`
```
shahafn: This isn't needed anymore as I removed native solc dep in #171

PR #170, opened 2019-07-10
use registerRelay's transactionFee as default

drortirosh:
in case the client doesn't specify txfee explicitly, use the value the
relay used to register with.
Note that anything higher is an overpay for the client, and anything
lower, will probably get rejected as too low.


drortirosh:
@spalladino : I don't think there's a need for `maxTxFee` parameter: 
The client can still explicitly specify `txfee`, and accept whatever consequences (overpay/reject)
Note that custom relay may advertise high fee, but accept low fee from known clients (or for known targets). 
The rationale:
- I have a dapp. I don't want to pay txfee to others, so I bring up my own relay.
- I want to "prioritize" my own clients, so I advertise a premium txfee, to prevent others to connect to it by default.
- my clients know they can use this relay, and that it will accept lower txfee from them.
- still, if my relay is offline, I will still get service from other relays (yes, paying their txfee )
- my relay is still acting as an "umbrella" relay to other dapps for exactly such cases - for a premium fee.

spalladino:
> The client can still explicitly specify txfee, and accept whatever consequences (overpay/reject)

Not sure I agree, a client should have the means to never overpay, while setting a cap to how much they want to disburse. Still, this is definitely not critical, so we can review it later.

> Note that custom relay may advertise high fee, but accept low fee from known clients (or for known targets).

Loved this use case! I had always had in mind app-specific relayers that only accept txs for a specific app, but never thought about still allowing other apps at a higher premium.

PR #178, opened 2019-07-12
 relayCall events overhaul (#150 continued)

nventuro:
Continuation of #150. I mistakenly deleted the branch on my fork backing that PR, and GitHub doesn't' allow reopening in such cases, so I created this new one.


src/js/relayclient/RelayClient.js
```diff
@@ -416,11 +416,19 @@ class RelayClient {
...
+
+            const setErrorStatus = (reason) => {
+                console.log(`${reason}. changing status to zero`)
                 resp.result.status = 0
             }
+
+            if (canRelayFailed) {
+                setErrorStatus(`canRelay failed: ${canRelayFailed.find(e => e.name == "reason").value}`)
+            } else if (transactionRelayed && transactionRelayed.events.find(e => e.name == "status").value != 0) {
+                setErrorStatus("reverted relayed transaction")
```
dtr2: any special reason to hide the status from the client?

nventuro: Note that the original message didn't display any status codes: I attempted to leave the client as unchanged as possible. I can of course add this value though, as I did in the other clause of the `if` statement.

PR #181, opened 2019-07-17
IRelayRecipient parameters

### LINKED ISSUE #155: Use ABIEncoderV2 to work around 'stack too deep' issues
`relayCall` and `recipientCallsAtomic` are hard to make changes to because they have too many local variables stored in the stack, causing the compiler to emit 'stack too deep' errors. This is particularly problematic because we actually need to have these functions hold _more_ variables, since some required values are not currently being forwarded, and we still need to rework the charge calculations, which will take some extra slots.

A clean and maintainable way to do achieve this is storing `relayCall`'s arguments in a `struct`, which are kept in `memory`. However, since `recipientCallsAtomic` is an `external` function, we need to enable the ABIEncoderV2 (an experimental feature) for this to work. This may raise concerns from a security point of view, since this encoder hasn't been as tested as the default one (though our use case is quite straightforward and doesn't require any of its tricky bits).

Additionally, it is likely that usage of the new encoder would bleed over into `IRelayRecipient`, which means that developers writing contracts that are GSN-callable will get the following warning when compiling their code:

```
Warning: Experimental features are turned on. Do not use experimental features on live deployments.
pragma experimental ABIEncoderV2;
^-------------------------------^
```

Are we ok with doing such a thing? I think this is a very reasonable course of action (and the only one I can think of), and fine for an MVP, but wanted to get thoughts and ideas around this.

### LINKED ISSUE #180: IRelayRecipient parameters
The `IRelayRecipient` functions currently on `master` are currently lacking a couple of values that are required for them to be able to achieve their intended purposes.

For example, `acceptRelayedCall` doesn't receive `gasLimit` or `nonce`, both of which should be used in an acceptance scheme where `approvalData` contains a signature over relay parameters. If, on the other hand, the recipient intends to charge the user, `accept` should first check that they have enough balance, for which `gasLimit` is required.

Similarly, `pre` and `post` require both the `gasPrice` and `gasLimit` in order to compute charges.

Furthermore, it makes sense to forward the max possible recipient charge to `accept` and `pre`, since a) the hub needs to compute this value anyway (to check the recipient has enough balance), and b) the recipient would need to know about RelayHub's constants, gas stipends, etc., in order to calculate this value.

---

Additionally, I wanted to propose _removing_ some of the received values, for which we haven't thought of a use case yet, such as `relay` and `encodedFunction` in both `pre` and `post` (but leaving them in `accept`). If `maxPossibleCharge` is added to `pre`, we could also get rid of `gasLimit`, `gasPrice` and `transactionFee`. 

Removing these means not only a smaller and easier to understand recipient interface, but also gas savings: I measured total gas costs reductions of ~8k gas after reducing the interface to the proposed set shown below:

```solidity
function acceptRelayedCall(
  address relay, 
  address from, 
  bytes memory encodedFunction, 
  uint256 transactionFee, 
  uint256 gasPrice, 
  uint256 gasLimit, 
  uint256 nonce, 
  bytes memory signature, 
  bytes memory approvalData, 
  uint256 maxPossibleCharge
) public view returns(uint);

function preRelayedCall(
  address from, 
  uint256 maxPossibleCharge
) public returns (bytes32);

function postRelayedCall(
  address from,
  bool success, 
  uint256 actualCharge, 
  bytes32 preRetVal
) public;
```

nventuro:
Closes #155 
Closes #180 

The main change in this PR is adding some values that were missing to `acceptRelayedCall`, `preRelayedCall` and `postRelayedCall`.

That required using the ABIEncoderV2 inside of RelayHub (see #155), but not on the external interface. However, multiple parts of the code (specially the server code) where interacting directly with `RelayHub.sol` and not `IRelayHub.sol`. Some of the Go parsers don't know how to handle this new encoder, so I changed those files to start using the interface as opposed to the implementation, and moved functions, getters, etc., around accordingly.

In the process, I also added the `requiredGas` and `maxPossibleCharge` functions to help the server, clients, recipients, etc. get these values (which  were being manually calculated in these places).


contracts/RelayHub.sol
```diff
@@ -352,11 +353,14 @@ contract RelayHub is IRelayHub {
 
     function recipientCallsAtomic(
         MetaTxData memory metaTx,
-        uint256 initialGas
+        uint256 preChecksGas
     )
     public
-    returns (RelayCallStatus)
+    returns (RelayCallStatus, uint256)
```
shahafn: Where is this value used?

contracts/RelayHub.sol
```diff
@@ -266,80 +288,94 @@ contract RelayHub is IRelayHub {
...
-        address relayAddr,
-        bytes calldata encodedFunction,
-        uint256 transactionFee,
-        uint256 gasLimit,
-        uint256 initialGas
+        MetaTxData memory metaTx,
+        uint256 preChecksGas
     )
-    external
+    public
```
shahafn: Shouldn't we instead change almost all other functions to external instead of public? this (and most other functions) are not going to be called internally anyway.

shahafn:
I think that this PR should be split as it contains several changes across the code not necessarily related to each other e.g. some of the helper functions refactoring, gas calculations etc.

contracts/RelayHub.sol
nventuro: That's not a bad idea, but I didn't want to include more changes that needed on this PR, considering its scope grew due to unrelated reasons.

Unfortunately, `recipientCallsAtomic` _has_ to remain `public` due to a bug in the Solidity compiler, which causes it to disallow copying structs from `calldata` into `memory`. The only way to have the struct be in `memory` is by having the function be `public` (`external` requires a `calldata` location).

I've spoken with Chris (lead Solidity dev) about this, and they don't have an estimate for when a fix will be released.

nventuro:
The current commits are messy in that sense due to how each change required fixing new issues that started popping up, so the overall order doesn't seem logical. I'll replay the changes in an order that will be easier to follow and update my branch.

drortirosh:
Great. If you keep a single or, then please at least  keep the resulting commits one per change, so we can merge them without a squash

nventuro:
This should be much easier to follow in commit order. The reason why seemlingly unrelated changes are performed before the actual API change is:

 * we need more parameters to be sent to the recipient
 * this requires more stack variables in `atomicRecipientCalls`, which isn't possible causing us to add the ABIEncoderV2
 * this causes issues with the Go code autogeneration (which cannot handle the encoder), so we need to have it use `IRelayHub` instead of `RelayHub`
 * this requires some functions that were not in the interface to be added there
 * all in all these changes increase total code size and cause it to go over the maximum limit (24k), so we need to reduce the API a little bit to bring this down (remove unnecessary constant getters, `stakeOf`/`ownerOf`, etc.)

server/src/librelay/relay_server_test.go
```diff
@@ -251,10 +251,10 @@ func TestMain(m *testing.M) {
 	defer os.Exit(exitStatus)
 }
 
-func resolveLibrary(path string, address common.Address, relayHubBinUnresolved string) string {
+func resolveLibrary(path string, address common.Address, IRelayHubBinUnresolved string) string {
 	libraryPlaceHolder := "__$" + hexutil.Encode(crypto.Keccak256([]byte(path)))[2:36] + "$__"
-	RelayHubBin := strings.Replace(relayHubBinUnresolved, libraryPlaceHolder, address.Hex()[2:], -1)
-	return RelayHubBin
+	IRelayHubBin := strings.Replace(IRelayHubBinUnresolved, libraryPlaceHolder, address.Hex()[2:], -1)
```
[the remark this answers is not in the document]
nventuro: Yes, but that test file still needs to deploy an instance, which is why I ended up hardcoding the bytecode of the contract. It'd be great to have it be read automatically from the artifacts, but I'm not very proficient in Go and wanted to touch these files as little as possible.

contracts/IRelayRecipient.sol
```diff
@@ -29,10 +29,22 @@ contract IRelayRecipient {
...
+    function acceptRelayedCall(
+        address relay,
+        address from,
+        bytes memory encodedFunction,
+        uint256 transactionFee,
+        uint256 gasPrice,
+        uint256 gasLimit,
+        uint256 nonce,
+        bytes memory approvalData,
+        uint256 maxPossibleCharge
```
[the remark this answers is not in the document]
nventuro: Simply because the cost of passing it along is very low, whereas calculating it (again, recall that RelayHub needs to compute this value anyway) can get expensive.

contracts/RelayHub.sol
```diff
@@ -211,7 +211,7 @@ contract RelayHub is IRelayHub {
             return uint256(PreconditionCheck.WrongSignature);
         }
 
-        // Verify the transaction is not being repalyed
+        // Verify the transaction is not being replayed
```
shahafn: If you're at it, there is one in line 387: responsability -> responsibility :+1:

server/src/librelay/relay_server_test.go
```diff
@@ -184,13 +184,13 @@ func NewRelay(relayHubAddress common.Address) {
...
...08115159050919050565b6000819050919050565b60007fffffffff0000000000000000000000000000000000000000000000000000000082169050919050565b6000819050615cc582615e06565b919050565b6000819050615cd882615e13565b919050565b600073ffffffffffffffffffffffffffffffffffffffff82169050919050565b6000819050919050565b600060ff82169050919050565b6000615d1f82615d4a565b9050919050565b6000615d3182615cb7565b9050919050565b6000615d4382615cca565b9050919050565b6000615d5582615d5c565b9050919050565b6000615d6782615cdd565b9050919050565b82818337600083830152505050565b60005b83811015615d9b578082015181840152602081019050615d80565b83811115615daa576000848401525b50505050565b6000615dbb82615dcc565b9050919050565b6000819050919050565b6000615dd782615df9565b9050919050565b6000819050919050565b6000601f19601f8301169050919050565b60008160601b9050919050565b60058110615e1057fe5b50565b60048110615e1d57fe5b50565b615e2981615c51565b8114615e3457600080fd5b50565b615e4081615c81565b8114615e4b57600080fd5b50565b615e5781615c8b565b8114615e6257600080fd5b50565b60058110615e7257600080fd5b50565b615e7e81615cfd565b8114615e8957600080fd5b5056fea365627a7a72305820140e1a345ff09aea1fd0f8d57c8980da2a20ebc054905960a535065b418bd3f36c6578706572696d656e74616cf564736f6c634300050a0040"
```
shahafn: As I wrote in another comment, this isn't necessary and can be fixed without changing the original Go code:



In order to solve it, you can have these changes:

    Change server/scripts/get_abi_bin.js:
    var contractsToExtract =[ "IRelayHub", "RelayHub", "SampleRecipient" ];
    Change server/Makefile:
    instead of TRUFFLE_OUT ( which should probably be renamed to RELAYHUB_OUT), have two vars:
    RELAYHUB_BIN=$(buildpath)/../contracts/RelayHub.json
    RELAYHUB_ABI=$(buildpath)/../contracts/IRelayHub.json

then use abigen in the GEN_FILE target like that:
abigen --solc ../node_modules/.bin/solcjs --abi $(subst contracts,server/contracts,$(RELAYHUB_ABI:json=abi)) --bin $(subst contracts,server/contracts,$(RELAYHUB_BIN:json=bin)) --pkg librelay --type RelayHub --out $@

then you wouldn't need to change any Go code.

nventuro:
Closing in favor of #186.

Issue #183, opened 2019-07-17
Relay server can be penalized after removal


nventuro:
The current implementation of the relay server sends all its funds to its owner once it sees a `RelayRemoved` event: https://github.com/tabookey/tabookey-gasless/blob/8bd4047630e001725acba6572022780faa5cc454/server/src/RelayHttpServer.go#L357

This is a relay transaction to a non-RelayHub address, and the relay can therefore be immediately penalized. This scenario is tested here: https://github.com/tabookey/tabookey-gasless/blob/8bd4047630e001725acba6572022780faa5cc454/test/RelayHub.test.js#L464

The relay should not send any transactions until the `unstakeDelay` has elapsed.

PR #184, opened 2019-07-17
Fix server to look after Unstaked event before sending balance

shahafn:
Fixing #183

### DIFF
--- a/test/relay_client_test.js
+++ b/test/relay_client_test.js
@@ -15,6 +15,7 @@ const localhostOne = "http://localhost:8090"
 
 const testutils = require('./testutils')
 const register_new_relay = testutils.register_new_relay;
+const increaseTime = testutils.increaseTime;
 
 const Big = require( 'big.js')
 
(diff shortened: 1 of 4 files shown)


test/relay_client_test.js
```diff
@@ -450,26 +451,49 @@ contract('RelayClient', function (accounts) {
...
-            relayBalance = await web3.eth.getBalance(relayServerAddress);
-            i++
-        }
-        assert.equal(0,relayBalance)
-        let afterOwnerBalance = await web3.eth.getBalance(relayOwner);
-        assert.equal(true,parseInt(afterOwnerBalance)  > parseInt(beforeOwnerBalance))
+    describe("relay balance management", async function () {
+        let relayServerAddress;
+        let beforeOwnerBalance;
+        it("should NOT send relay balance to owner after removed", async function () {
```
drortirosh: Essentially, you wait here 4 seconds to verify the relay does not send out it's balance...

Can't you test it with go unittest?

test/relay_client_test.js
```diff
@@ -450,26 +451,49 @@ contract('RelayClient', function (accounts) {
...
+
+            let i = 0;
+            let relayBalance = await web3.eth.getBalance(relayServerAddress);
+            while (relayBalance != 0 && i < 20) {
+                await testutils.sleep(200);
+                relayBalance = await web3.eth.getBalance(relayServerAddress);
+                i++
+            }
+            let afterOwnerBalance = await web3.eth.getBalance(relayOwner);
+            assert.equal(false,parseInt(afterOwnerBalance)  > parseInt(beforeOwnerBalance))
```
drortirosh: Huh? Either do assert.ok, our (better) asset.equal.
Not really sure why you test it this way: the relay don't do any operation in this period, so you should do assert equal (before, after)

shahafn: Yes I'm tired

test/relay_client_test.js
shahafn: I can, but this is not really a unit test, so decided to keep it in js.

test/relay_client_test.js
```diff
@@ -450,26 +451,50 @@ contract('RelayClient', function (accounts) {
...
+            relayServerAddress = JSON.parse(response.body).RelayServerAddress;
+            beforeOwnerBalance = await web3.eth.getBalance(relayOwner);
+            let res = await rhub.removeRelayByOwner(relayServerAddress, {from:relayOwner});
+            let etherSpentByTx = res.receipt.gasUsed * (await web3.eth.getGasPrice());
+            assert.equal("RelayRemoved", res.logs[0].event);
+            assert.equal(relayServerAddress.toLowerCase(), res.logs[0].args.relay.toLowerCase());
+
+            let i = 0;
+            let relayBalance = await web3.eth.getBalance(relayServerAddress);
+            while (relayBalance != 0 && i < 10) {
```
drortirosh: this loop is compeltely redundant here. simply do sleep(2000), and read the balance afterwards. 
You're not looking for the *shortest* time until it changes - you verify it doesn't change at all after 2 secoinds...

drortirosh: and don't force-push changes. I can't see what changes you've made after my comments

test/relay_client_test.js
alex-forshtat-tbk: I am not sure js tests should cover server behavior.

shahafn: This was already there. I think that as this one is more like integration tests so it's not the end of the world having these tests in js.

PR #191, opened 2019-07-23
Update singleton deployment script

nventuro:
Related to #111 

This updates the deployment scripts for the singleton instance, including droping the requirement of having a native `solc` installation by using `solcjs` directly, and reducing the gas limit to the actual required value of 4.2 million, since any leftover ether is lost in the process.


nventuro:
>when (and how) do we use this compute.sh script?

I didn't add much documentation about this yet since it is still somewhat in flux, but now is probably a good time to start.

Once a version has been frozen and is ready to be deployed, `./scripts/singleton.compute.sh` should be run.  Note that _any_ changes to the source, including comments, will result in a new deployment address, so we should be sure all pending issues have been tackled.

The script will compile the contract and generate:
 1. the flattened source (used to verify e.g. on Etherscan)
 2. binaries and abi files
 4. a custom file (`singleton/deploy.json`) with the resulting singleton address, the deployment transaction, and the address that needs to be funded for the deployment

We will want to commit these files to the repo, since they contain all required information to verify and repeat the deployment process on any network (including local networks, i.e. ganache).

I've also included a script to run the deployment process, via `npx truffle exec scripts/singleton/deploy.js --network <network>`, which will send funds to the deployment address and then broadcast the deployment transaction.

shahafn:
@nventuro can you add this documentation to the code/readme? It's pretty simple but still would be nice to have there.

nventuro:
@shahafn I've added the documentation to the scripts themselves: I didn't want to include this in the readme since it is already quite busy.

PR #198, opened 2019-07-31
Add cross compile task in server makefile

spalladino:
Uses karalabe/xgo for cross compiling the relay server for darwin, windows, and linux. Note that the vanilla go cross compiling does not work here, since we have dependencies with native extensions.

As part of this, I had to move two go files from the root of the src directory. By convention, go requires that all code is in a subfolder of the main src folder - and xgo would not detect the files to compile otherwise.


spalladino:
The command is `make release-binaries`. I didn't automate the publishing to github yet, I was considering doing that manually for now and then automating it. I can work on that sooner though, and send it as another PR on top of this one.

spalladino:
That's true for production deployments (though I think that https support should be handled on a script outside the relayer process, and not as part of the relayer itself), but for development there is a lot of value on having a standalone process.

spalladino:
@shahafn rebased on top of master, if the build is green this should be ready to merge!

PR #199, opened 2019-07-31
Add dev mode to relayer server

spalladino:
Adds a new bool flag -DevMode (defaults to false). When set, the relayer does not cache its account nonce, and does not try resending unconfirmed txs. This is useful when running unit tests against a relayer instance using ganache: if the test suite rollbacks the blockchain state on each test, the relayer account nonce falls out of sync. This mode prevents this, and allows having a long-running relayer for the whole test suite.

### DIFF
--- a/server/src/RelayHttpServer.go
+++ b/server/src/RelayHttpServer.go
@@ -23,7 +23,7 @@ const VERSION = "0.4.0"
 
 var KeystoreDir = filepath.Join(os.Getenv("PWD"), "build/server/keystore")
 var delayBetweenRegistrations = 24 * int64(time.Hour/time.Second) // time.Duration is in nanosec - converting to sec like unix
-var shortSleep bool                                               // Whether we wait after calls to blockchain or return (almost) immediately. Usually when testing...
+var devMode bool                                                  // Whether we wait after calls to blockchain or return (almost) immediately. Usually when testing...
 
 var ready = false
 var removed = false
@@ -51,7 +51,7 @@ func main() {
 	http.HandleFunc("/getaddr", getEthAddrHandler)
 
 	timeUnit = time.Minute
-	if shortSleep {
+	if devMode {
 		timeUnit = 100 * time.Millisecond
 	}
 	stopKeepAlive = schedule(keepAlive, 60*timeUnit, 0)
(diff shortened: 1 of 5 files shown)


yoav-tabookey:
When this mode is set, it should refuse to come up if connected to mainnet.  Otherwise it sets itself to get penalized.  I'm not sure we can detect all mainnets, e.g. other EVM-compatible blockchains like POA or ThunderCore, so maybe it would be best to only support this mode with a known network ID that we set in ganache-cli in our test scripts.

spalladino:
> so maybe it would be best to only support this mode with a known network ID that we set in ganache-cli in our test scripts.

I think it may restrict a lot its usage if we force a specific network ID for it to run. That said, I agree on refusing to connect to well-known chains where it can be penalized. I'm adding a commit that aborts execution if devmode is on and [chain ID is under 1000](https://ethereum.stackexchange.com/questions/17051/how-to-select-a-network-id-or-is-there-a-list-of-network-ids).

server/src/RelayHttpServer.go
```diff
@@ -185,6 +187,7 @@ func parseCommandLine() (relayParams librelay.RelayParams) {
 	ethereumNodeUrl := flag.String("EthereumNodeUrl", "http://localhost:8545", "The relay's ethereum node")
 	workdir := flag.String("Workdir", filepath.Join(os.Getenv("PWD"), "build/server"), "The relay server's workdir")
 	flag.BoolVar(&shortSleep, "ShortSleep", false, "Whether we wait after calls to blockchain or return (almost) immediately")
+	flag.BoolVar(&devMode, "DevMode", false, "Enable developer mode (do not retry unconfirmed txs, do not cache account nonce)")
```
drortirosh: I think it can be merged with ShortSleep, which is our initial code for test-enabled relay..

spalladino:
> The one thing I miss here is using it: you enable this DevMode in the CI tests

There are tests that check behaviour that is specifically disabled by devmode, so it's not possible to just toggle it and re-run all tests with it. Honestly, I'd stick with the current test that manually enables DevMode and checks relaying a tx.

server/src/RelayHttpServer.go
spalladino: Sounds good! I was tempted to do it, but didn't want to introduce that breaking change. Should I remove ShortSleep then in favor of DevMode?

shahafn: Yes

PR #201, opened 2019-08-02
Add destination parameter to rhub withdraw

spalladino:
If a recipient wants to withdraw funds from the rhub, the contract itself needs to call the `withdraw` function on the hub, which transfers funds to the contract. This forces the contract to implement a fallback payable function (ie to accept funds) in order to receive that transfer. Adding a fallback payable function is generally not a good idea, unless there is a mechanism in place to process arbitrary payments made to the contract.

To fix this, we can add a `dest` parameter to rhub's `withdraw`, which indicates who actually receives the withdrawn funds. This removes the requirement for any recipient to have to accept (and manage) funds from any source. Old behaviour can be preserved by just specifying the contract's own address as `dest`.

We considered the alternative of adding a fallback payable function to the base Recipient with a check `require(msg.sender == rhub)`, but that still requires a separate mechanism built-in for moving those funds. We believe this solution is much cleaner, as the funds do not need to go through the recipient contract at all, and are sent directly to a target wallet.

### DIFF
--- a/contracts/IRelayHub.sol
+++ b/contracts/IRelayHub.sol
@@ -69,10 +69,10 @@ contract IRelayHub {
     // Withdraws from an account's balance, sending it back to it. Relay owners call this to retrieve their revenue, and
     // contracts can also use it to reduce their funding.
     // Emits a Withdrawn event.
-    function withdraw(uint256 amount) public;
+    function withdraw(uint256 amount, address payable dest) public;
 
     // Emitted when an account withdraws funds from RelayHub.
-    event Withdrawn(address indexed dest, uint256 amount);
+    event Withdrawn(address indexed account, address indexed dest, uint256 amount);
 
     // Relaying
 
--- a/contracts/RelayHub.sol
+++ b/contracts/RelayHub.sol
@@ -172,14 +172,14 @@ contract RelayHub is IRelayHub {
      * note that while everyone can `depositFor()` a contract, only
      * the contract itself can withdraw its funds.
      */
-    function withdraw(uint256 amount) public {
+    function withdraw(uint256 amount, address payable dest) public {
         address payable account = msg.sender;
         require(balances[account] >= amount, "insufficient funds");
 
         balances[account] -= amount;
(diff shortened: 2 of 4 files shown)


yoav-tabookey:
We discussed this long ago and decided against it, but let's discuss the tradeoffs and decide on the best solution.

The reason we decided against it, was that we didn't want code that sends funds to an arbitrary address which isn't known to RelayHub.  Otherwise, even relatively simple bugs can be used to drain the funds.  We figured the best way is to always return the money to its known owner (the contract that owns the deposit) and let that contract send the funds back to its hardcoded owner (or whatever address that makes sense for that contract).  Transfers to fixed addresses are generally safer than user-provided addresses.

But I do see your point about payable fallback functions.  We don't like those either.

Conceptually, what we want to happen, is that the withdraw() call returns the funds to the caller (RelayRecipient) and let it handle, but in practice it shouldn't be a payable fallback function.

How about adding a payable function to RelayRecipient, receiveRefund(), callable by RelayHub only, and have RelayHub.withdraw() always send the funds to recipient's receiveRefund()?  This way it's not a fallback function, but we can still send to a fixed address.  For most recipient contracts, at least Ownable ones, receiveRefund() would just transfer to owner.  If they want to send to an arbitrary address, let them do it from there.

spalladino:
Thanks for the response, Yoav! I see your point on trying to mitigate bugs by preventing transfers to arbitrary addresses. However, moving this burden to the user will only make it more likely for them to lose funds, since it makes the implementation more complex now for them. I'd like to avoid adding yet another function to the interface they have to implement - even if it's a rather small one.

I'd appreciate if you can reconsider this implementation, especially considering it is really short, and does not interact with any other parts of the code. That said, if you still think it's potential unsafe, I'm fine with going with your proposed solution in the interest of time.

yoav-tabookey:
Looking at the code again, I think it's fine.  Deposits are supposed to be fairly small anyway, so no need for a complex design.

spalladino:
Thanks folks!

PR #202, opened 2019-08-02
Fix recipient balance check

spalladino:
The check that the recipient had enough balance for paying for a relayed tx was not considering the tx fee.

### DIFF
--- a/contracts/RelayHub.sol
+++ b/contracts/RelayHub.sol
@@ -287,7 +287,7 @@ contract RelayHub is IRelayHub {
 
         // We don't yet know how much gas will be used by the recipient, so we make sure there are enough funds to pay
         // for the maximum possible charge.
-        require(gasPrice * initialGas <= balances[recipient], "Recipient balance too low");
+        require(calculateCharge(initialGas, gasPrice, transactionFee) <= balances[recipient], "Recipient balance too low");
 
         bytes4 functionSelector = LibBytes.readBytes4(encodedFunction, 0);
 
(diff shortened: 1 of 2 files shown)


spalladino:
@drortirosh do you think we can still merge this PR, and add the missing tests later?

contracts/RelayHub.sol
```diff
@@ -287,7 +287,7 @@ contract RelayHub is IRelayHub {
 
         // We don't yet know how much gas will be used by the recipient, so we make sure there are enough funds to pay
         // for the maximum possible charge.
-        require(gasPrice * initialGas <= balances[recipient], "Recipient balance too low");
+        require(calculateCharge(initialGas, gasPrice, transactionFee) <= balances[recipient], "Recipient balance too low");
```
shahafn: Shouldn't it be `maxPossibleCharge(gasLimit, gasPrice, transactionFee) <= balances[recipient]`?

shahafn:
I opened another PR also changing the test to handle recipient balance better. Please take a look and tell me what you think.

contracts/RelayHub.sol
spalladino: I'm not sure, doesn't `initialGas` already account for the additional pre and post gas costs? L286 checks that `initialGas` is already over `requiredGas(gasLimit)`, where `requiredGas` adds all the overheads (`gasOverhead + gasReserve + acceptRelayedCallMaxGas + preRelayedCallMaxGas + postRelayedCallMaxGas`). Changing this into `maxPossibleCharge` would end up counting those overheads twice.

Either way, I've asked for another set of eyes to look into this, just in case.

spalladino: I've just noted I misread your suggestion: I had understood `maxPossibleCharge(initialGas, gasPrice, transactionFee)` instead of `maxPossibleCharge(gasLimit, gasPrice, transactionFee)`. It makes sense now, and I have asked around and there is agreement on it. Thanks Shahaf!

spalladino:
Closing in favor of #204. Thanks for the fix, @shahafn!!

PR #214, opened 2019-08-27
RelayClient: should check if "canRelay()" before calling a relay.

drortirosh:
The client should check if the transaction it generates is valid
on-chain, before sending it to a relay.

This is done by calling the same `RelayHub.canRelay()` method used by
the relay itself to check the transaction validity.

Without this check, in case the transaction would fail, the relay would
reject it. But in this case, the client would continue to try and send
the transaction through all relays, instead of stopping after first
failure.

When signing a transaction with Metamask, this also has the side-effect
of requiring the user to confirm signing the transaction again for each
such relay access attempt.

### DIFF
--- a/src/js/relayclient/RelayClient.js
+++ b/src/js/relayclient/RelayClient.js
@@ -51,10 +51,10 @@ class RelayClient {
     constructor(web3, config) {
         // TODO: require sign() or privKey
         //fill in defaults:
-        this.config = {
+        this.config = Object.assign( {
             validateCanRelay: true,
-            httpTimeout : DEFAULT_HTTP_TIMEOUT,
-        ...config }
+            httpTimeout : DEFAULT_HTTP_TIMEOUT
+        }, config )
 
         this.web3 = web3;
         this.httpSend = new HttpWrapper({ timeout: this.config.httpTimeout });
@@ -290,7 +290,7 @@ class RelayClient {
     async relayTransaction(encodedFunctionCall, options) {
 
         //validateCanRelay defaults (in config). to disable, explicitly set options.validateCanRelay=false
-        options = { validateCanRelay:this.config.validateCanRelay, ...options  }
+        options = Object.assign( { validateCanRelay:this.config.validateCanRelay }, options )
 
         var self = this;
         let relayRecipient = this.createRelayRecipient(options.to);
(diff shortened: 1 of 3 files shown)


src/js/relayclient/RelayClient.js
```diff
@@ -377,6 +386,28 @@ class RelayClient {
             }
             let relayMaxNonce = (await this.web3.eth.getTransactionCount(relayAddress)) + allowed_relay_nonce_gap;
 
+            //on first found relay, call canRelay to make sure that on-chain this request can pass
+            if ( options.validateCanRelay && firstTry ) {
+                firstTry = false;
```
shahafn: If the recipient is whitelisting relays, we can still have an error in case the first relay is whitelisted so `canRelay()` succeeds but the relay itself didn't respond, and we move on to the second relay which is not whitelisted - then we count on the relay to run it and return the correct response to client as it was before.

shahafn:
> > When signing a transaction with Metamask, this also has the side-effect
> > of requiring the user to confirm signing the transaction again for each
> > such relay access attempt.
> 
> Where does this come from? Don't relayers already receive the signature along with the relay request?

He means that when sending the request to the relay, it is first signed by the client and passed to the relay as part of the request, who then pass the signature as one of the parameters to `relayCall()`.
Since the signed data includes the relay address, it means that you need to sign a message dedicated for each relay, and with Metamask meaning a prompt to user for each signature.

drortirosh:
Before this PR, the client didn't actively checked the request, and trusted the relay to do so.
The client didn't distinguish between relay failures and on-chain failures - in any such case, it would continue to the next relay, just in case that relay could send the request.

However, if the cause for failure is on-chain, then the same error would repeat with all of them. This problem becomes a UX issue if using **metamask**, since the user is required to sign the request for each relay request.
With this PR, the on-chain **canRelay** check is done before calling the first relay, and if it fails, not only we don't call the relay at all, we also don't continue to get the same error for the next relays...

There is a very _slight_ chance that another relay could answer the request, but that's only if our RelayRecipient actively checks the _relay_ field in `canRelay`. But in that case, the  client shouldn't try all relays blindly, and should actively filter only the relays it is interested in (e.g. by URL)

PR #215, opened 2019-08-27
Added some slack to required gas of relayed call

shahafn:
Since the on-chain calculation is too tight. It seeems as if the size of the encodedFunction might be responsible for this, So I added a calculation per the yellowpaper for its gas cost and added it to the relayed tx's gas limit.

Also removed duplicate line in test removed unused params

### DIFF
--- a/server/src/librelay/relay_server.go
+++ b/server/src/librelay/relay_server.go
@@ -10,6 +10,7 @@ import (
 	"librelay/txstore"
 	"log"
 	"math/big"
+	"strings"
 	"sync"
 	"time"
 
@@ -134,12 +135,9 @@ type RelayServer struct {
 	Url                   string
 	Port                  string
 	RelayHubAddress       common.Address
-	StakeAmount           *big.Int
-	GasLimit              uint64
 	DefaultGasPrice       int64
 	GasPricePercent       *big.Int
 	PrivateKey            *ecdsa.PrivateKey
-	UnstakeDelay          *big.Int
 	RegistrationBlockRate uint64
 	EthereumNodeURL       string
 	gasPrice              *big.Int // set dynamically as suggestedGasPrice*(GasPricePercent+100)/100
@@ -169,12 +167,9 @@ func NewRelayServer(
 	Url string,
 	Port string,
 	RelayHubAddress common.Address,
-	StakeAmount *big.Int,
-	GasLimit uint64,
 	DefaultGasPrice int64,
 	GasPricePercent *big.Int,
 	PrivateKey *ecdsa.PrivateKey,
-	UnstakeDelay *big.Int,
 	RegistrationBlockRate uint64,
 	EthereumNodeURL string,
 	Client IClient,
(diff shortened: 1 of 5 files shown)


server/src/librelay/relay_server.go
```diff
@@ -355,9 +347,6 @@ func (relay *RelayServer) RegistrationDate() (when int64, err error) {
 	if (iter.Event == nil && !iter.Next()) ||
 		(bytes.Compare(iter.Event.Relay.Bytes(), relay.Address().Bytes()) != 0) ||
 		(iter.Event.TransactionFee.Cmp(relay.Fee) != 0) ||
-		(iter.Event.Stake.Cmp(relay.StakeAmount) < 0) ||
```
drortirosh: I'm trying to undertstand what did this line meant...

server/src/librelay/relay_server.go
```diff
@@ -486,6 +475,8 @@ func (relay *RelayServer) CreateRelayTransaction(request RelayTransactionRequest
 		log.Println(err)
 		return
 	}
+	// Since the on chain calculation is too tight and sometime insufficient, we need to give some slack
```
drortirosh: actually, the comment is not relevant, its not just adding slack, is properly calculating gas used.

server/src/librelay/relay_server_test.go
```diff
@@ -81,41 +82,41 @@ type TestServer struct {
 	*RelayServer
 }
 
-func (relay *TestServer) Stake(ownerKey *ecdsa.PrivateKey) (err error) {
-	tx, err := relay.sendStakeTransaction(ownerKey)
+func (relay *TestServer) Stake(ownerKey *ecdsa.PrivateKey, stakeAmount *big.Int, unstakeDelay *big.Int) (err error) {
```
drortirosh: a. not sure what you gained by removing the fields, if you now need to move them manually around to methods.
b. clearly its not a tiny cleanup, and doesn't fit into the gas-price fix.

server/src/librelay/relay_server_test.go
```diff
@@ -128,6 +129,8 @@ var client *TestClient
 var relayKey1 *ecdsa.PrivateKey
 var gaslessKey2 *ecdsa.PrivateKey
 var ownerKey3 *ecdsa.PrivateKey
+var unstakeDelay = big.NewInt(1 * 60 * 60 * 24 * 7) // 1 week in seconds
```
drortirosh: it bothers me to see these constants... did we ever test that a relay works with DIFFERENT values for stake delay/amount ?

drortirosh:
also, you're changing the semantics a bit. change the version to `0.4.1`

server/src/librelay/relay_server_test.go
shahafn: These methods are just in the tests, they aren't called anywhere else so no reason to keep fields just for methods defined and used only in a test file.

server/src/librelay/relay_server_test.go
shahafn: It looks like we do in `RelayHub.test.js`

server/src/librelay/relay_server.go
shahafn: I was validating the values of the staked event but it doesn't really add much and can be removed.

server/src/librelay/relay_server.go
shahafn: Fixed

shahafn:
> Approved too early. Need to update the version

Updated

drortirosh:
@nventuro : please upgrade your relays to this version (0.4.1).

PR #218, opened 2019-08-29
Gas cost test

shahafn:
Adding another test to the gas cost calculation of the relayed encoded function.

### DIFF
--- a/server/src/librelay/relay_server_test.go
+++ b/server/src/librelay/relay_server_test.go
@@ -506,6 +506,40 @@ func TestReuseNonceOnDevMode(t *testing.T) {
 	relay.DevMode = false
 }
 
+func TestTransactionTotalGasCost(t *testing.T) {
+	test.ErrFail(relay.TxStore.Clear(), t)
+	// We create two relayed txs of equal byte length, with the second tx having one more non-zero byte and one less zero byte
+	// meaning that the expected gas cost diff between them is 64 per the yellowpaper
+	expectedGasDiff := uint64(64)
+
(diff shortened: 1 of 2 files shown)


server/src/librelay/relay_server_test.go
```diff
@@ -506,6 +506,40 @@ func TestReuseNonceOnDevMode(t *testing.T) {
...
+	receipt1, err := client.TransactionReceipt(context.Background(), signedTx1.Hash())
+	test.ErrFailWithDesc(err, t, fmt.Sprint("Fetching transaction receipt for hash ", signedTx1.Hash()))
+	signedTx2, err := relay.CreateRelayTransaction(request2)
+	test.ErrFailWithDesc(err, t, "Creating relay transaction 2")
+	clk.IncrementBySeconds(60)
+	receipt2, err := client.TransactionReceipt(context.Background(), signedTx2.Hash())
+	test.ErrFailWithDesc(err, t, fmt.Sprint("Fetching transaction receipt for hash ", signedTx2.Hash()))
+
+	if  (getEncodedFunctionGas(request2.EncodedFunction).Uint64() - getEncodedFunctionGas(request1.EncodedFunction).Uint64() != expectedGasDiff){
+		test.ErrFail(errors.New("Wrong gas cost difference between encoded functions"), t)
```
drortirosh: don't you have `assert.equals` in go? that is, not only fail, but also show "expected" and "actual" values.

shahafn: Not in the standard libraries.

PR #219, opened 2019-08-30
Increase relay re-register frequency

nventuro:
Relays report themselves as down if they cannot find a `RelayRegisted` event in the last `delayBetweenRegistrations` (see `refreshBlockchainView()`). `keepAlive()` re-registers them, but its frequency is much lower than `refreshBlockChainView()` (60 time units vs 1), so relays experience downtime between the execution of `refreshBlockChainView()` and `keepAlive()`.

This PR increases the re-register frequency, so that `keepAlive()` makes sure the relay will not be de-registered before the execution of the next `keepAlive()`.

### DIFF
--- a/server/src/relay/RelayHttpServer.go
+++ b/server/src/relay/RelayHttpServer.go
@@ -36,6 +36,7 @@ var stopUpdatingPendingTxs chan bool
 var stopListeningToRelayRemoved chan bool
 
 var timeUnit time.Duration
+var keepAlivePeriod time.Duration
 
 var minimumRelayBalance = big.NewInt(1e17) // 0.1 eth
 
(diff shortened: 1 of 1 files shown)


drortirosh:
We think the problem is different:
we moved the code to calculate time difference in BLOCKS instead of time: the client searches for relay registration in the past 6000 blocks (roughly 24 hours in 14-second blockrate)
the relay should re-register every 5800 (the `RegistrationBlockRate` parameter)
The function `RegistrationDate()` looks only that many blocks back to find the registration time.
This means it can return a value which is at most 24hours back, and zero ("never") if last registration was before that.

The entire code that is based on "timestamp" is not in use: we just need to find a registration in the past 5800 blocks (giving the relay a slack of 30min) and re-register.

actually, RegistrationDate is now quite obsolete, since we don't need it in time-units anymore.
Shahaf is working on this.

nventuro:
Are you quite sure `RegistrationDate` is not being used? I can see the calls to both `refreshBlockchainView` and `keepAlive` on my logs, and both of those functions query the last registration date in the manner described above.

Switching to a blockrate can also work (though I suspect the diff will be larger), but even then the slack needs to be higher, since the issue occurs not due to a lack of slack, but due to _the slack being lower than the time between `keepAlive` calls_.

shahafn:
It is being used, but there was a discrepancy between the block rate and registration time, as we actually care more for block rate I changed it accordingly in #222. Please look as I think that would also solve this issue.

nventuro:
Closing in favor of #222

PR #222, opened 2019-09-03
Fixed registration date/block discrepancy

shahafn:


server/src/librelay/relay_server.go
```diff
@@ -324,38 +324,35 @@ func (relay *RelayServer) IsUnstaked() (removed bool, err error) {
 	return true, nil
 }
 
-func (relay *RelayServer) RegistrationDate() (when int64, err error) {
+func (relay *RelayServer) BlockCountSinceRegistration() (when uint64, err error) {
```
drortirosh: Block count is not a `when`, but `count`

server/src/librelay/relay_server.go
```diff
@@ -324,38 +324,35 @@ func (relay *RelayServer) IsUnstaked() (removed bool, err error) {
...
 	if err != nil {
 		log.Println(err)
 		return
 	}
-
+	// We only care about the last registration event
+	for iter.Next() {}
 	if (iter.Event == nil && !iter.Next()) ||
 		(bytes.Compare(iter.Event.Relay.Bytes(), relay.Address().Bytes()) != 0) ||
 		(iter.Event.TransactionFee.Cmp(relay.Fee) != 0) ||
```
drortirosh: If you check the txfee when checking for the last event, then it means a relay can't change it's txfee... Remove this check, as it adds nothing (you can argue whether a check for the url is required. I think it isn't, but at least, it's not harmful either, but txfee shouldn't be checked)

server/src/relay/RelayHttpServer.go
```diff
@@ -178,7 +178,7 @@ func parseCommandLine() (relayParams librelay.RelayParams) {
 	relayHubAddress := flag.String("RelayHubAddress", "0x537F27a04470242ff6b2c3ad247A05248d0d27CE", "RelayHub address")
 	defaultGasPrice := flag.Int64("DefaultGasPrice", int64(params.GWei), "Relay's default gasPrice per (non-relayed) transaction in wei")
 	gasPricePercent := flag.Int64("GasPricePercent", 10, "Relay's gas price increase as percentage from current average. GasPrice = (100+GasPricePercent)/100 * eth_gasPrice() ")
-	registrationBlockRate := flag.Uint64("RegistrationBlockRate", 5800, "Relay registeration rate (in blocks)")
+	RegistrationBlockRate = *flag.Uint64("RegistrationBlockRate", 5800, "Relay registeration rate (in blocks)")
```
drortirosh: Please put 6000-200 here, to reference the value the client checks.

drortirosh:
one thing I don't see is reducing the interval between checks

shahafn:
I did update it

server/src/librelay/relay_server.go
shahafn: I actually thought that if you change the tx fee then it makes sense to verify it in the event, why does it mean that you can't change it?

package.json
```diff
@@ -33,7 +33,7 @@
   "scripts": {
     "test": "npm run test-server && npm run test-js",
     "test-server": "make test-server",
-    "test-js": "run-with-testrpc --port 8544 --gasLimit 8000000 --defaultBalanceEther 1000 --deterministic 'npx truffle --network npmtest test'",
+    "test-js": "make build-server && run-with-testrpc --port 8544 --gasLimit 8000000 --defaultBalanceEther 1000 --deterministic 'npx truffle --network npmtest test'",
```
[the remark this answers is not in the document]
shahafn: Fixed

PR #230, opened 2019-09-23
Update web3 to 1.2.x

dmihal:
Background:
I'm using this library for my project (Burner Wallet 2), and I'm trying to reduce the size of the compiled bundle of the app. The largest cause of bundle inflation is coming from us bundling many different versions of web3, since different libraries all have different dependencies.

-----
This PR updates web3 to version 1.2, which is the LTS version. It also updates abi-decoder to bring its web3 version inline. `bignumber.js` is no longer bundled in with web3, so it is added to the devDependencies so the tests will pass.

### DIFF
--- a/package.json
+++ b/package.json
@@ -47,7 +47,7 @@
   },
   "dependencies": {
     "@0x/contracts-utils": "3.1.1",
-    "abi-decoder": "^1.2.0",
+    "abi-decoder": "^2.2.0",
     "axios": "^0.18.0",
     "big-js": "^3.1.3",
     "big.js": "^5.2.2",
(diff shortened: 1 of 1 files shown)


drortirosh:
@nventuro ? 
LGTM, but I think you're more versed with web3 versions.

dmihal:
Awesome! I'm guessing I need one of you guys to merge it? @nventuro @drortirosh

dmihal:
@drortirosh Will this be published to NPM soon?

drortirosh:
@dmihal : it is now published (as v0.4.2)

PR #255, opened 2019-12-03
add useGSN, approvalData options

drortirosh:
add missing transaction options:
- useGSN - set to `false` to bypass the GSN network for this specific transaction.
- approvalData - set the `approvalData` value of the transaction. This value can be checked by the `RelayRecipient` to verify the call (e.g. some sort of signature over tx content, sender, etc)

### DIFF
--- a/src/js/relayclient/RelayProvider.js
+++ b/src/js/relayclient/RelayProvider.js
@@ -67,9 +67,6 @@ class RelayProvider {
         let ret=false
         if ( payload.params && payload.params[0] ) {
             let useGSN = payload.params[0].useGSN
-            if ( useGSN ) {
-                console.log( "tada")
-            }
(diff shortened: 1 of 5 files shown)


src/js/relayclient/RelayProvider.js
```diff
@@ -64,8 +64,25 @@ class RelayProvider {
 
     //hook method: skip relay if the "from" address appears in optins.skipSenders
     skipRelay(payload) {
-        return !this.relayOptions.isRelayEnabled ||
-            this.relayOptions.skipSenders && this.relayOptions.skipSenders[payload.params.from]
+        let ret=false
+        if ( payload.params && payload.params[0] ) {
+            let useGSN = payload.params[0].useGSN
+            if ( useGSN ) {
+                console.log( "tada")
```
shahafn: remove..?

PR #260, opened 2019-12-16
Deep import eth-crypto/signer

dmihal:
By importing only the `sign` function instead of the entire ETHCrypto library, the bundle size can be significantly decreased.


src/js/relayclient/utils.js
```diff
@@ -92,7 +92,7 @@ module.exports = {
         else
             signed = hash
         let keyHex = "0x" + Buffer.from(privKey).toString('hex')
-        const sig_ = EthCrypto.sign(keyHex, signed)
+        const sig_ = sign(keyHex, signed)
```
shahafn: I believe that we can actually do `const sig_ = ethUtils.ecsign(signed, keyHex)` and remove eth-crypto altogether.

PR #263, opened 2019-12-23
Fix case-sensitive address as string comparison

forshtat:


src/js/relayclient/RelayClient.js
```diff
@@ -122,7 +122,7 @@ class RelayClient {
         let transaction_orig_params_hash = utils.getTransactionHash(
             from, to, transaction_orig, transaction_fee, gas_price, gas_limit, nonce, relay_hub_address, relay_address);
 
-        if (returned_tx_params_hash === transaction_orig_params_hash && address_relay === signer) {
+        if (returned_tx_params_hash === transaction_orig_params_hash && utils.isSameAddress(address_relay, signer)) {
```
drortirosh: while is correct, how you hit this problem:
`address_relay` is returned by our relay, which return it in all lowercase.
`signer` is the return of `ethUtils.bufferToHex`

The only way to hit this wrong check is if you have a different implementation of a relay (e.g. a mock), and return something that is not all lowercase...

alex-forshtat-tbk: I did use the mock relay from the sample. Still, the interface does not define addresses as case-sensitive, so alternative relays would break.

PR #279, opened 2020-02-06
Improvements

forshtat:


server/src/librelay/tbk_ethclient.go
```diff
@@ -12,14 +12,14 @@ import (
 /* We hook ethclient since EstimateGas returns inaccurate estimation, due to state differences
  * of executing a tx locally through eth_call vs on the blockchain ( msg.sender.balance for example)
 */
-type TbkClient struct {
```
shahafn: Like the fact that you renamed it, how about renaming the file as well?

test/relay_client_test.js
```diff
@@ -709,7 +709,20 @@ contract('RelayClient', function (accounts) {
...
+
+  it('should fail to relay if provided Gas Sponsor and Relay Recipient do not use same Relay Hub', async function () {
+    // TODO: all tests rely on 'SampleRecipient.web3.setProvider(relayProvider)' being called in other test!!!
+    const recipient = await SampleRecipient.deployed()
+    await recipient.setHub(accounts[4], { from: accounts[0], useGSN: false })
+    try {
+      await recipient.emitMessage("ain't gonna work mate", { from: accounts[0], gasSponsor: gasSponsor.address })
+      assert.fail()
+    } catch (error) {
+      assert.equal(true, error.message.includes('Sponsor and recipient RelayHub addresses do not match'),
```
drortirosh: There is `assert.includes` which reports the error nicer than `assert.ok`

src/js/relayclient/RelayClient.js
```diff
@@ -299,6 +299,15 @@ class RelayClient {
     const gasSponsor = options.gasSponsor || options.to
     const relayHub = await this.createRelayHubFromSponsor(gasSponsor)
 
+    // TODO: refactor! wrong instance is created for accidentally same method!
+    if (!utils.isSameAddress(gasSponsor, options.to)) {
+      const recipientHub = await this.createGasSponsor(options.to).methods.getHubAddr().call()
+
+      if (!utils.isSameAddress(relayHub._address, recipientHub)) {
+        throw Error('Sponsor and recipient RelayHub addresses do not match')
```
drortirosh: The error message should be "wrong sponsor: uses different RelayHub than contact'

contracts/utils/EIP712Sig.sol
```diff
@@ -39,10 +39,13 @@ contract EIP712Sig {
         "EIP712Domain(string name,string version,address verifyingContract)"
     );
 
+    // solhint-disable-next-line max-line-length
```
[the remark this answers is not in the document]
forshtat: https://github.com/protofire/solhint/issues/187

src/js/relayclient/RelayClient.js
```diff
@@ -299,6 +299,15 @@ class RelayClient {
     const gasSponsor = options.gasSponsor || options.to
     const relayHub = await this.createRelayHubFromSponsor(gasSponsor)
 
+    // TODO: refactor! wrong instance is created for accidentally same method!
```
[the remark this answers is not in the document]
forshtat: ```this.createGasSponsor``` is a wrong method to call, it is given an 'options.to', which is not necessarily implementing 'IGasSponsor'. I added the same ```getHubAddr``` method to ```BaseRelayRecipient``` to get a hub.
Basically, there should be ```IRelayRecipient``` with a single ```getHubAddr``` method in it, and ```this.createRelayRecipient``` method instead, but I thought this one line wasn't worth it.

src/js/relayclient/RelayClient.js
forshtat: Because you prefer your error messages obscure and incorrect?

test/relay_client_test.js
forshtat: Right. Нou live and you learn. I might have one more like this... Or, maybe, like, thirteen. Fixed.

contracts/utils/EIP712Sig.sol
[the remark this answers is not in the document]
drortirosh: How about:
```
    string constant CALLDATA_SIG = "CallData(address target,uint256 gasLimit,uint256 gasPrice,bytes encodedFunction)";
    string constant RELAYDATA_SIG = "RelayData(address senderAccount,uint256 senderNonce,address relayAddress,uint256 pctRelayFee,address gasSponsor)";
    string constant RELAYREQUEST_SIG = "RelayRequest(CallData callData,RelayData relayData)";

    bytes32 public constant RELAY_REQUEST_TYPEHASH = keccak256(abi.encodePacked(
            RELAYREQUEST_SIG,
            CALLDATA_SIG,
            RELAYDATA_SIG
        ));

    bytes32 public constant CALLDATA_TYPEHASH = keccak256(bytes(CALLDATA_SIG));

    bytes32 public constant RELAYDATA_TYPEHASH = keccak256(bytes(RELAYDATA_SIG));
```

src/js/relayclient/RelayClient.js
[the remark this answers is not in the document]
drortirosh: I agree. There should be a `IRelayRecipient` has 2 methods: the `getHubAddr() public` and  `getSender() internal`: its an interface that defines a behaviour (which also explains about setting the hub address, though there's no interface method for that) and the baseclass is merely a suggested implementation.

src/js/relayclient/RelayClient.js
drortirosh: The error messages are not for us, but for the end-users - should explain the error, and preferably, when went wrong. The problem is that probably the user doesn't know when sponsor (or even "gasPayer") is - he might know the specific implementation.
he might also not know what we call "the contract" - but we don't have a way to give them better names...

contracts/utils/EIP712Sig.sol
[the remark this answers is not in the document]
forshtat: You could, but I don't like ```encodePacked``` used as a string concatenator for some reason. This contract is not exactly visible, and lint is more of an advice than law, you know...

src/js/relayclient/RelayClient.js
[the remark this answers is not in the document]
forshtat: Interfaces cannot define non-external methods. I have this in my TODO list but with low priority for now.

src/js/relayclient/RelayClient.js
forshtat: Still, if you don't know what is going on no amount of error messages will teach you everything, and if you have even a vogue idea this error should be clear enough.

PR #280, opened 2020-02-09
Sendbatch

drortirosh:
A utility function for creating batch transactions, to be used by RelayRecipients.

The relay themselves don't use it: a relay process client request immediately, and can't wait for another client to send a request and batch them together.


contracts/RelayHub.sol
```diff
@@ -403,17 +404,25 @@ contract RelayHub is IRelayHub {
 
         atomicData.preReturnValue = abi.decode(retData, (bytes32));
 
+        bool abortOnFirstRevert = true;
+        (uint count, bytes memory error) = CallBatcher.sendBatch(
```
drortirosh: commenting out the call to `sendBatch` makes the compilation work.
the same call exists in the TestBatch.sol, so not sure why it fails the compiler here.

PR #283, opened 2020-02-14
Variable gas limits for sponsor calls & variable calldata size compensation

forshtat:


.idea/dictionaries/alexf.xml
```diff
@@ -3,13 +3,18 @@
     <words>
       <w>blockchain</w>
       <w>calldata</w>
+      <w>calldatagascost</w>
```
shahafn: Why is this file indexed?

contracts/BaseGasSponsor.sol
```diff
@@ -9,15 +9,33 @@ import "./interfaces/IRelayHub.sol";
...
  * A subclass must implement:
- *  - acceptRelayCall
+ *  - acceptRelayedCall
  *  - preRelayedCall
  *  - postRelayedCall
  */
 contract BaseGasSponsor is IGasSponsor, Ownable {
 
+    // Gas stipends for acceptRelayedCall, preRelayedCall and postRelayedCall
+    uint256 constant private ACCEPT_RELAYED_CALL_MAX_GAS = 50000;
```
shahafn: Why are they private?

contracts/RelayHub.sol
```diff
@@ -288,42 +334,36 @@ contract RelayHub is IRelayHub {
...
-                relayRequest.relayData.senderAccount,
-                relayRequest.callData.target,
-                sponsor,
-                functionSelector,
-                preconditionCheck);
-            return;
+        GSNTypes.SponsorLimits memory sponsorGasLimits;
+        {
+            uint256 maxPossibleGas;
+            (maxPossibleGas, sponsorGasLimits) = getValidGasLimits(relayRequest.gasData, relayRequest.relayData.gasSponsor);
```
shahafn: The name `getValidGasLimits` is confusing, since it's not only a getter, but also a validator, which its name doesn't necessarily imply ( I understand that it contains "valid" but it still looks like just a getter).

contracts/RelayHub.sol
```diff
@@ -288,42 +334,36 @@ contract RelayHub is IRelayHub {
...
+
+            // We now verify the legitimacy of the transaction (it must be signed by the sender, and not be replayed),
+            // and that the sponsor will agree to be charged for it.
+            uint256 preconditionCheck;
```
shahafn: Not sure if this is the PR to do so, but you already renamed several vars here - the name `preconditionCheck` is just bad, it should simply be `canRelayStatus`.

contracts/RelayHub.sol
```diff
@@ -398,40 +442,40 @@ contract RelayHub is IRelayHub {
...
             IGasSponsor(address(0)).preRelayedCall.selector, recipientContext
         );
-
-        // preRelayedCall may revert, but the recipient will still be charged: it should ensure in
-        // acceptRelayedCall that this will not happen.
-        (bool success, bytes memory retData) = relayRequest.relayData.gasSponsor.call.gas(PRE_RELAYED_CALL_MAX_GAS)(data);
-
-        if (!success) {
-            revertWithStatus(RelayCallStatus.PreRelayedFailed);
+        bytes memory retData;
```
shahafn: Is there a reason to define retData outside and success inside the new block? it seems that you could move the line `atomicData.preReturnValue = abi.decode(retData, (bytes32));` inside the block as well, and you don't use retData elsewhere.

contracts/RelayHub.sol
```diff
@@ -458,39 +502,12 @@ contract RelayHub is IRelayHub {
...
-        uint256 gasPrice,
-        uint256 transactionFee)
-    public
-    view
-    returns (uint256) {
-        return calculateCharge(requiredGas(relayedCallStipend), gasPrice, transactionFee);
-    }
-
-    function calculateCharge(uint256 gas, uint256 gasPrice, uint256 fee) private pure returns (uint256) {
+    function calculateCharge(uint256 gas, uint256 gasPrice, uint256 fee) public view returns (uint256) {
```
shahafn: why is it view and not pure?

server/src/librelay/relay_server.go
```diff
@@ -460,13 +461,60 @@ func (relay *RelayServer) CreateRelayTransaction(request RelayTransactionRequest
...
+	maxPossibleGas := big.NewInt(0)
+	maxPossibleGas.Add(maxPossibleGas, hubOverhead)
+	maxPossibleGas.Add(maxPossibleGas, gasLimits.AcceptRelayedCallGasLimit)
+	maxPossibleGas.Add(maxPossibleGas, gasLimits.PreRelayCallGasLimit)
+	maxPossibleGas.Add(maxPossibleGas, &request.GasLimit)
+	maxPossibleGas.Add(maxPossibleGas, gasLimits.PostRelayCallGasLimit)
+
+	// TODO: well, this sucks! I need to calculate max possible gas using entire msg.data!!!
+	//  In the meantime, these 400'000 extra gas should help :-)
+	maxPossibleGas.Add(maxPossibleGas, big.NewInt(400000))
```
shahafn: Not sure what's that

server/src/librelay/relay_server_test.go
```diff
@@ -545,18 +546,18 @@ func TestReuseNonceOnDevMode(t *testing.T) {
 	// Clean up for the next test
 	relay.DevMode = false
 }
-
+/*
```
shahafn: If you think it's not meaningful, just remove it.

test/RelayHubPenalizations.test.js
```diff
@@ -11,13 +11,15 @@ const rlp = require('rlp')
 
 const { expect } = require('chai')
 
-contract('RelayHub Penalizations', function ([_, relayOwner, relay, otherRelay, sender, other]) { // eslint-disable-line no-unused-vars
+const GTXDATANONZERO = 68
+
+contract.skip('RelayHub Penalizations', function ([_, relayOwner, relay, otherRelay, sender, other]) { // eslint-disable-line no-unused-vars
```
shahafn: Why is it skipped?

.idea/dictionaries/alexf.xml
forshtat: So it can be checked out. I do look at the spelling errors in code and comments. Also, though the files are per-user, idea seems to take all dictionaries into account so everyone is invited to check in theirs.

contracts/BaseGasSponsor.sol
forshtat: Well, there is a getter right there, why wouldn't you use it?

contracts/RelayHub.sol
forshtat: Can you come up with a good name for it?

contracts/RelayHub.sol
forshtat: Will do next thing

contracts/RelayHub.sol
forshtat: No, no reason whatsoever. Not a bug though, so I guess does not require fixing, what do you think?

contracts/RelayHub.sol
forshtat: It is allowed for a RelayHub (in the future) to use some storage variable to affect the charge. Wanted to stress that out.

server/src/librelay/relay_server.go
forshtat: Hack to pass a big enough 'maxPossibleGas' to the ARC method. What I want is
```relayHub.methods.relayCall(...).encodeABI().length()```
but I don't know how to achieve that in Go and it is too late to put effort to it.

File: server/src/librelay/relay_server_test.go (hunk 8 above)
Comment (forshtat) [reply server/src/librelay/relay_server_test.go]: It is a go server test... Will remove it soon enough I hope :-)

File: test/RelayHubPenalizations.test.js (hunk 9 above)
Comment (forshtat) [reply test/RelayHubPenalizations.test.js]: Because tests fail for me locally (EIP-155 issue) and I don't want to run these tests even if they pass on CircleCI without fixing it, as they depend solely on having a very old ganache to pass.

File: contracts/BaseGasSponsor.sol [hunk 10]
```diff
@@ -9,15 +9,33 @@ import "./interfaces/IRelayHub.sol";
...
+    uint256 constant private PRE_RELAYED_CALL_MAX_GAS = 100000;
+    uint256 constant private POST_RELAYED_CALL_MAX_GAS = 110000;
+
     /// The IRelayHub singleton which is allowed to call us
     IRelayHub internal relayHub;
 
+    function getGasLimitsForSponsorCalls()
```
drortirosh: this is a method we expect people to override, right?
If so, we should document it as such..

contracts/interfaces/IRelayHub.sol
```diff
@@ -256,5 +236,9 @@ interface IRelayHub {
     function penalizeIllegalTransaction(bytes calldata unsignedTx, bytes calldata signature) external;
 
     function getNonce(address from) external view returns (uint256);
+
+    function getHubOverhead() external view returns (uint256);
```
drortirosh: ?

contracts/test/TestSponsorEverythingAccepted.sol
```diff
@@ -32,13 +32,18 @@ contract TestSponsorEverythingAccepted is BaseGasSponsor {
     }
 
     function postRelayedCall(
-        bytes calldata context, bool success, uint actualCharge, bytes32 preRetVal
+        bytes calldata context,
+        bool success,
+        bytes32 preRetVal,
+        uint256 gasUseWithoutPost,
+        uint256 txFee,
```
drortirosh: what's txFee ? how was it calculated?

contracts/utils/GSNTypes.sol
```diff
@@ -0,0 +1,31 @@
...
+    struct RelayRequest {
+        address target;
+        bytes encodedFunction;
+        GasData gasData;
+        RelayData relayData;
+    }
+
+    struct SponsorLimits {
+        uint256 acceptRelayedCallGasLimit;
+        uint256 preRelayCallGasLimit;
```
drortirosh: preRelaye**d**Call

contracts/utils/GSNTypes.sol
```diff
@@ -0,0 +1,31 @@
+pragma solidity ^0.5.16;
+
+library GSNTypes {
+
+    struct GasData {
+        uint256 gasLimit;
+        uint256 gasPrice;
+        uint256 pctRelayFee;
+        uint256 baseRelayFee;
```
drortirosh: I didnt' see anywhere you use the `baseRelayFee`

forshtat: It is not yet supported. But I was already changing the signature format, so...

contracts/utils/GSNTypes.sol
forshtat: fixed

contracts/test/TestSponsorEverythingAccepted.sol
forshtat: this is the old 'pctRelayFee'; will be replaced with 'GasData' struct next

contracts/interfaces/IRelayHub.sol
forshtat: The 'canRelay' now takes parameter 'maxPossibleGas', and it depends on the overhead of the hub. As we plan for upgradeability, the overhead should be exposed by the Hub.

contracts/BaseGasSponsor.sol
forshtat: fixed.

test/RelayHubPenalizations.test.js
drortirosh: what "old ganache"? we `yarn install` our ganache version.
either show why this method is redundant with another test - or let's fix it.

test/RelayHubGasCalculations.test.js
```diff
@@ -122,12 +122,12 @@ contract('RelayHub gas calculations', async function ([_, relayOwner, relayAddre
...
         // Magic numbers seem to be gas spent on calldata. I don't know of a way to calculate them conveniently.
         await expectEvent.inTransaction(tx, TestSponsorVariableGasLimits, 'SampleRecipientPreCallWithValues', {
-          gasleft: (gasData.preRelayCallGasLimit - magicNumbers.pre).toString(),
+          gasleft: (gasData.preRelayedCallGasLimit - magicNumbers.pre).toString(),
           arcGasleft: (gasData.acceptRelayedCallGasLimit - magicNumbers.arc).toString(),
           maxPossibleGas: gasData.maxPossibleGas.toString()
         })
         await expectEvent.inTransaction(tx, TestSponsorVariableGasLimits, 'SampleRecipientPostCallWithValues', {
-          gasleft: (gasData.postRelayCallGasLimit - magicNumbers.post).toString()
+          gasleft: (gasData.postRelayedCallGasLimit - magicNumbers.post).toString()
```
drortirosh: sorry to nag on unrelated issues, but such "magic numbers" should *at least* be used in several separate tests, which takes different params.
Otherwise, this test is meaningless

test/RelayHubPenalizations.test.js
forshtat: We yarn install the old ganache version. I can enable these tests on CircleCI, but they will likely fail locally if you run a different ganache. I don't want tests flickering depending on the ganache version in the repo.
This "method" is not redundant. This "method" is all tests of penalizations. 
Penalizations are completely broken as described in detail.

test/RelayHubGasCalculations.test.js
forshtat: Well, this test is semi-meaningless. I want the value to be 50'000, but instead, it is 49'123, so I know the magic 877 gas was used for calldata and stuff. If instead of 49'123 I one day get 12, of "potato", the test will fail and catch a regression.

contracts/RelayHub.sol
```diff
@@ -79,8 +82,13 @@ contract RelayHub is IRelayHub {
 
     EIP712Sig private eip712sig;
 
-    constructor () public {
+    constructor (uint256 _gtxdatanonzero) public {
```
drortirosh: our first RelayHub constructor param...
- I thought it is "16". you write "68" everywhere
-  what is the source of the value? its left undocumented. 
- should be written in one place, not spread all over the deployment scripts, tests
- you didn't update "singleton/deploy" script

PR #284, opened 2020-02-16
Lower-cost relays

drortirosh:
### Purpose:
Don't require relays to re-register. once is enough.

### New relay lookup mechanism:
- look for all relay events (not only RelayAdded) in the past X blocks (~30 days)
- after finding unique relays, look up for latest RelayAdded for each one of them.

### API change
- client can specify `preferredRelays` (urls)
- these relays are always added at the beginning of the relays list (only checked for "ping", not txfee/stake/delay)
- only if none of the preferred relays answer, it will continue with the rest of the "public" relays

### DIFF
--- a/server/src/librelay/relay_server.go
+++ b/server/src/librelay/relay_server.go
@@ -92,7 +92,7 @@ type IRelay interface {
 
 	IsUnstaked() (removed bool, err error)
 
-	BlockCountSinceRegistration() (when uint64, err error)
+	BlockCountSinceLastEvent() (when uint64, err error)
 
 	GetRegistrationBlockRate() (rate uint64)
 
@@ -351,7 +351,8 @@ func (relay *RelayServer) IsUnstaked() (removed bool, err error) {
 	return true, nil
 }
 
-func (relay *RelayServer) BlockCountSinceRegistration() (count uint64, err error) {
+//find last TransactionRelayed or RelayAdded
+func (relay *RelayServer) BlockCountSinceLastEvent() (count uint64, err error) {
 	lastBlockHeader, err := relay.Client.HeaderByNumber(context.Background(), nil)
 	if err != nil {
 		log.Println(err)
@@ -380,7 +381,24 @@ func (relay *RelayServer) BlockCountSinceRegistration() (count uint64, err error
 		(iter.Event.Url != relay.Url) {
 		return 0, fmt.Errorf("Could not receive RelayAdded events for our relay")
 	}
-	count = lastBlockNumber - iter.Event.Raw.BlockNumber
+	blockNumber := iter.Event.Raw.BlockNumber
+
+	//Now find also last TransactionRelayed request, and use the latest of these:
+	iter1, err1 := relay.rhub.FilterTransactionRelayed(filterOpts, []common.Address{relay.Address()}, nil, nil)
+	if err1 != nil {
+		log.Println(err1)
+		return
+	}
+	// We only care about the last event
+	for iter1.Next() {
+	}
+	if (iter1.Event != nil &&
+		(bytes.Compare(iter1.Event.Relay.Bytes(), relay.Address().Bytes()) == 0) &&
+		iter1.Event.Raw.BlockNumber > blockNumber) {
+        blockNumber = iter1.Event.Raw.BlockNumber
+    }
+
+	count = lastBlockNumber - blockNumber
 	return
 }
 
--- a/src/js/relayclient/ServerHelper.js
+++ b/src/js/relayclient/ServerHelper.js
@@ -1,18 +1,24 @@
 const BN = require('web3').utils.toBN
 
+const { event2topic } = require('./utils')
 // relays are "down-scored" in case they timed out a request.
 // they are "forgiven" after this timeout.
 const DEFAULT_RELAY_TIMEOUT_GRACE_SEC = 60 * 30
 
+// assume veri high stake values for preferred relays, so we don't need
+// to change the filter logic to always support them.
+const PREFERRED_STAKE = 1e30
+const PREFERRED_DELAY = 1e10
 class ActiveRelayPinger {
   // TODO: 'httpSend' should be on a network layer
-  constructor (filteredRelays, httpSend, gasPrice, verbose) {
+  constructor (filteredRelays, httpSend, gasPrice, verbose, nextPinger) {
     this.remainingRelays = filteredRelays.slice()
     this.httpSend = httpSend
     this.pingedRelays = 0
     this.relaysCount = filteredRelays.length
     this.gasPrice = gasPrice
     this.verbose = verbose
+    this.nextPinger = nextPinger
   }
 
   /**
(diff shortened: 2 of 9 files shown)


server/src/librelay/relay_server.go
```diff
@@ -372,7 +373,24 @@ func (relay *RelayServer) BlockCountSinceRegistration() (count uint64, err error
...
+	blockNumber := iter.Event.Raw.BlockNumber
+
+	//Now find also last TransactionRelayed request, and use the latest of these:
+	iter1, err1 := relay.rhub.FilterTransactionRelayed(filterOpts, []common.Address{relay.Address()}, nil, nil)
+	if err1 != nil {
+		log.Println(err1)
+		return
+	}
+	// We only care about the last event
+	for iter1.Next() {
```
shahafn: Are you sure about the order when getting events?

drortirosh: I don't care. I actually find the last block number of "RelayAdded", the last block of "TransactionRelayed" - and use the higher of these.
If you look at the tests, the RelayAdded is tested when creating the relay.
Then after sending a transaction, I check again, and verify that it is indeed one block into the past (the RelayAdded is 3-4 blocks down at this point)

src/js/relayclient/ServerHelper.js
```diff
@@ -1,29 +1,40 @@
...
-      return null
+    let ret = await this.nextRelay1()
+    // only if no relay found in this ActivePinger, start using the next ActivePinger
+    if (!ret && this.nextPinger) {
+      ret = await this.nextPinger.nextRelay()
     }
+    return ret
+  }
 
+  async nextRelay1 () {
```
forshtat: well, 'nextRelay1' is not a legit name for a function

src/js/relayclient/ServerHelper.js
```diff
@@ -1,29 +1,40 @@
...
   // TODO: 'httpSend' should be on a network layer
-  constructor (filteredRelays, httpSend, gasPrice, verbose) {
+  constructor (filteredRelays, httpSend, gasPrice, verbose, nextPinger) {
     this.remainingRelays = filteredRelays.slice()
     this.httpSend = httpSend
     this.pingedRelays = 0
     this.relaysCount = filteredRelays.length
     this.gasPrice = gasPrice
     this.verbose = verbose
+    this.nextPinger = nextPinger
```
forshtat: What is 'nextPinger' supposed to be? Why did you add it?

src/js/relayclient/ServerHelper.js
```diff
@@ -1,29 +1,40 @@
...
    * Ping those relays that were not returned yet. Remove the returned relay (first to respond) from {@link remainingRelays}
    * @returns the first relay to respond to a ping message. Note: will never return the same relay twice.
    */
   async nextRelay () {
-    if (this.remainingRelays.length === 0) {
-      return null
+    let ret = await this.nextRelay1()
+    // only if no relay found in this ActivePinger, start using the next ActivePinger
+    if (!ret && this.nextPinger) {
+      ret = await this.nextPinger.nextRelay()
```
forshtat: Weird

src/js/relayclient/ServerHelper.js
```diff
@@ -186,51 +201,92 @@ class ServerHelper {
...
+      const prefs = this.preferredRelays.map(relayUrl => ({
+        relayUrl,
+        transactionFee: 0,
+        stake: PREFERRED_STAKE,
+        unstakeDelay: PREFERRED_DELAY
+      })
+      )
+      // if there are preferred relays, create a pinger that uses them FIRST, and only
+      // falls-back to the "normal" pinger, based on the filtered list.
+      pinger = new ActiveRelayPinger(prefs, httpSend, gasPrice, verbose, pinger)
```
forshtat: Create a new object. But wait... We didn't check for preferredRelays. Let's create a different one...

src/js/relayclient/ServerHelper.js
```diff
@@ -186,51 +201,92 @@ class ServerHelper {
...
-          relayUrl: args.url,
-          transactionFee: args.transactionFee,
-          stake: args.stake,
-          unstakeDelay: args.unstakeDelay
-        }
-        relay.score = this.calculateRelayScore(relay)
-        activeRelays[args.relay] = relay
-      } else if (event.event === 'RelayRemoved') {
-        delete activeRelays[event.returnValues.relay]
+    const foundRelays = {}
```
forshtat: There is a 'Set' built-in class, makes it more clear what this variable is

forshtat:
```
> const s = new Set()
Set {}
> s.add("aaa")
Set { 'aaa' }
> s.add("aaa")
Set { 'aaa' }
> s.add("aaaa")
Set { 'aaa', 'aaaa' }
> s.delete('aaa')
true
> s
Set { 'aaaa' }

File: src/js/relayclient/ServerHelper.js (hunk 2 above)
Comment (drortirosh) [reply src/js/relayclient/ServerHelper.js]: yes, it should be `nextRelayInternal`, so that the logic of chaining a pinger can be outside this method (above)

File: src/js/relayclient/ServerHelper.js (hunk 4 above)
Comment (drortirosh) [reply src/js/relayclient/ServerHelper.js]: I wanted the pinger mechanism to run first on the list of "preferredRelays", and only if none is found, start with the normal relays.
At first, I simply put the preferred relays at the beginning of the list. But then it occurred to me that if there's exactly 1 (or two) preferred ones, and we ping 3 simultaneously, you might start using "non-preferred" relay just because it answered the ping earlier.
So instead, I added this chaining of pingers.

File: src/js/relayclient/ServerHelper.js (hunk 5 above)
Comment (drortirosh) [reply src/js/relayclient/ServerHelper.js]: please explain your pain-point, and suggest a change.

File: src/js/relayclient/ServerHelper.js (hunk 6 above)
Comment (drortirosh) [reply src/js/relayclient/ServerHelper.js]: good idea.

File: src/js/relayclient/ServerHelper.js (hunk 5 above)
Comment (forshtat) [reply src/js/relayclient/ServerHelper.js]:
``` if (preferredRelays) {
    return new ActiveRelayPinger(bla)
} else {
    return new ActiveRelayPinger(blabla)
}```
makes much more sense. I don't understand the need to wrap one ActiveRelayPinger inside another ActiveRelayPinger,

File: src/js/relayclient/ServerHelper.js (hunk 2 above)
Comment (forshtat) [reply src/js/relayclient/ServerHelper.js]: Well, it is still `nextRelay1` after the last commit

File: src/js/relayclient/ServerHelper.js (hunk 4 above)
Comment (forshtat) [reply src/js/relayclient/ServerHelper.js]: In this case, I would consider a different approach
Existing `ActiveRelayPinger` tries the first 3 relays because it is told to do so. If it could differentiate between 'preferred' and 'fallback' relays, it would be able to contain the etire business logic of it's class, instead of spreading it between two nested instances of itself.
Consider smth like:
```    let slice = this.preferredRelays.slice(0, bulkSize)
    if (slice.length === 0){
       slice = this.fallbackRelays.slice(0, bulkSize)
```

src/js/relayclient/ServerHelper.js
drortirosh: but that's not the case. its not either X or Y:
 its " ActiveRelayPinger(X), and if they all fail, ActiveRelayPinger(Y)."

I can achieve this either by a "wrapper" ActiveRelayPinger that do it, or that every pinger can have a "next" pinger to fallback to it, once all the relays it manages by itself fail.

src/js/relayclient/ServerHelper.js
drortirosh: that's not enough: after trying a relay, we need to remove it from the list, and try with the rest. this logic on handling "remainingRelays" need to be duplicated into "preferredRelays".

I think its more duplicated code than adding nextPinger

PR #287, opened 2020-03-03
Fix EIP-155 signatures before 'penalize'; create "environment" file; move to Istanbul

forshtat:


contracts/utils/ECDSAEIP155.sol
```diff
@@ -0,0 +1,92 @@
...
+        // with 0xFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFEBAAEDCE6AF48A03BBFD25E8CD0364141 - s1 and flip v from 27 to 28 or
+        // vice versa. If your library also generates signatures with 0/1 for v instead 27/28, add 27 to v to accept
+        // these malleable signatures as well.
+        if (uint256(s) > 0x7FFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF5D576E7357A4501DDFE92F46681B20A0) {
+            return address(0);
+        }
+
+        // EIP-155 defines a rule of setting "v += CHAIN_ID * 2 + 8"
+        // In case the value of V seems to match this value, use the signature as EIP-155
+        if (v > 28) {
```
drortirosh: why do you have to check in the library? if this calculation is all you need to do, then you can do it just before calling the library

src/js/relayclient/Environments.js
```diff
@@ -0,0 +1,13 @@
+/**
```
drortirosh: its a good idea to define "Environments" outside of actual code and tests.
its bad idea to scatter `istanbul` everywhere - it should be a single value here, at the environment
(e.g. all should access `Environments.default`

contracts/utils/ECDSAEIP155.sol
forshtat: I would need to have the decode and encode of the signature to components in the hub only to pass it to the library

package.json
```diff
@@ -33,7 +33,7 @@
   "scripts": {
     "test": "npm run test-server && npm run test-js",
     "test-server": "make test-server",
-    "test-js": "make build-server && run-with-testrpc --hardfork 'constantinople' --port 8544 --gasLimit 10000000 --defaultBalanceEther 1000 --deterministic --allowUnlimitedContractSize 'npx truffle --network npmtest test'",
+    "test-js": "make build-server && npx run-with-testrpc --hardfork 'istanbul' --port 8544 --gasLimit 10000000 --defaultBalanceEther 1000 --deterministic --allowUnlimitedContractSize 'npx truffle --network npmtest test'",
```
drortirosh: you don't need "npx" it: scripts always run with `node_modules/.bin` in their path (otherwise, `yarn test` wouldn't have worked..)

contracts/utils/ECDSAEIP155.sol
drortirosh: why ? You will have to modify a single byte in-place: bytes is a modifyable array.

drortirosh:
please update title, description as the issue and fix

PR #289, opened 2020-03-10
Nonview acceptRelayedCall

drortirosh:
make `acceptRelayedCall` *nonview* function.
added `dryRun` utility method, that can run a method call, only to check if a method doesn't revert.
`DryRunSponsor` uses it, as it checks if `encodedFunction` would succeed


forshtat:
@drortirosh Should I close this PR?

PR #296, opened 2020-03-15
fixed deployment script

drortirosh:
support new constructor parameter
(and bypass "flattener" unsupported "pragma experimental")


scripts/singleton/compute.sh
```diff
@@ -28,7 +28,7 @@ rm -rf "$output_dir" # Delete possible remains from previous run
 mkdir -p "$output_dir"
 
 echo "Flattening source files into a single file"
-npx truffle-flattener contracts/RelayHub.sol > $output_dir/RelayHub.flattened.sol
+( echo "pragma experimental ABIEncoderV2;"; npx truffle-flattener contracts/RelayHub.sol |grep -v 'pragma experimental ABIEncoderV2') > $output_dir/RelayHub.flattened.sol
```
forshtat: could you explain what is going on here?

scripts/singleton/get-deploy-data.js
```diff
@@ -1,19 +1,24 @@
...
 
+// gas price for nonzero byte. 68 pre-instanbul, 16 >= istanbul
+const gtxdatanonzero = 68
+const gasLimit = 5400000
+const gasPrice = 100000000000 /// 100 gigawei
+
 const tx = new Transaction({
   nonce: 0,
-  data: '0x' + bin,
+  data: '0x' + bin + gtxdatanonzero.toString(16).padStart(64, '0'),
```
forshtat: We will have a lot of constructor parameters in the near future. I don't think manually encoding them is a viable long-term solution.

scripts/singleton/get-deploy-data.js
```diff
@@ -23,6 +28,7 @@ const deployer = tx.getSenderAddress()
 
 console.log(JSON.stringify({
   deployer: toHexString(deployer),
+  gas: gasPrice * gasLimit,
```
forshtat: 'gas' is often referring to 'gas used' or oven 'gas limit', this value is 'tx fee' I assume (if there is no refund, that is)

scripts/singleton/deploy.js
```diff
@@ -1,11 +1,10 @@
 const deployData = require('../../singleton/deploy.json')
-const { BN, toWei } = require('web3-utils')
 
 // Deploys a RelayHub instance, reading required data from singleton/deploy.json, the output of scripts/singleton/compute.sh
 // Call this file with truffle exec on the target network.
 async function deploy () {
   const funder = (await web3.eth.getAccounts())[0]
-  await web3.eth.sendTransaction({ from: funder, to: deployData.deployer, value: new BN(toWei('0.42', 'ether')) })
+  await web3.eth.sendTransaction({ from: funder, to: deployData.deployer, value: deployData.gas })
```
forshtat: If 'deployData.gas' is calculated with absolute precision (which is possible), this will not give enough by 21'000 gas units

forshtat: PS (for the record)
Hard-coding constructor parameters that are added now for differentiation between different chains with different opcode pricings, base currency cost, etc. defeats the purpose, so I propose deprecating this script entirely.

scripts/singleton/compute.sh
shahafn: It removes all occurrences of  "pragma experimental ABIEncoderV2;" and puts only one of it at the top of the file...

PR #298, opened 2020-03-19
Stake manager

forshtat:


contracts/StakeManager.sol
```diff
@@ -0,0 +1,165 @@
...
+    /// maps registrees to their authorized hubs map
+    mapping(address => mapping(address => RelayHubInfo)) public authorizedHubs;
+
+    /// Put a stake for a registree and set its unstake delay.
+    /// If the entry does not exist, it is created, and the caller of this function becomes its owner.
+    /// If the entry already exists, only the owner can call this function.
+    /// @param registree - address that represents a stake entry and controls relay registrations on relay hubs
+    /// @param unstakeDelay - number of blocks to elapse before the owner can retrieve the stake after calling 'unlock'
+    function stakeForAddress(address registree, uint256 unstakeDelay) external payable {
+        require(registrees[msg.sender] == address(0) || registrees[msg.sender] == registree, "different stake exists");
```
drortirosh: at least the comment string is wrong (no stake is defined yet)

forshtat: This particular require? This means this owner already made a stake earlier, and for a different registree address. I don't support it, I think you'll agree this is too much complexity

contracts/StakeManager.sol
```diff
@@ -0,0 +1,165 @@
...
+        require(msg.sender != registree, "registree cannot stake for itself");
+        require(stakes[msg.sender].owner == address(0), "sender is a registree itself");
+        registrees[msg.sender] = registree;
+        stakes[registree].owner = msg.sender;
+        stakes[registree].stake += msg.value;
+        stakes[registree].unstakeDelay = unstakeDelay;
+        emit StakeAdded(registree, stakes[registree].owner, stakes[registree].stake, stakes[registree].unstakeDelay);
+    }
+
+    function unlockStake() external {
```
drortirosh: `unlockStake(registree)`
owner can have multiple registrees.

forshtat: No he can't

contracts/StakeManager.sol
```diff
@@ -0,0 +1,165 @@
...
+        uint256 withdrawBlock;
+        address payable owner;
+    }
+
+    struct RelayHubInfo {
+        uint256 removalBlock;
+    }
+
+    /// maps owners to their registrees
+    mapping(address => address) public registrees;
```
drortirosh: this means a single owner can haev a single registree. That's a problem.
what can I do to have multiple (separate) relays, and still use the same owner address ?
(each relay can have multiple addresses, but that's a separate issue, unrelated to SM)

contracts/StakeManager.sol
```diff
@@ -0,0 +1,165 @@
...
+        hubInfo.removalBlock = removalBlock;
+        emit HubUnauthorized(registree, relayHub, removalBlock);
+    }
+
+    function isRegistreeStaked(address registree, uint256 minAmount, uint256 minUnstakeDelay)
+    external
+    view
+    returns (bool) {
+        StakeInfo storage info = stakes[registree];
+        bool isAmountSufficient = info.stake > minAmount;
```
drortirosh: better have them as separate `require` items. This way they also provide some info on failure

contracts/StakeManager.sol
forshtat: Why do you want one owner to have multiple registrees, again?

contracts/StakeManager.sol
forshtat: I am not sure I want this to revert for failure. I can provide return codes, but this is just 4 flags, it should be reasonably easy to figure out anyways

contracts/StakeManager.sol
```diff
@@ -0,0 +1,165 @@
...
+        isStakeLocked &&
+        isHubAuthorized;
+    }
+
+    /// Slash the stake of the relay registree. In order to prevent stake kidnapping, burns half of stake on the way.
+    /// @param registree - entry to penalize
+    /// @param beneficiary - address that receives half of the penalty amount
+    /// @param amount - amount to withdraw from stake
+    function penalizeRegistree(address registree, address payable beneficiary, uint256 amount) external {
+        require(authorizedHubs[registree][msg.sender].removalBlock > block.number, "hub not authorized");
```
drortirosh: need first `require(  removalBlock!=0, "not authorized" )`
and then  `require ( removalBlock > block.Number, "authorization expired" )`

contracts/StakeManager.sol
drortirosh: I can have multiple relays. Each relay appear once in the list a client use, regardless of how many addresses it has. I want to have multiple relays - and to do so I have to stake more than once.

contracts/StakeManager.sol
forshtat: That would only affect a revert string, why bother? In tests I care about the right outcome, in production you don't get the revert reason in tx receipt anyway

contracts/StakeManager.sol
drortirosh: revert is excellent, especially where this is exactly what the caller (RelayHub) would do with the result

contracts/StakeManager.sol
forshtat: No. Client uses list of relays. This is controlled by Relay Hub.
Owner puts a stake for "registree" address. This address can call "registerNewRelay". This is what gives him this name

contracts/StakeManager.sol
drortirosh: just like your require "penalty exceeds stake" below: its redundant (since SafeMath.sub does the same), but gives much better error message.
Better errors are always better. tools are getting there to give more info (e.g. tenderly shows exact error for failed txs)

forshtat: SafeMath errors are more obscure. I will add a separate require if you insist

contracts/StakeManager.sol
forshtat: I don't want to make this decision for the Relay Hub. I think returning boolean does the job

drortirosh: You don't make a decision here. returning a boolean is simply hiding the reason for returning false.

contracts/StakeManager.sol
```diff
@@ -0,0 +1,160 @@
...
+
+    struct RelayHubInfo {
+        uint256 removalBlock;
+    }
+
+    /// maps registrees to their stakes
+    mapping(address => StakeInfo) public stakes;
+
+    /// maps registrees to their authorized hubs map
+    mapping(address => mapping(address => RelayHubInfo)) public authorizedHubs;
```
drortirosh: reading from top to bottom, I can't understand the mapping: `authHubs[registree][??]`

forshtat: RelayHubInfo i = authHubs[registree][hubAddress] ;

drortirosh: (that it, add a proper comment). its is clearer in the stakes above,

contracts/StakeManager.sol
```diff
@@ -0,0 +1,160 @@
...
+
+    /// Put a stake for a registree and set its unstake delay.
+    /// If the entry does not exist, it is created, and the caller of this function becomes its owner.
+    /// If the entry already exists, only the owner can call this function.
+    /// @param registree - address that represents a stake entry and controls relay registrations on relay hubs
```
drortirosh: just look at this documentation line.. "address that represnts a stake" is abreviated much better as "stakedAddress" than "registree"...

contracts/StakeManager.sol
```diff
@@ -0,0 +1,160 @@
...
+    /// @param unstakeDelay - number of blocks to elapse before the owner can retrieve the stake after calling 'unlock'
+    function stakeForAddress(address registree, uint256 unstakeDelay) external payable {
+        require(stakes[registree].owner == address(0) || stakes[registree].owner == msg.sender, "not owner");
+        require(unstakeDelay >= stakes[registree].unstakeDelay, "unstakeDelay cannot be decreased");
+        require(msg.sender != registree, "registree cannot stake for itself");
+        require(stakes[msg.sender].owner == address(0), "sender is a registree itself");
+        stakes[registree].owner = msg.sender;
```
drortirosh: maybe its only optimization: but might do 
`if stakes[registree].owner != msg.sender ) stakes[registree].owner = msg.sender`

contracts/StakeManager.sol
```diff
@@ -0,0 +1,160 @@
...
+    /// maps registrees to their authorized hubs map
+    mapping(address => mapping(address => RelayHubInfo)) public authorizedHubs;
+
+    /// Put a stake for a registree and set its unstake delay.
+    /// If the entry does not exist, it is created, and the caller of this function becomes its owner.
+    /// If the entry already exists, only the owner can call this function.
+    /// @param registree - address that represents a stake entry and controls relay registrations on relay hubs
+    /// @param unstakeDelay - number of blocks to elapse before the owner can retrieve the stake after calling 'unlock'
+    function stakeForAddress(address registree, uint256 unstakeDelay) external payable {
+        require(stakes[registree].owner == address(0) || stakes[registree].owner == msg.sender, "not owner");
```
drortirosh: in general, its good to add a prefix, `gsn.sm: not owner`

contracts/StakeManager.sol
```diff
@@ -0,0 +1,160 @@
...
+        StakeInfo storage info = stakes[registree];
+        require(info.owner == msg.sender, "not owner");
+        require(info.withdrawBlock == 0, "already pending");
+        info.withdrawBlock = block.number + info.unstakeDelay;
+        emit StakeUnlocked(registree, msg.sender, info.withdrawBlock);
+    }
+
+    function withdrawStake(address registree) external {
+        StakeInfo storage info = stakes[registree];
+        require(info.owner == msg.sender, "not owner");
```
drortirosh: modifier ?

contracts/StakeManager.sol
forshtat: Done

contracts/StakeManager.sol
forshtat: I will rename it, and I hope to find a good name we all agree on in the coming days. I think it is fine to merge with this name in the meanwhile.

contracts/StakeManager.sol
forshtat: Then I can steal your stake by topping it

package.json
```diff
@@ -74,23 +77,33 @@
...
+    "eslint": "^6.8.0",
     "eslint-config-standard": "14.1.0",
-    "eslint-plugin-import": "2.18.2",
+    "eslint-config-standard-with-typescript": "^14.0.0",
+    "eslint-plugin-import": "^2.20.1",
     "eslint-plugin-mocha-no-only": "1.1.0",
-    "eslint-plugin-node": "10.0.0",
-    "eslint-plugin-promise": "4.2.1",
-    "eslint-plugin-standard": "4.0.1",
+    "eslint-plugin-node": "^9.2.0",
```
drortirosh: why it was taken back?

contracts/StakeManager.sol
forshtat: In general, we spend a lot of on-chain storage keeping our revert reasons as human readable sentences that are only useful for ganache and truffle tests, not mainnet. What I want to do consistently is replace errors with error codes and have a helper that converts these in truffle. That would take some time though.

patches/source-map-support+0.5.16.patch
```diff
@@ -0,0 +1,13 @@
+diff --git a/node_modules/source-map-support/source-map-support.js b/node_modules/source-map-support/source-map-support.js
```
drortirosh: where did yo get this from? no documentation (either here or where its applied)

package.json
forshtat: don't know, will try to fix

patches/source-map-support+0.5.16.patch
forshtat: I will DM you on this

test/StakeManager.test.ts
```diff
@@ -0,0 +1,378 @@
+import sourceMapSupport from 'source-map-support'
```
drortirosh: any component must import it separately?

test/StakeManager.test.ts
```diff
@@ -0,0 +1,378 @@
...
+        stakeManager.penalizeRegistree(registree, nonOwner, initialStake, { from: anyRelayHub }),
+        'hub not authorized'
+      )
+    })
+
+    it('should allow querying registree\'s stake', async function () {
+      // @ts-ignore (typechain does not declare names or iterator for return types)
+      const { stake: actualStake, unstakeDelay: actualUnstakeDelay, owner: actualOwner } =
+        await stakeManager.stakes(registree)
+      expect(actualOwner).to.equal(owner)
```
drortirosh: I prefer doing something like `assert.deepEqual( {stake, unstakeDelay, owner}, {stake:initialStake, unstakeDelay: initialUnstakeDelay, owner: intialOwner })`
this way, you have a single error on all (it is a little tricky with bignumbers, though..)

test/StakeManager.test.ts
forshtat: I will revert this in favor of 0x library fix (see next commit)

test/StakeManager.test.ts
forshtat: The single error for 3 variables is not a good thing in any univers

test/server_helper_test.js
```diff
@@ -13,7 +13,7 @@ const gasPricePercent = 20
 
 // ServerHelper adds "noise" to shuffle requests with the same score.
 // this will prevent this randomness, to make tests deterministic.
-const noRandomness = () => 0
+const noRandomness = () => 0.5
```
drortirosh: no real need to modify it, as you completely disable the test.
I will implement the randomness differently.

forshtat: There are some tests that I didn't disable. I only disabled the one test that depends on randomness

contracts/StakeManager.sol
drortirosh: no. its only "don't write if it doesn't modify". 
in Solidity, doing "this.a=1" costs you a storage, even if a is already 1. wrapping with "if" makes you just pay for condition and read, but avoid writing the same value (though I'm sure future EVM will remove the extra cost of writing same-value)

contracts/StakeManager.sol
```diff
@@ -0,0 +1,160 @@
...
+    /// Put a stake for a registree and set its unstake delay.
+    /// If the entry does not exist, it is created, and the caller of this function becomes its owner.
+    /// If the entry already exists, only the owner can call this function.
+    /// @param registree - address that represents a stake entry and controls relay registrations on relay hubs
+    /// @param unstakeDelay - number of blocks to elapse before the owner can retrieve the stake after calling 'unlock'
+    function stakeForAddress(address registree, uint256 unstakeDelay) external payable {
+        require(stakes[registree].owner == address(0) || stakes[registree].owner == msg.sender, "not owner");
+        require(unstakeDelay >= stakes[registree].unstakeDelay, "unstakeDelay cannot be decreased");
+        require(msg.sender != registree, "registree cannot stake for itself");
+        require(stakes[msg.sender].owner == address(0), "sender is a registree itself");
```
drortirosh: interesting test, but I'm not sure it is required. why to force owner not to be a registree ?

contracts/StakeManager.sol
drortirosh: Numeric Error codes are horrible. we should remove them, not add new ones.
what we can do is have short descriptions. they should be unique, and easily understandable. we might add comments with long descriptions.
"not owner" is great, even with prefix. actually most of your errors are quite short. maybe "registree cannot stake for itself" can be shortened to "cannot state self"
see SafeMath as an example.

contracts/StakeManager.sol
forshtat: We don't generally allow one address to take multiple roles, so if this can be easily verified I try to disable anomalies.

PR #301, opened 2020-03-24
Slow typescript adoption - migrate two tests

forshtat:


package.json
```diff
@@ -48,7 +48,7 @@
     "gsn-dock-relay-ganache": "./scripts/gsn-dock-relay-ganache",
     "web": "./restart-relay.sh web",
     "trace": "MODE=trace truffle test",
-    "coverage": "./scripts/run-coverage",
+    "coverage": "yarn generate && ./scripts/run-coverage",
     "clean-coverage": "shx rm -rf coverage coverage.json",
     "generate": "rm -rf ./build/contracts && truffle compile && typechain --target trufflehotfix './build/contracts/**/*.json'",
```
drortirosh: any special reason to remove the folder before each build? it slows down repeated build.

package.json
```diff
@@ -124,7 +125,7 @@
     "truffle-hdwallet-provider": "1.0.17",
     "ts-node": "8.6.2",
     "typechain": "1.0.5",
-    "typechain-target-trufflehotfix": "0.0.2-alpha",
+    "typechain-target-trufflehotfix": "0.0.4-alpha",
```
drortirosh: what exactly is this? no readme explains. it looks (from title) like a generic fix, but it depends on `openeth/truffle-typings` - so its NOT generic..

test/RelayHubGasCalculations.test.ts
```diff
@@ -39,14 +46,14 @@ contract('RelayHub gas calculations', async function ([_, relayOwner, relayAddre
     post: 2080
   }
 
-  let relayHub
-  let recipient
-  let paymaster
+  let relayHub: RelayHubInstance
```
drortirosh: something is strange. `relayHub` is an instance, of type `RelayHub`. but what's the type `RelayHubInstance` ?

test/RelayHubGasCalculations.test.ts
```diff
@@ -173,8 +181,13 @@ contract('RelayHub gas calculations', async function ([_, relayOwner, relayAddre
     })
   })
 
-  async function getBalances () {
+  async function getBalances (): Promise<{
+    relayRecipient: BN
+    relay: Big
```
drortirosh: looking at it this way, it IS ugly (mixing "Big" and "BN" params...) but I guess you'll say this commit is just for adding typescript, not fixing the code..

package.json
forshtat: Yes. If the contract was renamed/removed, it will remain there. The folder should be removed.

test/RelayHubGasCalculations.test.ts
```diff
@@ -184,25 +197,25 @@ contract('RelayHub gas calculations', async function ([_, relayOwner, relayAddre
...
     if (gasDiff !== '0') {
       console.log('== zero-fee unmatched gas. RelayHub.gasOverhead should be increased by: ' + gasDiff.toString())
     }
   }
 
-  describe('check calculation does not for different fees', async function () {
+  describe('check calculation does not break for different fees', function () {
     before(async function () {
-      await relayHub.depositFor(relayOwner, { value: 1 })
+      await relayHub.depositFor(relayOwner, { value: (1).toString() })
```
drortirosh: strange. can't this be a number? (or at least BN)

package.json
forshtat: It is not generic fix, it fixed a few typing issues I had because TypeChain uses truffle 4 and we use truffle 5. 
I cannot publish this package under 'openeth' because the lookup for it is hardcoded in TypeChain itself

test/RelayHubGasCalculations.test.ts
forshtat: Generated type of `RelayHub` truffle contract instance.
`relayHub` is an instance of `RelayHubInstance`.
RelayHub is an instance of artifact `TruffleContract<RelayHub>`

test/RelayHubGasCalculations.test.ts
forshtat: exactly. already fixed in my current branch. will merge separately.

test/RelayHubGasCalculations.test.ts
forshtat: value is `string|BN`, and that is actually smart considering how badly `number` holds up for big values

PR #304, opened 2020-03-27
TrustedForwarder contract

drortirosh:
TrustedForwarder contract
- client recipient has a reference to its trusted forwarder (NOT relayhub)
- probably can be set at the constructor -  less likely to change (still, it might change - e.g. if signature struct changes)
- holds all functionality of nonce, signature verification.

### DIFF
--- a/contracts/RelayHub.sol
+++ b/contracts/RelayHub.sol
@@ -15,6 +15,8 @@ import "./utils/GsnUtils.sol";
 import "./utils/RLPReader.sol";
 import "./interfaces/IRelayHub.sol";
 import "./interfaces/IPaymaster.sol";
+import "./interfaces/ITrustedForwarder.sol";
+import "./BaseRelayRecipient.sol";
 
 contract RelayHub is IRelayHub {
 
@@ -43,7 +45,7 @@ contract RelayHub is IRelayHub {
     */
 
     // Gas cost of all relayCall() instructions after actual 'calculateCharge()'
-    uint256 constant private GAS_OVERHEAD = 21717;
+    uint256 constant private GAS_OVERHEAD = 21739;
 
     function getHubOverhead() external view returns (uint256) {
         return GAS_OVERHEAD;
@@ -53,10 +55,6 @@ contract RelayHub is IRelayHub {
 
     uint256 public gtxdatanonzero;
     uint256 constant public GTRANSACTION = 21000;
-
-
-    // Nonces of senders, used to prevent replay attacks
-    mapping(address => uint256) private nonces;
 
     enum AtomicRecipientCallsStatus {OK, CanRelayFailed, RelayedCallFailed, PreRelayedFailed, PostRelayedFailed}
 
@@ -78,10 +76,7 @@ contract RelayHub is IRelayHub {
 
     string public version = "1.0.0";
 
-    EIP712Sig private eip712sig;
-
     constructor (uint256 _gtxdatanonzero) public {
-        eip712sig = new EIP712Sig(address(this));
         gtxdatanonzero = _gtxdatanonzero;
     }
 
@@ -120,6 +115,7 @@ contract RelayHub is IRelayHub {
 
         emit Staked(relay, relays[relay].stake, relays[relay].unstakeDelay);
     }
+
     function registerRelay(uint256 baseRelayFee, uint256 pctRelayFee, string memory url) public {
         address relay = msg.sender;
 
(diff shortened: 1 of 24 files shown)


contracts/TrustedForwarder.sol
```diff
@@ -0,0 +1,40 @@
...
+
+    constructor() public {
+        eip712sig = new EIP712Sig(address(this));
+    }
+
+    function getNonce(address from) external view returns (uint256) {
+        return nonces[from];
+    }
+
+    function verify(GSNTypes.RelayRequest memory req, bytes memory sig) view public returns (bool){
```
forshtat: I do not like that there is a bi-directional API between Forwarder and Relay Hub. This means there is a need for all participants to trust each other and communicate.
A malicious forwarder can drain a Paymaster by always returning `true` but reverting all transactions.
Relay Hub should not express any guarantees about forwarder. With this structure, the Paymaster will need to decide if it wants to trust this Forwarder's `verify` method.
This may be acceptable, but can we design around it?

contracts/TrustedForwarder.sol
```diff
@@ -0,0 +1,40 @@
...
+        return eip712sig.verify(req, sig);
+    }
+
+    function verifyAndCall(GSNTypes.RelayRequest memory req, bytes memory sig) public returns (bool success, bytes memory ret) {
+        if (!verify(req, sig))
+            return (false, "can't call: wrong signature");
+
+        nonces[req.relayData.senderAddress]++;
+
+        return req.target.call.gas(req.gasData.gasLimit)
```
forshtat: I suggest we add smth like `require(gasleft() - req.gasData.gasLimit > gasleft()/64)` here to enforce the gas limit. It is not strictly a responsibility of the Forwarder, but it may be inconvenient the meta-tx signer cannot control the gas actually given to the relayed call.

contracts/TrustedForwarder.sol
```diff
@@ -0,0 +1,40 @@
...
+        (abi.encodePacked(req.encodedFunction, req.relayData.senderAddress));
+    }
+}
+
+library asd {
+    function dsa() internal {new TrustedForwarder();}
+}
```
forshtat: ?

contracts/interfaces/ITrustedForwarder.sol
```diff
@@ -0,0 +1,16 @@
...
+
+import "../utils/GSNTypes.sol";
+
+contract ITrustedForwarder {
+
+    // verify the signature matches the request.
+    //  that is, the senderAccount is the signer
+    function verify(GSNTypes.RelayRequest memory req, bytes memory sig) view public returns (bool);
+
+    function getNonce(address from) external view returns (uint256);
```
forshtat: implementation detail

contracts/TrustedForwarder.sol
```diff
@@ -0,0 +1,40 @@
...
+
+    function getNonce(address from) external view returns (uint256) {
+        return nonces[from];
+    }
+
+    function verify(GSNTypes.RelayRequest memory req, bytes memory sig) view public returns (bool){
+        return eip712sig.verify(req, sig);
+    }
+
+    function verifyAndCall(GSNTypes.RelayRequest memory req, bytes memory sig) public returns (bool success, bytes memory ret) {
```
forshtat: do you need a return bool? just revert

drortirosh: if I revert, I'd lose the reason - unless I parse the resulting byte array, and extract the reason.
instead, verifyAndCall uses the same return values `address.call` would

contracts/TrustedForwarder.sol
drortirosh: the RelayHub can't trust the forwarder to enforce the gas limit, so we need t remove it from here, and move it (with some offset) to the RelayHub..

contracts/RelayHub.sol
```diff
@@ -240,16 +240,21 @@ contract RelayHub is IRelayHub {
...
+                    forwarder.verify.selector,
+                    relayRequest, signature
+                ));
+            if (!success) {
+                ret = bytes(GsnUtils.getError(ret));
+                //temporary check (until we get rid of the status codes, and rely only on description)
+                if (ret[0] == "n")
+                    return (uint256(CanRelayStatus.WrongNonce), ret);
+                else
+                    return (uint256(CanRelayStatus.WrongSignature), ret);
```
forshtat: :-)

forshtat:
@shahafn please have a look as well

PR #309, opened 2020-03-31
Accept relayed call revert alexf

forshtat:


contracts/RelayHub.sol
```diff
@@ -238,45 +238,36 @@ contract RelayHub is IRelayHub {
...
-            // This can be either CanRelayStatus.OK or a custom error code
-            if ((status == 0) || (status > 10)) {
-                return (status, recipientContext);
-            } else {
-                // Error codes [1-10] are reserved to RelayHub
-                return (uint256(CanRelayStatus.InvalidRecipientStatusCode), "InvalidRecipientStatusCode");
-            }
+            return (false, GsnUtils.getError(ret));
         }
+        returnValue = abi.decode(ret, (string));
```
drortirosh: No. the context is the return value, without decode.
The reason I made canRelay return `string` and not `bytes`, is to ease the client parse (which cares nothing about the context, only about the error string).
on-chain is simply `string(x)` or `bytes(y)` to convert between them.

PR #310, opened 2020-03-31
TrustedBatchedForwarder

drortirosh:
a mechanism to send a batch of request from the same sender.
usage:
- add the TrustedBatchForwarder as the forwarder of your RelayRecipient
contract
- make sure the Paymaster accepts calls through this forwarder.
- make a relayed request to forwarder.sendBatch(targets,
encodedFunctions)
- each target will be called in sequence. each will have the caller set
as the getSender().


test/TestBatchForwarder.test.js
```diff
@@ -0,0 +1,156 @@
...
+const StakeManager = artifacts.require('StakeManager')
+const TrustedBatchForwarder = artifacts.require('./TrustedBatchForwarder.sol')
+const TestRecipient = artifacts.require('TestRecipient.sol')
+const { getEip712Signature } = require('../src/common/utils')
+const getDataToSign = require('../src/common/EIP712/Eip712Helper')
+const { expectEvent } = require('@openzeppelin/test-helpers')
+
+const RelayRequest = require('../src/common/EIP712/RelayRequest')
+
+contract.only('TrustedBatchForwarder', ([from, relayManager, relayWorker, relayOwner]) => {
```
forshtat, line 18: only

PR #318, opened 2020-04-13
Penalizer

shahafn:
Added also try..finally & test on server to avoid nonce mutex deadlock

### DIFF
--- a/test/RelayHubGasCalculations.test.ts
+++ b/test/RelayHubGasCalculations.test.ts
@@ -9,12 +9,13 @@ import RelayRequest from '../src/common/EIP712/RelayRequest'
 import {
   RelayHubInstance,
   TestRecipientInstance,
-  TestPaymasterVariableGasLimitsInstance, StakeManagerInstance, ITrustedForwarderInstance
+  TestPaymasterVariableGasLimitsInstance, StakeManagerInstance, ITrustedForwarderInstance, PenalizerInstance
 } from '../types/truffle-contracts'
 
 const RelayHub = artifacts.require('RelayHub')
 const TrustedForwarder = artifacts.require('TrustedForwarder')
 const StakeManager = artifacts.require('StakeManager')
+const Penalizer = artifacts.require('Penalizer')
 const TestRecipient = artifacts.require('TestRecipient')
 const TestPaymasterVariableGasLimits = artifacts.require('TestPaymasterVariableGasLimits')
 const TestPaymasterConfigurableMisbehavior = artifacts.require('TestPaymasterConfigurableMisbehavior')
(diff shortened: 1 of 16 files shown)


test/RelayHubGasCalculations.test.ts
```diff
@@ -43,13 +44,14 @@ contract('RelayHub gas calculations', function ([_, relayOwner, relayWorker, rel
   const gasLimit = new BN('1000000')
   const senderNonce = new BN('0')
   const magicNumbers = {
-    arc: 805 - 6,
-    pre: 1839 + 22,
-    post: 2080
+    arc: 805 - 6 - 62,
+    pre: 1839 + 22 - 95,
+    post: 2080 - 112
```
forshtat, line 49: we got to do something about it :-)

test/RelayServer.test.js
```diff
@@ -667,6 +670,22 @@ contract('RelayServer', function (accounts) {
       const promises = [relayTransaction(options), relayTransaction(options2)]
       await Promise.all(promises)
     })
+    it('should not deadlock if server returned error while locked', async function () {
```
forshtat, line 673: damn right should not 😂

test/RelayServer.test.js
```diff
@@ -667,6 +670,22 @@ contract('RelayServer', function (accounts) {
...
+      try {
+        relayServer.keyManager.signTransactionOrig = relayServer.keyManager.signTransaction
+        relayServer.keyManager.signTransaction = function () {
+          throw new Error('no tx for you')
+        }
+        try {
+          await relayTransaction(options)
+        } catch (e) {
+          assert.equal(e.message, 'no tx for you', e.message)
+          assert.isFalse(relayServer.nonceMutex.isLocked(), 'nonce mutex not released after exception')
```
forshtat, line 683: but this is quite a white-box test... it would be more articulate if you made a second 'relayTransaction' request and assert your computer did not catch fire

truffle.js
```diff
@@ -55,7 +55,11 @@ module.exports = {
     solc: {
       version: packageJson.devDependencies.solc,
       settings: {
-        evmVersion: 'istanbul'
+        evmVersion: 'istanbul',
+        optimizer: {
+          enabled: true,
+          runs: 200 // Optimize for how many times you intend to run the code
```
forshtat, line 61: Does it take noticeably more time to compile with optimizer?

test/RelayHubGasCalculations.test.ts
shahafn: Yes we do

truffle.js
shahafn: No, from what I saw

PR #320, opened 2020-04-15
Tests new rc

forshtat:


src/common/utils.js
```diff
@@ -72,7 +72,7 @@ module.exports = {
     return 21000 +
       hubOverhead +
       calldataSize * gtxdatanonzero +
-      relayCallGasLimit +
+      parseInt(relayCallGasLimit) +
```
drortirosh, line 75: I thnk the field members can be defined as ints. no reason to parseInt them.

src/relayclient/AccountManager.ts
```diff
@@ -7,25 +7,36 @@ import { getEip712Signature, isSameAddress } from '../common/utils'
...
+import Web3 from 'web3'
 
 export interface AccountKeypair {
   privateKey: Buffer
   address: Address
 }
 
+function toAddress (wallet: any): string {
+  // eslint-disable-next-line @typescript-eslint/restrict-template-expressions
+  return `0x${wallet.getAddress().toString('hex')}`
```
drortirosh, line 20: instead of disable typescript, simply use `'0x'+xxx.toString()`

src/relayclient/AccountManager.ts
```diff
@@ -34,7 +45,7 @@ export default class AccountManager {
     const keypair = {
       privateKey: a.privKey,
       // eslint-disable-next-line @typescript-eslint/restrict-template-expressions
```
drortirosh, line 47: remove disable lint..

src/relayclient/AccountManager.ts
```diff
@@ -43,34 +54,52 @@ export default class AccountManager {
...
+      throw new Error(`Failed to sign relayed transaction for ${relayRequest.relayData.senderAddress}`)
     }
-    // Sanity check only
-    // @ts-ignore
-    const rec = sigUtil.recoverTypedSignature_v4({
-      data: signedData,
-      sig: signature
-    })
     if (!isSameAddress(relayRequest.relayData.senderAddress.toLowerCase(), rec)) {
       // eslint-disable-next-line @typescript-eslint/restrict-template-expressions
```
drortirosh, line 82: better disable restrict-template-expressions globally. we never have templates, and have to add this each time.

src/relayclient/AccountManager.ts
```diff
@@ -43,34 +54,52 @@ export default class AccountManager {
...
-      throw new Error(`signature error: sender=${relayRequest.relayData.senderAddress}, recovered=${rec}`)
+      throw new Error(`Internal RelayClient exception: signature is not correct: sender=${relayRequest.relayData.senderAddress}, recovered=${rec}`)
     }
     return signature
   }
+
+  // These methods is extracted to
+  // a) allow different implementations in the future, and
+  // b) allow spying on Account Manager in tests
+  async _signWithExternallyControlledKey (signedData: any): Promise<string> {
```
drortirosh, line 91: I'm not sure about the method naming: its not "external", but using provider's sign method (that is, provider-controlled accounts)

src/relayclient/ContractInteractor.ts
```diff
@@ -26,10 +28,12 @@ let IRelayHubContract: Contract<IRelayHubInstance>
 let IForwarderContract: Contract<ITrustedForwarderInstance>
 
 export default class ContractInteractor {
+  private readonly web3: Web3
   private readonly provider: provider
   private readonly config: ContractInteractorConfig
 
   constructor (provider: provider, config: ContractInteractorConfig) {
+    this.web3 = new Web3(provider)
```
drortirosh, line 36: you actually took most references to web3 out. maybe it can be left confined to the interactor...

src/common/utils.js
forshtat: All of these will be fixed once utils are translated to TypeScript. This will require server to be compiled as well. Now I am afraid to change types of fields because instead of compilation errors it may cause hard-to-find bugs in server where, in JS logic, `'1' + 2 = '12'`

src/relayclient/AccountManager.ts
forshtat: That is not a solution, it is hiding the problem. The problem is that wallet does not have types. It is not that hard to declare the types for it, which is a solution that I think we should do.

src/relayclient/AccountManager.ts
forshtat: I don't want to

src/relayclient/AccountManager.ts
forshtat: You only need to add it if TypeScript does not know what type your expression has. This is a symptom, I do not want to globally silence the symptom.
Some libraries won't have any type annotations, and this is ok, as long as we mark those explicitly.

src/relayclient/AccountManager.ts
forshtat: Give me a name then.

src/relayclient/ContractInteractor.ts
forshtat: I do intend to remove all references to web3 from anywhere but interactor, and rename it BlockchainInteractor.

src/relayclient/RelayClient.ts
```diff
@@ -42,68 +40,72 @@ export interface RelayingResult {
...
-   * @param asyncApprove
    */
   constructor (
-    web3: Web3,
-    httpClient: HttpClient,
-    contractInteractor: ContractInteractor,
-    knownRelaysManager: KnownRelaysManager,
-    accountManager: AccountManager,
-    relayedTransactionValidator: RelayedTransactionValidator,
+    dependencyTree: TmpDependencyTree,
```
drortirosh, line 68: bad name. if its the RelayClient, why its args called "Tmp"?

forshtat: I do not like this solution. I want it to have an annoying name :-)

src/relayclient/RelayClient.ts
```diff
@@ -42,68 +40,72 @@ export interface RelayingResult {
...
+    this.knownRelaysManager = dependencyTree.knownRelaysManager
+    this.transactionValidator = dependencyTree.transactionValidator
+    this.accountManager = dependencyTree.accountManager
+    this.pingFilter = dependencyTree.pingFilter
     this.relayHub = relayHub
-    this.asyncApprove = asyncApprove
+    this.asyncApprove = dependencyTree.asyncApprove
+  }
+
+  static getDefaultDependencies (provider: HttpProvider, gsnConfig: GSNConfig): TmpDependencyTree {
```
drortirosh, line 83: again: why not default args? in order to change one dependency, you need to create all, then then modify some..

src/relayclient/RelayClient.ts
```diff
@@ -42,68 +40,72 @@ export interface RelayingResult {
...
    */
-  static new (web3: Web3, gsnConfig: GSNConfig): RelayClient {
-    const httpWrapper = new HttpWrapper()
-    const httpClient = new HttpClient(httpWrapper, { verbose: false })
-    const contractInteractor = new ContractInteractor(web3.currentProvider, gsnConfig.contractInteractorConfig)
-    const knownRelaysManager = new KnownRelaysManager(web3, gsnConfig.relayHubAddress, contractInteractor, EmptyFilter, gsnConfig.knownRelaysManagerConfig)
-    const accountManager = new AccountManager(web3, defaultEnvironment.chainId, gsnConfig.accountManagerConfig)
-    const transactionValidator = new RelayedTransactionValidator(contractInteractor, gsnConfig.relayHubAddress, defaultEnvironment.chainId, gsnConfig.transactionValidatorConfig)
-    return new RelayClient(web3, httpClient, contractInteractor, knownRelaysManager, accountManager, transactionValidator, gsnConfig.relayHubAddress, GasPricePingFilter, EmptyApprove, gsnConfig.relayClientConfig)
+  static new (provider: HttpProvider, gsnConfig: GSNConfig): RelayClient {
```
drortirosh, line 106: don't depend on HttpProvider.

src/relayclient/RelayClient.ts
```diff
@@ -136,12 +141,9 @@ export default class RelayClient {
...
     await this.knownRelaysManager.refresh()
     gsnTransactionDetails.gasPrice = gsnTransactionDetails.forceGasPrice ?? await this._calculateGasPrice()
-    if (gsnTransactionDetails.gasPrice == null) {
-      gsnTransactionDetails.gas = await this._calculateGasPrice()
-    }
     if (gsnTransactionDetails.gas == null) {
-      const estimated = await this.web3.eth.estimateGas(gsnTransactionDetails)
-      gsnTransactionDetails.gas = estimated.toString()
+      const estimated = await this.contractInteractor.estimateGas(gsnTransactionDetails)
+      gsnTransactionDetails.gas = `0x${estimated.toString(16)}`
```
drortirosh, line 146: I'm sure there's `toHex()` somewhere, and if not, then need to create one

src/relayclient/RelayClient.ts
```diff
@@ -212,28 +214,33 @@ export default class RelayClient {
...
     const relayWorker = relayInfo.pingResponse.RelayServerAddress
-    const gasPrice = gsnTransactionDetails.gasPrice
-    const gasLimit = gsnTransactionDetails.gas
-    if (gasPrice == null || gasLimit == null) {
+    const gasPriceHex = gsnTransactionDetails.gasPrice
+    const gasLimitHex = gsnTransactionDetails.gas
+    if (gasPriceHex == null || gasLimitHex == null) {
       throw new Error('RelayClient internal exception. Gas price or gas limit still not calculated. Cannot happen.')
     }
+    if (gasPriceHex.indexOf('0x') !== 0 || gasLimitHex.indexOf('0x') !== 0) {
```
drortirosh, line 222: `/0x\w+/.test(gasPriceHex)`
or, create `isHex()` method, which checks for non-null, prefix and character set)

(to be strict, the check is `0x[0-9a-fA-F]+`.test(..)`

src/relayclient/RelayProvider.ts
```diff
@@ -1,45 +1,46 @@
...
    * @param origProvider - the underlying web3 provider
    * @param gsnConfig
    */
-  constructor (web3: Web3, origProvider: provider | RelayProvider, gsnConfig: GSNConfig) {
-    if (origProvider instanceof RelayProvider ||
-      origProvider == null ||
-      typeof origProvider === 'string') {
-      throw new Error('Missing underlying provider')
-    }
+  constructor (relayClient: RelayClient | undefined, origProvider: HttpProvider, gsnConfig: GSNConfig) {
```
drortirosh, line 33:
```
 constructor (origProvider: HttpProvider, gsnConfig: GSNConfig,
relayClient = new RelayClient(origProvider,gsnConfig)) 
```

I think this is shorter and more understandable.

test/dummies/BadRelayClient.ts
```diff
@@ -0,0 +1,37 @@
+import RelayClient, { RelayingResult, TmpDependencyTree } from '../../src/relayclient/RelayClient'
+import GsnTransactionDetails from '../../src/relayclient/types/GsnTransactionDetails'
+import { Address } from '../../src/relayclient/types/Aliases'
+import { RelayClientConfig } from '../../src/relayclient/GSNConfigurator'
+
+export default class BadRelayClient extends RelayClient {
```
drortirosh, line 6: its only for tests, so can be achieved much simpler. e.g.
```
badCliet = new RelayClient()
badCleint.relayTransaction = ()=>{throw new Error()}}
```

test/relayclient/AccountManager.test.ts
```diff
@@ -0,0 +1,116 @@
...
+      // @ts-ignore
+      const rec = sigUtil.recoverTypedSignature_v4({
+        data: signedData,
+        sig: signature
+      })
+      assert.ok(isSameAddress(relayRequest.relayData.senderAddress.toLowerCase(), rec))
+      expect(accountManager._signWithControlledKey).to.have.been.calledWith(keypair, signedData)
+      expect(accountManager._signWithExternallyControlledKey).to.have.not.been.called
+    })
+    it('should ask web3 to sign if key is not controlled', async function () {
```
drortirosh, line 93: ask *provider* to sign

src/relayclient/RelayClient.ts
forshtat: We've got it for buffer and bignumber. This is just 'number'. We don't need number-to-hex usually, this is the only place I can think of it being used right now.
There is some legacy mess leftover with types of things. I fix these on the go.

src/relayclient/RelayClient.ts
forshtat: What do I depend on then? I can define interface that has only 'send' method, cause it is the only method we use. Next PR, though.

src/relayclient/RelayClient.ts
forshtat: Regex are nice but hard to read. This is verbose but dead obvious what it does. There is absolutely no benefit in using regex here.

test/dummies/BadRelayClient.ts
forshtat: I have a very strong opinion against assigning methods of objects. You are not allowed (at least without reflection) to do it in any sane language. 
The fact you can do it in JS/TS does not mean you should.

test/relayclient/RelayProvider.test.ts
```diff
@@ -0,0 +1,274 @@
...
+        params: [
+          {
+            from: gasLess,
+            gas: '0x186a0',
+            gasPrice: '0x4a817c800',
+            forceGasPrice: '0x51f4d5c00',
+            paymaster,
+            forwarder,
+            to: testRecipient.address,
+            data: '0x2ac0df260000000000000000000000000000000000000000000000000000000000000020000000000000000000000000000000000000000000000000000000000000000b68656c6c6f20776f726c64000000000000000000000000000000000000000000'
```
drortirosh, line 114: why not "encodeABI" ?

test/relayclient/RelayProvider.test.ts
```diff
@@ -0,0 +1,274 @@
...
+      return expect(promisified).to.be.eventually.rejectedWith(`Rejected relayTransaction call - should not happen. Reason: Error: ${BadRelayClient.message}`)
+    })
+
+    it('should call callback with error containing relaying results dump if relayTransaction does not return a transaction object', function () {
+      const badRelayClient = new BadRelayClient(false, true, dependencyTree, gsnConfig.relayHubAddress, gsnConfig.relayClientConfig)
+      const relayProvider = new RelayProvider(badRelayClient, underlyingProvider, gsnConfig)
+      const promisified = new Promise((resolve, reject) => relayProvider._ethSendTransaction(jsonRpcPayload, (error: Error | null): void => {
+        reject(error)
+      }))
+      return expect(promisified).to.be.eventually.rejectedWith('Failed to relay call. Results:')
```
drortirosh, line 135: how does it check any? you didn't `await`

test/relayclient/RelayProvider.test.ts
```diff
@@ -0,0 +1,274 @@
...
+      const promisified = new Promise((resolve, reject) => relayProvider._ethSendTransaction(jsonRpcPayload, (error: Error | null): void => {
+        reject(error)
+      }))
+      return expect(promisified).to.be.eventually.rejectedWith(`Rejected relayTransaction call - should not happen. Reason: Error: ${BadRelayClient.message}`)
+    })
+
+    it('should call callback with error containing relaying results dump if relayTransaction does not return a transaction object', function () {
+      const badRelayClient = new BadRelayClient(false, true, dependencyTree, gsnConfig.relayHubAddress, gsnConfig.relayClientConfig)
+      const relayProvider = new RelayProvider(badRelayClient, underlyingProvider, gsnConfig)
+      const promisified = new Promise((resolve, reject) => relayProvider._ethSendTransaction(jsonRpcPayload, (error: Error | null): void => {
```
drortirosh, line 132: better use `promisify`: it handles both rejects and resolved value.

test/relayclient/AccountManager.test.ts
forshtat: fixed

test/relayclient/RelayProvider.test.ts
forshtat: I returned a promise. Equivalent.

test/relayclient/RelayProvider.test.ts
forshtat: We don't use `promisify` anywhere in code currently. I don't need it for this simple thing.

test/relayclient/RelayProvider.test.ts
drortirosh: (that is: if you did you some calculation to get a value, then put the calculation, not the value. the same is for gasPrice: I'm sure you didn't decide on this string of hex digits. you probably had some integer value in mind, whcih is no longer visible

test/relayclient/RelayProvider.test.ts
```diff
@@ -0,0 +1,274 @@
...
+      assert.equal(innerTxFailedReceiptTruffle.logs[0].args.status.toString(), '2') // PreRelayFailed
+      innerTxFailedReceipt = await web3.eth.getTransactionReceipt(innerTxFailedReceiptTruffle.tx)
+
+      await misbehavingPaymaster.setRevertPreRelayCall(false)
+      const innerTxSuccessReceiptTruffle = await relayHub.relayCall(relayRequest, signature, '0x', {
+        from: accounts[0],
+        gasPrice: '1'
+      })
+      assert.equal(innerTxSuccessReceiptTruffle.logs[0].event, 'SampleRecipientEmitted')
+      assert.equal(innerTxSuccessReceiptTruffle.logs[1].event, 'TransactionRelayed')
```
drortirosh, line 240: `expectEvent( 'TransacxtionRelayed', { status: '0' } )`
is nicer to read.

test/relayclient/RelayProvider.test.ts
```diff
@@ -0,0 +1,274 @@
...
+    let testRecipient: TestRecipientInstance
+    let canRelayFailedTxReceipt: BaseTransactionReceipt
+    let innerTxFailedReceipt: BaseTransactionReceipt
+    let innerTxSucceedReceipt: BaseTransactionReceipt
+    let notRelayedTxReceipt: BaseTransactionReceipt
+    let misbehavingPaymaster: TestPaymasterConfigurableMisbehaviorInstance
+    let relayRequest: RelayRequest
+    // It is not stirctly necessary to make this test against actual tx receipt, but I prefer to do it anyway
+    before(async function () {
+      const TestRecipient = artifacts.require('TestRecipient')
```
drortirosh, line 167: why one big fixture for all tests? (e.g. canRelayFailedTxReceipt belong to its test, not to the global `before`)

src/relayclient/AccountManager.ts
drortirosh: good idea. regardless, string concat is clearer

test/dummies/BadRelayClient.ts
shahafn: I agree with Dror, this is one of the few good things about testing in js.

src/relayclient/AccountManager.ts
drortirosh: then its better to move it to be beginning of the file, to ignore all errors...
seriously: if you refactored a line to use a function, then the `eslint-disable` should be moved with it.

src/relayclient/AccountManager.ts
drortirosh: we use backtick notation generously, not only for template values.
(e.g. if you put an expression, it will complain, because it expects a single-var name for template values)

src/relayclient/AccountManager.ts
drortirosh: signWithProviderAccount

src/relayclient/ContractInteractor.ts
drortirosh: ... and then maybe use "ethers", AFAIK its ligther than web3...

src/relayclient/RelayClient.ts
drortirosh: AFAIK i did saw you have `provider` type.

src/relayclient/RelayClient.ts
drortirosh: AFAIK, single test is better than checking both for null and prefix (can have `isHex` too..)

test/dummies/BadRelayClient.ts
forshtat: It is good until it is bad. Your `badClient` no longer implements it's own class or interface, it is just object that has `relayTransaction = ()=>{}` and that's it. If you remove the entire `relayTransaction` from the original object, or rename it to 'relayTra**sn**action' again, test will still pass. 
With TypeScript if you define `badClient` as `RelayClient` it will fail in compile-time with method not found.

drortirosh: the only place I think its good is when writing tests. tests should be self-contained: having a utilty class merely to "stub" a method to return an error makes the test hard to read.  Your construction of `BadRelayClient()` gets some boolean params, which are meaningless when reading the test. its better to assign a method - or use `sinon`

test/relayclient/RelayProvider.test.ts
forshtat: gas price everywhere is 22M that is set in server as minimum, I think, but it can be just a random big enough number, it is not under test here

test/relayclient/RelayProvider.test.ts
drortirosh: I think this is not clear: the method is not declared async, and I don't see amywhere that it is.

test/relayclient/RelayProvider.test.ts
forshtat: They share 90% of setup code, and only differ in what I tell MisbehavingPaymaster to do. Would be tedious to separate into individual `before`s

test/relayclient/RelayProvider.test.ts
forshtat: fixed

test/relayclient/RelayProvider.test.ts
forshtat: Fixed

src/relayclient/AccountManager.ts
drortirosh: blocker: "_signWithExternallyControlledKey" is bad name, especially since it hides where the signature actually comes from.
(the provider call is buried inside getEip712Signature)

test/dummies/BadRelayClient.ts
drortirosh: Future: we need guidelines on how to write tests that need to modify behaviour of existing objects (subclass with modified methods, stubbing/mocking existing ones, plain modification inside a test, etc)

test/relayclient/RelayProvider.test.ts
drortirosh: (except that you do it 4 times in this file, and maybe elsewhere. the syntax of promisify is much simpler (hence much less possible mistakes, and less eye-strain)
(not critical, tough)

test/relayclient/RelayProvider.test.ts
drortirosh: ok got it: its a new way of writing tests: the entire flow runs in the "before" section, and then multiple tests to check its side-effects.
(actually much better than re-running the same "fixture" for each test, and just do its tests)

PR #322, opened 2020-04-16
Improve Relay Client bootstrapping

forshtat:


src/relayclient/GSNConfigurator.ts
```diff
@@ -1,42 +1,30 @@
...
+  gasPriceFactorPercent: GAS_PRICE_PERCENT,
+  minGasPrice: 0,
+  maxRelayNonceGap: MAX_RELAY_NONCE_GAP,
+  sliceSize: 3,
+  relayTimeoutGrace: DEFAULT_RELAY_TIMEOUT_GRACE_SEC,
+  methodSuffix: '',
+  jsonStringifyRequest: false,
   chainId: defaultEnvironment.chainId,
-  relayHubAddress: '0x0000000000000000000000000000000000000000'
+  relayHubAddress: '0x0000000000000000000000000000000000000000',
```
drortirosh, line 26: why should we put a broken value for relayHub?

src/relayclient/GSNConfigurator.ts
```diff
@@ -71,17 +59,18 @@ function mergeDeep (target: any, ...sources: any[]): Object {
...
   }
   const source = sources.shift()
-
-  if (isObject(target) && isObject(source)) {
-    for (const key in source) {
-      if (Object.prototype.hasOwnProperty.call(source, key)) {
-        if (isObject(source[key])) {
-          if (target[key] == null) {
-            Object.assign(target, { [key]: {} })
+  if (source != null) {
```
drortirosh, line 62: yuck. don't recurse - iterate.

test/relayclient/RelayClient.test.ts
```diff
@@ -194,48 +188,47 @@ contract('RelayClient', function (accounts) {
...
     })
 
     it('should return error if transaction returned by a relay does not pass validation', async function () {
-      const badHttpClient = new BadHttpClient(false, false, false, pingResponse, '0x123')
-      const dependencyTree = RelayClient.getDefaultDependencies(underlyingProvider, gsnConfig)
-      const badTransactionValidator = new BadRelayedTransactionValidator(true, dependencyTree.contractInteractor, gsnConfig.relayHubAddress, defaultEnvironment.chainId, gsnConfig.transactionValidatorConfig)
-      dependencyTree.httpClient = badHttpClient
-      dependencyTree.transactionValidator = badTransactionValidator
+      const badHttpClient = new BadHttpClient(configureGSN(gsnConfig), false, false, false, pingResponse, '0x123')
+      let dependencyTree = getDependencies(configureGSN(gsnConfig), underlyingProvider)
```
drortirosh, line 224: thought we renamed dependencyTree ..
(```gsnDependencies``` sounds great name. better then ```overrideDependencies```)

src/relayclient/GSNConfigurator.ts
drortirosh: also - I didn't see a single test of this method.

drortirosh: one last thing: you MODIFY your target parameter. its bad API (I know, shared with Object.assign, but still bad)
also, you don't use that fact - you pass in `{}` - so better remove it altogether, and just merge inputs without modifying any of them.

drortirosh: or better: replace it with `const deepmerge = require('deepmerge')`

src/relayclient/GSNConfigurator.ts
forshtat: What should I put? At least zero address is well-understood missing value. RelayClient is not functional without relay hub and I hate defaults, as you've probably noticed.

src/relayclient/GSNConfigurator.ts
forshtat: This is a copy-pasted code. It is not strictly needed anymore, but I decided to keep it around in case there is any hierarchy in the GSNConfig again.

test/relayclient/RelayClient.test.ts
forshtat: dependencyTree is a test var name, it does not matter. `overrideDependencies` because it is not a complete object, unlike `gsnDependencies` which is.

src/relayclient/GSNConfigurator.ts
drortirosh: it shoud be added with a type, but not with a value.

src/relayclient/GSNConfigurator.ts
drortirosh: so if you use it, then add a test.

src/relayclient/GSNConfigurator.ts
drortirosh: there's no default for it. so leave it out.

forshtat: There are no nullable types in GSNConfig.

src/relayclient/GSNConfigurator.ts
forshtat: No

drortirosh: ok. so remove it. unused, untested (and replacable by single standard method) do not belong to the code.

forshtat: Done

PR #324, opened 2020-04-17
Support StakeManager events for RelaysManager lookup & custom relay score calculation

forshtat:


src/relayclient/KnownRelaysManager.ts
```diff
@@ -1,107 +1,102 @@
...
 
 export const EmptyFilter: RelayFilter = (): boolean => {
   return true
 }
+/**
+ * Basic score is reversed transaction fee, higher is better.
+ * Relays that failed to respond recently will be downgraded for some period of time.
+ */
+export const DefaultRelayScore = async function (relay: RelayRegisteredEventInfo, txDetails: GsnTransactionDetails, failures: RelayFailureInfo[]): Promise<number> {
+  const gasLimit = parseInt(txDetails.gas ?? '0')
```
shahafn, line 16: 1. ?? '0' is prone to bugs imo
2. both gasLimit and gasPrice are txDetails.gas...

forshtat: 1. It is called here after the actual default values are set, but the lint's null type check does not know that
2. Fixed

PR #328, opened 2020-04-20
RelayServer separate worker and manager

drortirosh:
- RelayServer separation of manager from worker (initial support for multiple workers, though only 1 is currently implemented)
- added support for hdkeys in KeyManager
- TxStorage now support unique **(nonce,signer)**
other changes:
- better handling of startup errors
- handling of intermediate states (waiting for fund/register) not as errors
- TxStorage, KeyManager now support in-memory state

### DIFF
--- a/migrations/2_deploy_contracts.js
+++ b/migrations/2_deploy_contracts.js
@@ -1,12 +1,9 @@
 var RelayHub = artifacts.require('./RelayHub.sol')
-var RLPReader = artifacts.require('./RLPReader.sol')
 var StakeManager = artifacts.require('./StakeManager.sol')
 var Penalizer = artifacts.require('./Penalizer.sol')
 
-module.exports = function (deployer) {
-  deployer.deploy(RLPReader)
-  deployer.link(RLPReader, RelayHub)
-  deployer.deploy(RelayHub, 16, '0x0000000000000000000000000000000000000000', '0x0000000000000000000000000000000000000000')
-  deployer.deploy(StakeManager)
-  deployer.deploy(Penalizer)
+module.exports = async function (deployer) {
+  await deployer.deploy(StakeManager)
+  await deployer.deploy(Penalizer)
+  await deployer.deploy(RelayHub, 16, StakeManager.address, Penalizer.address)
 }
--- a/src/relayclient/DevRelayClient.ts
+++ b/src/relayclient/DevRelayClient.ts
@@ -17,8 +17,7 @@ import net from 'net'
 
 import KeyManager = require('../relayserver/KeyManager')
 
-const dayInSec = 24 * 60 * 60
-const weekInSec = dayInSec * 7
+const unstakeDelay = 2000
 
 import TruffleContract = require('@truffle/contract')
 import Contract = Truffle.Contract
@@ -208,17 +207,19 @@ export class DevRelayClient extends RelayClient {
 
     const stakeManager = await IStakeManagerContract.at(stakeManagerAddress)
 
-    await stakeManager.stakeForAddress(relayServer.address, weekInSec, {
-      from: this.devConfig.relayOwner,
-      value: ether('1')
-      // gas: estim
-    })
+    await stakeManager.contract.methods.stakeForAddress(relayServer.address, unstakeDelay).send({ from: this.devConfig.relayOwner, value: ether('1') })
+    // not sure why: the line below started to crash on: Number can only safely store up to 53 bits
+    // (and its not relayed - its a direct call)
+    // await stakeManager.stakeForAddress(relayServer.address, unstakeDelay, {
+    //   from: this.devConfig.relayOwner,
+    //   value: ether('1'),
+    // })
     this.debug('== sending balance to relayServer', relayServer.address)
     await stakeManager.authorizeHub(relayServer.address, this.config.relayHubAddress, { from: this.devConfig.relayOwner })
     await this.contractInteractor.getWeb3().eth.sendTransaction({
       from: this.devConfig.relayOwner,
       to: relayServer.address,
-      value: 1e18
+      value: ether('1')
     })
     this.debug('== waiting for relay')
     // @ts-ignore
--- a/src/relayclient/DevRelayClient.ts
+++ b/src/relayclient/DevRelayClient.ts
@@ -17,8 +17,7 @@ import net from 'net'
 
 import KeyManager = require('../relayserver/KeyManager')
 
-const dayInSec = 24 * 60 * 60
-const weekInSec = dayInSec * 7
+const unstakeDelay = 2000
 
 import TruffleContract = require('@truffle/contract')
 import Contract = Truffle.Contract
@@ -52,7 +51,7 @@ export function runServer (
   relayHub: string,
   devConfig: DevGSNConfig
 ): RunServerReturn {
-  const keyManager = new KeyManager({ ecdsaKeyPair: KeyManager.newKeypair(), workdir: devConfig.relayWorkdir })
+  const keyManager = new KeyManager({ count: 2, workdir: devConfig.relayWorkdir })
   const txStoreManager = new TxStoreManager({ inMemory: true })
 
   // @ts-ignore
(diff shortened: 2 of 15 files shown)


test/RelayServer.test.js
```diff
@@ -313,21 +312,15 @@ contract('RelayServer', function (accounts) {
 
       const expectedGasPrice = (await _web3.eth.getGasPrice()) * defunctRelayServer.gasPriceFactor
       const expectedBalance = await _web3.eth.getBalance(defunctRelayServer.address)
-      const chainId = await _web3.eth.getChainId()
-      const networkId = await _web3.eth.net.getId()
       assert.notEqual(defunctRelayServer.gasPrice, expectedGasPrice)
       assert.notEqual(defunctRelayServer.balance, expectedBalance)
-      assert.notEqual(defunctRelayServer.chainId, chainId)
```
drortirosh, line 320: removed, since it implicitly assumes that `refreshStake()` can run without fully initializing the RelayServer .

src/relayclient/DevRelayClient.ts
```diff
@@ -208,19 +207,19 @@ export class DevRelayClient extends RelayClient {
...
     const stakeManager = await IStakeManagerContract.at(stakeManagerAddress)
 
-    await stakeManager.stakeForAddress(relayServer.address, weekInSec, {
-      from: this.devConfig.relayOwner,
-      value: ether('1')
-      // gas: estim
-    })
-    this.debug('== sending balance to relayServer', relayServer.address)
-    await stakeManager.authorizeHub(relayServer.address, this.config.relayHubAddress, { from: this.devConfig.relayOwner })
+    await stakeManager.contract.methods.stakeForAddress(relayServer.getManagerAddress(), unstakeDelay).send({ from: this.devConfig.relayOwner, value: ether('1') })
```
drortirosh, line 210: strangely, calling TruffleContract's stakeForAddress crashed. solved by using web3 calling

drortirosh: `Number can only safely store up to 53 bits`
wasted quite some time to understand why (there are no numbers in the request.. didn't understand where it got there)

migrations/2_deploy_contracts.js
```diff
@@ -1,12 +1,9 @@
...
-module.exports = function (deployer) {
-  deployer.deploy(RLPReader)
-  deployer.link(RLPReader, RelayHub)
-  deployer.deploy(RelayHub, 16, '0x0000000000000000000000000000000000000000', '0x0000000000000000000000000000000000000000')
-  deployer.deploy(StakeManager)
-  deployer.deploy(Penalizer)
+module.exports = async function (deployer) {
+  await deployer.deploy(StakeManager)
+  await deployer.deploy(Penalizer)
+  await deployer.deploy(RelayHub, 16, StakeManager.address, Penalizer.address)
```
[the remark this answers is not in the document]
drortirosh: We never used linking: our libs were always "internal" functions.

Thinking about it, we __could__ implement Penalizer, StakeManager as libraries, instead of contract. 
This would make the tie to the RelayHub stronger (not constructor param, but actual address in the deployed code) - at the expense of a bit more complex deployment (require linking phase)

src/relayserver/KeyManager.js
```diff
@@ -1,50 +1,83 @@
...
+const ow = require('ow')
 
 class KeyManager {
-  constructor ({ ecdsaKeyPair, workdir }) {
-    this.ecdsaKeyPair = ecdsaKeyPair
+  // workdir - read seed from keystore file (or generate one and write it)
+  // seed - if working in memory (no workdir), you can specify a seed - or use randomly generated one.
+  constructor ({ count, workdir = null, seed = null }) {
+    ow(count, ow.number)
+    if (seed && workdir) { throw new Error('Can\'t specify both seed and workdir') }
```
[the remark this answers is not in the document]
drortirosh: It is most likely a bug. the only reason to specify a seed is for deterministic testing.
you should copy the configuration file if you migrate

src/relayserver/KeyManager.js
```diff
@@ -1,50 +1,83 @@
...
-  ecSignWithPrefix ({ hash }) {
+  getAddress (index) {
+    return this.getAddresses()[index]
+  }
+
+  getAddresses () {
+    return Object.keys(this._privateKeys)
+  }
+
+  ecSignWithPrefix ({ signer, hash }) {
```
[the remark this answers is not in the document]
drortirosh: I think we don't. can be removed.

src/relayclient/RelayedTransactionValidator.ts
```diff
@@ -58,7 +58,8 @@ export default class RelayedTransactionValidator {
         console.log('validateRelayResponse - valid transaction response')
       }
 
-      const receivedNonce = transaction.nonce.readUIntBE(0, transaction.nonce.byteLength)
+      // TODO: the relayServer encoder returns zero-length buffer for nonce=0.`
+      const receivedNonce = transaction.nonce.length === 0 ? 0 : transaction.nonce.readUIntBE(0, transaction.nonce.byteLength)
```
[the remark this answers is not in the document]
drortirosh: couldn't agree with you more...

migrations/2_deploy_contracts.js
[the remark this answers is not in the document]
shahafn: I don't see a clear advantage in switching to libs here.

src/relayclient/DevRelayClient.ts
shahafn: It probably means that one of the parameters is a string instead of number, and it tries to add a number to it and it ends up being a concatenation. These kind of errors happened to me as well.

src/relayserver/KeyManager.js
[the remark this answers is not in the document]
shahafn: Yes it can be removed, leftovers from safechannels.

src/relayserver/RelayServer.js
```diff
@@ -270,39 +311,70 @@ class RelayServer extends EventEmitter {
...
 
+  fatal (message) {
+    console.error('FATAL: ' + message)
+    process.exit(1)
+  }
+
+  async _init () {
+    const relayHubAddress = this.relayHubContract.options.address
+    console.log('Server address', this.managerAddress)
+    const code = await this.web3.eth.getCode(relayHubAddress)
```
shahafn, line 322: This check seems redundant when you check version right after.

src/relayserver/RelayServer.js
```diff
@@ -270,39 +311,70 @@ class RelayServer extends EventEmitter {
...
+      console.log(`== replenishWorker(${workerIndex}): mgr balance=${this.balance.toString() / 1e18} worker balance=${workerBalance.toString() / 1e18} refill=${refill.toString() / 1e18}`)
+      if (refill.lt(this.balance.sub(toBN(minimumRelayBalance)))) {
+        await this._sendTransaction({
+          signerIndex: 0,
+          destination: workerAddress,
+          value: refill,
+          gasLimit: 300000
+          // gasPrice:1
+        })
+        this.balance = this.balance.sub(refill)
```
shahafn, line 358: You should use refreshBalance() since this doesn't take into account the tx cost other than value sent.

test/RelayServer.test.js
```diff
@@ -48,14 +50,13 @@ contract('RelayServer', function (accounts) {
...
   let id
-  let serverError
   let encodedFunction
   let relayClient
   let options, options2
   let keyManager
 
   before(async function () {
-    ethereumNodeUrl = web3.currentProvider.host
+    ethereumNodeUrl = web3.currentProvider.host || 'http://localhost:8545'
```
shahafn, line 59: Why? when is it ever not set?

test/TestUtils.ts
```diff
@@ -6,9 +6,26 @@ import { RelayHubInstance, StakeManagerInstance } from '../types/truffle-contrac
 import HttpWrapper from '../src/relayclient/HttpWrapper'
 import HttpClient from '../src/relayclient/HttpClient'
 import { configureGSN } from '../src/relayclient/GSNConfigurator'
+import { ether } from '@openzeppelin/test-helpers'
+import fs from 'fs'
 
 const localhostOne = 'http://localhost:8090'
 
+// from: https://stackoverflow.com/questions/18052762/remove-directory-which-is-not-empty
+export function cleanFolder (folder: string): void {
```
shahafn, line 15: You can simply do `fs.rmdirSync(folder, {recursive: true})`

src/relayserver/RelayServer.js
drortirosh: it is for adding better error handling: I don't like diagnostics like "no contract at address or contract is not a RelayHub" since it doesn't really help the user.

src/relayserver/RelayServer.js
drortirosh: fixed

test/RelayServer.test.js
drortirosh: leftover, removed.
 (with buidler, the provider is different. alas, there's no standard API for getting a URL, e.g. in websocket there's no "url"

test/TestUtils.ts
drortirosh: fixed.

migrations/2_deploy_contracts.js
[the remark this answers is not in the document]
drortirosh: The advantage is no configuration param, as the lib reference is hard-coded in the code.
There is not reason for us to switch now, as it adds deployment complexity, and very little security

src/relayserver/KeyManager.js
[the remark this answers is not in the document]
drortirosh: removed

PR #329, opened 2020-04-21
Implement Preferred Relays configuration support

forshtat:


src/relayclient/KnownRelaysManager.ts
```diff
@@ -27,39 +33,54 @@ const activeManagerEvents = ['RelayServerRegistered', 'TransactionRelayed', 'Can
...
+    this.httpClient = httpClient
     this.relayFilter = relayFilter ?? EmptyFilter
     this.scoreCalculator = scoreCalculator ?? DefaultRelayScore
     this.contractInteractor = contractInteractor
   }
 
   /**
-   * Iterates through all relevant logs emitted by GSN contracts
-   * initializes an array {@link activeRelays}
+   * Pings all preferred arrays to discover their manager, fee information.
```
drortirosh, line 63: ping doesn't give you fee info: for preferred relays, you don't know (and don't care) about advertised fees: they are always first
(you only use the worker (and maybe manager) you get from ping)

src/relayclient/KnownRelaysManager.ts
```diff
@@ -91,8 +112,20 @@ export default class KnownRelaysManager implements IKnownRelaysManager {
       }
     })
     const origRelays = Array.from(activeRelays.values())
-    const filteredRelays = origRelays.filter(this.relayFilter)
-    filteredRelays.forEach(relay => this.activeRelays.add(relay))
+    return origRelays.filter(this.relayFilter)
+  }
+
+  async _fetchPreferredRelayManagers (): Promise<Set<Address>> {
```
drortirosh, line 118: why "pre-ping" the preferred relays? there's no need. just accept them as valid.

drortirosh: moreover, you ping them linearly: that is, if I have 5 prefered relays, then I ping 5 times in sequence, then send one "burst" of 3 pings to select the fastest. 
that means if any prefered relay is down, I'll have time TIMEOUT before continue to the next one (in the pre-ping. the normal "raceToSuccess ping" works ok.

src/relayclient/DevRelayClient.ts
```diff
@@ -225,7 +226,7 @@ export class DevRelayClient extends RelayClient {
     const relayInfo = await waitForRelay(relayServer.url as string + '/getaddr', 5000, (res) => {
       if (res?.data?.Ready === true) {
         return {
-          ...res.data,
+          relayManager: res.data.RelayServerAddress,
           pctRelayFee: '0',
```
drortirosh, line 230: why add fees here? they never returned by a relay getaddr

src/relayclient/KnownRelaysManager.ts
```diff
@@ -134,23 +155,30 @@ export default class KnownRelaysManager implements IKnownRelaysManager {
...
+  async _sortRelaysInternal (gsnTransactionDetails: GsnTransactionDetails, activeRelays: RelayInfoUrl[]): Promise<RelayInfoUrl[]> {
     const scores = new Map<string, number>()
-    for (const activeRelay of this.activeRelays) {
-      const score = await this.scoreCalculator(activeRelay, gsnTransactionDetails, this.relayFailures.get(activeRelay.relayUrl) ?? [])
+    for (const activeRelay of activeRelays) {
+      let score = 0
+      if (isInfoFromEvent(activeRelay)) {
+        score = await this.scoreCalculator((activeRelay as RelayRegisteredEventInfo), gsnTransactionDetails, this.relayFailures.get(activeRelay.relayUrl) ?? [])
+      }
       scores.set(activeRelay.relayUrl, score)
```
drortirosh, line 173: you keep scoring URLs. its wrong. you score managers. these are unique identities (I know, unrelated to preferred relays. its an error entered in previous commit)

test/relayclient/KnownRelaysManager.test.ts
```diff
@@ -11,28 +11,34 @@ import {
...
+  await stakeManager.stakeForAddress(manager, 1000, {
     value: ether('1'),
     from: owner
   })
-  await stakeManager.authorizeHub(manager1, relayHub.address, { from: owner })
+  await stakeManager.authorizeHub(manager, relayHub.address, { from: owner })
 }
 
-async function register (stakeManager: StakeManagerInstance, relayHub: RelayHubInstance, manager: string, worker: string, url: string): Promise<void> {
+export async function register (stakeManager: StakeManagerInstance, relayHub: RelayHubInstance, manager: string, worker: string, url: string, baseRelayFee?: string, pctRelayFee?: string): Promise<void> {
```
drortirosh, line 35: nitpick: you don't use the stakeManager, so why have it as parameter?

test/relayclient/KnownRelaysManager.test.ts
```diff
@@ -168,21 +182,29 @@ contract('KnownRelaysManager 2', function (accounts) {
...
+
     it('should consider all relay managers with stake and authorization as active', async function () {
       await knownRelaysManager.refresh()
-      const relays = knownRelaysManager.getRelays()
-      assert.equal(relays.length, 2)
-      assert.equal(relays[0].relayUrl, 'stakeAndAuthorization1')
-      assert.equal(relays[1].relayUrl, 'stakeAndAuthorization2')
+      const preferredRelays = knownRelaysManager.knownRelays[0]
+      const activeRelays = knownRelaysManager.knownRelays[1]
+      assert.equal(preferredRelays.length, 1)
```
drortirosh, line 193: nitpick: a single `assert.deepEqual` can compare the entire structure

src/relayclient/types/RelayInfo.ts
```diff
@@ -1,9 +1,14 @@
 import PingResponse from '../../common/PingResponse'
-import RelayRegisteredEventInfo from './RelayRegisteredEventInfo'
+import { RelayInfoUrl, RelayRegisteredEventInfo } from './RelayRegisteredEventInfo'
 
 // Well, I still don't like it
 // Some info is known from the event, some from ping
-export default interface RelayInfo {
+export interface PartialRelayInfo {
```
drortirosh, line 6: I really don't understand your nested structure. why relayInfo has a relayInfo member?
is inheriting (and then checking that members are missing) is any better than optional fields?

src/relayclient/KnownRelaysManager.ts
```diff
@@ -27,39 +33,55 @@ const activeManagerEvents = ['RelayServerRegistered', 'TransactionRelayed', 'Can
...
 
   private latestScannedBlock: number = 0
+  private relayFailures = new Map<string, RelayFailureInfo[]>()
+
+  public readonly knownRelays: RelayInfoUrl[][] = []
 
-  constructor (contractInteractor: ContractInteractor, config: GSNConfig, relayFilter?: RelayFilter, scoreCalculator?: AsyncScoreCalculator) {
+  constructor (contractInteractor: ContractInteractor, httpClient: HttpClient, config: GSNConfig, relayFilter?: RelayFilter, scoreCalculator?: AsyncScoreCalculator) {
     this.config = config
+    this.httpClient = httpClient
```
drortirosh, line 58: You don't use it. remove the field
(also it shouldn't use it: for "separation of concerns" , the KnownRelaysManager cares only about the static (on-chain) info, not pigns)

src/relayclient/DevRelayClient.ts
forshtat: What do you mean? This is your code that adds fees, not mine. I convert ```RelayServerAddress``` to ```relayManager``` so that response confirms to the expected EventInfo interface

src/relayclient/KnownRelaysManager.ts
forshtat: Well, I will get to fix the keys to be consistent soon

test/relayclient/KnownRelaysManager.test.ts
forshtat: will fix on next commit

test/relayclient/KnownRelaysManager.test.ts
forshtat: This is about expicitness, not shortness

src/relayclient/types/RelayInfo.ts
forshtat: Optional fields remain optional all the way. So if you have
``` let a = getWithOptional()
a = fillMissingFields(a)

console.log(a.foo) 
                     ^ ... error: foo is 'string | undefined'
```
This way I convert from a class with optional fields to one with all fields assigned.

src/relayclient/KnownRelaysManager.ts
forshtat: fixed

src/relayclient/KnownRelaysManager.ts
forshtat: fixed

PR #332, opened 2020-04-23
Fix flows test

drortirosh:
- support paymaster in RelayClient config (can override in request config)
- detect forwarder from recipient (unless overridden from RelayClient options)
- extract-abi: now extracts all `contracts/interfaces`
- AsyncApprove renamed to AsyncApprovalData (matches better the approvalData solidity param)
- useGSN works (including to construct object (with a caveat that in TruffleContract, it can't be the only option, so you can use `MyContract.new({ useGSN:true, from:accounts[0] })


test/TestUtils.ts
```diff
@@ -1,13 +1,13 @@
 /* eslint-disable @typescript-eslint/strict-boolean-expressions */
 import childProcess, { ChildProcessWithoutNullStreams } from 'child_process'
 import path from 'path'
+import fs from 'fs'
```
forshtat, line 4: This file does not have meaningful changes

test/flows_test.js
```diff
@@ -12,21 +14,23 @@ const TestPaymasterPreconfiguredApproval = artifacts.require('tests/TestPaymaste
...
 const options = [
   { title: 'Direct-', relay: 0 },
   { title: 'Relayed-', relay: 1 }
 ]
 
+if (!contract.only) { contract.only = contract }
+
 options.forEach(params => {
-  contract.skip(params.title + 'Flow', async (acc) => {
+  contract.only(params.title + 'Flow', async (acc) => {
```
forshtat, line 28: .only
the CircleCI only ran 8 tests

src/relayclient/RelayProvider.ts
```diff
@@ -148,7 +148,8 @@ export default class RelayProvider implements HttpProvider {
 
   _useGSN (payload: JsonRpcPayload): boolean {
     const gsnTransactionDetails: GsnTransactionDetails = payload.params[0]
-    return gsnTransactionDetails.useGSN ?? true
+    const ret = gsnTransactionDetails?.useGSN ?? true
+    return ret
```
forshtat, line 152: This will raise a warning, and also Shahaf said this method did not work for him so waiting for him to provide fix.

contracts/interfaces/IRelayRecipient.sol
```diff
@@ -0,0 +1,5 @@
+pragma solidity ^0.5.16;
+
+contract IRelayRecipient {
```
forshtat, line 3: I would actually give it a different name, it may be confusing to people what to inherit: IRelayRecipient or BaseRelayRecipient. IForwardable or something in this manner should be better, I think.

drortirosh: Actually, they can use IRelayRecipient, without our base-class. I acutally need to bring getSender here too.
(didn't, because it need to be internal, but internals can be in interface just the same)
I don't like the name RelayRecipient - but that's another issue

src/relayclient/RelayProvider.ts
drortirosh: why warning? it passed silently our lint & tsc (and IDE)

PR #334, opened 2020-04-23
Gsn helpers 

shahafn:
Also bug fixes & start preparing to npm publish


helpers/src/data.js
```diff
@@ -0,0 +1,1822 @@
+module.exports = {
```
forshtat, line 1: How is this file generated?

helpers/src/helpers.js
```diff
@@ -0,0 +1,115 @@
+const data = require('./data')
```
forshtat, line 1: I don't like the fact we have 'helpers/src/helpers.js', knowing we also have 'helpers/index.js'

migrations/2_deploy_contracts.js
```diff
@@ -1,9 +1,11 @@
 var RelayHub = artifacts.require('./RelayHub.sol')
 var StakeManager = artifacts.require('./StakeManager.sol')
 var Penalizer = artifacts.require('./Penalizer.sol')
+var SampleRecipient = artifacts.require('./test/TestRecipient.sol')
 
 module.exports = async function (deployer) {
   await deployer.deploy(StakeManager)
   await deployer.deploy(Penalizer)
   await deployer.deploy(RelayHub, 16, StakeManager.address, Penalizer.address)
+    deployer.deploy(SampleRecipient)
```
forshtat, line 10: please await for it

package.json
```diff
@@ -1,10 +1,10 @@
 {
   "version": "0.4.5",
-  "description": "Openeth Gas Stations Network",
-  "name": "@openeth/gsn",
+  "description": "Opegsn Gas Stations Network",
```
forshtat, line 3: Are we 'Opengsn' or 'OpenGSN'? Does it event have any sense to be 'Open**GSN** **G**as **S**tations **N**etwork'? 😕 
Maybe we are 'Open Gas Stations Network'?

package.json
```diff
@@ -28,7 +28,9 @@
   "bin": {
     "gsn-stat": "./scripts/gsn-stat.js",
     "gsn-dock-relay": "./scripts/gsn-dock-relay",
-    "gsn-dock-relay-ganache": "./scripts/gsn-dock-relay-ganache"
+    "gsn-dock-relay-ganache": "./scripts/gsn-dock-relay-ganache",
```
forshtat, line 31: Maybe we don't need to pupulate bin with docker commands? This feels excessive to me, tbh.

src/relayclient/GSNConfigurator.ts
```diff
@@ -79,7 +79,7 @@ export function getDependencies (config: GSNConfig, provider?: HttpProvider, ove
   let accountManager = overrideDependencies?.accountManager
   if (accountManager == null) {
     if (provider != null) {
-      accountManager = new AccountManager(provider, defaultEnvironment.chainId, config)
+      accountManager = new AccountManager(provider, config.chainId ?? defaultEnvironment.chainId, config)
```
forshtat, line 82: ChainID is not optional in GSNConfig. It will never be null, and will never reach the `defaultEnvironment.chainId`.

src/relayclient/RelayProvider.ts
```diff
@@ -18,7 +18,7 @@ export interface BaseTransactionReceipt {
 
 export type JsonRpcCallback = (error: Error | null, result?: JsonRpcResponse) => void
 
-export default class RelayProvider implements HttpProvider {
+export class RelayProvider implements HttpProvider {
```
forshtat, line 21: Why not default?

src/relayclient/RelayProvider.ts
```diff
@@ -147,6 +147,9 @@ export default class RelayProvider implements HttpProvider {
   }
 
   _useGSN (payload: JsonRpcPayload): boolean {
+    if (payload.params[0] === undefined) {
+      return false
+    }
```
forshtat, line 152: When is it defined and when is it undefined?

helpers/src/data.js
forshtat: +1,1822 OMG

migrations/2_deploy_contracts.js
drortirosh, line 10: We should either use migration as they should (for target/production contract: the hub and friends), or don't use them at all. 
We shouldn't "migrate" test code.

package.json
```diff
@@ -149,7 +152,9 @@
     "contracts/",
     "webtools/",
     "scripts/gsn*",
-    "webtools.webpack.js"
+    "webtools.webpack.js",
+    "helpers/",
+    "dist/"
   ],
-  "main": "src/relayclient/index.js"
+  "main": "dist/src/relayclient/index.js"
```
drortirosh, line 159: is this the right "main"? we're exposing more than just the client

helpers/src/data.js
drortirosh: dont' commit it: it is built by "publish" of the package.
(also, I'd put it in a "build" folder, not in "src")
why do you change the json file to JS ? you can "require" json files just as well, and thus use truffle compiler output as-is (just move folder)

src/relayclient/GSNConfigurator.ts
shahafn: Except for all the times that it was null and reached it when I integrated with Metacoin.

helpers/src/data.js
shahafn: I took the oz approach of compiling it and putting it in data.js file, you're free to change it as well.

helpers/src/helpers.js
shahafn: I just kept the structure from oz/gsn-helpers, so probably need to rename the inner dir.

migrations/2_deploy_contracts.js
shahafn: fixed

package.json
shahafn: fixed

package.json
shahafn: I didn't work on removing docker scripts from here, maybe we can remove it.

src/relayclient/RelayProvider.ts
shahafn: Because it is then imported after compiled to js as `Gsn = require('...').default` which is ugly. Now it is imported as `Gsn = require('...').RelayProvider`.

src/relayclient/RelayProvider.ts
shahafn: When running `web3.eth.getAccounts()` for example it was undefined. Probably defined on any call with params...

package.json
```diff
@@ -14,7 +14,7 @@
     },
     {
       "name": "Yoav Weiss",
-      "email": "yoav@tabookey.com"
+      "email": "weiss.yoav@gmail.com"
```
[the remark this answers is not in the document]
shahafn: fixed

package.json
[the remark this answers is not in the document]
shahafn: Ok, didn't clean up the obsolete scripts yet. I'll remove that one but the others are still there.

migrations/2_deploy_contracts.js
shahafn: truffle is a test environment first and foremost. I don't see any issue with having it here. We don't use it yet to deploy production code.

package.json
shahafn: I don't know, just fixed it to point to dist since it is compiled by tsc to that location. We probably need to "clean" the publish files but  I didn't filter them yet.

package.json
shahafn: removed

PR #340, opened 2020-04-24
obsolete 'require('/data')

drortirosh:


helpers/src/deploy.js
```diff
@@ -1,4 +1,3 @@
-const data = require('./data')
```
shahafn, line 1: Already removed that in #339 that we just merged

PR #341, opened 2020-04-24
cmdline fix

drortirosh:
`gsn stat` for current network status
`gsn run-relayer` to start relay
  - RelayHubAddress is mandatory.

### DIFF
--- a/helpers/src/deploy.js
+++ b/helpers/src/deploy.js
@@ -58,6 +58,7 @@ async function deployRelayHub (web3, options = {}) {
     gas: 1e8,
     gasPrice: 1e9
   })
+  saveContractToFile(fInstance, options.workdir, 'Forwarder.json')
   console.log('forwarder', fInstance.options.address)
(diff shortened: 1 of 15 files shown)


helpers/src/deploy.js
```diff
@@ -52,6 +52,13 @@ async function deployRelayHub (web3, options = {}) {
     gasPrice: 1e9
   })
   console.log('paymaster ', pmInstance.options.address)
+  const forwarder = getForwarder(web3)
+  const fInstance = await forwarder.deploy({}).send({
+    from: options.from,
+    gas: 1e8,
```
shahafn, line 58: That won't work on testnet, can you reduce the gas to 1e6? we can actually have this options for all the txs as they are identical.

shahafn: And there's no need to save the contract addres for dapps to use? 
I mean `saveContractToFile(fInstance, options.workdir, 'Forwarder.json')

PR #351, opened 2020-04-26
Fixes

forshtat:


src/relayserver/RelayServer.js
```diff
@@ -574,13 +573,13 @@ class RelayServer extends EventEmitter {
    */
   async _resendUnconfirmedTransactions (blockHeader) {
     // repeat separately for each signer (manager, all workers)
-    let receipt;
-    [0, 1].forEach(signerIndex => {
-      const ret = this._resendUnconfirmedTransactionsForSigner(blockHeader, signerIndex)
+    let receipt
+    for (const signerIndex of [0, 1]) {
+      const ret = await this._resendUnconfirmedTransactionsForSigner(blockHeader, signerIndex)
```
shahafn, line 578: This bug is so hard to detect, when you return a promise and you await for `_resendUnconfirmedTransactions` when you call it. If you have only one unconfirmed tx then you won't ever notice it.

src/relayclient/RelayProvider.ts
```diff
@@ -79,15 +79,19 @@ export class RelayProvider implements HttpProvider {
       console.log('calling sendAsync' + JSON.stringify(payload))
     }
     this.origProviderSend(payload, (error: Error | null, rpcResponse?: JsonRpcResponse): void => {
-      if (error != null) {
-        callback(error)
+      // Sometimes, ganache seems to return 'false' for 'no error' (breaking TypeScript declarations)
+      // eslint-disable-next-line @typescript-eslint/strict-boolean-expressions
+      if (error) {
```
drortirosh, line 84: ok, but we might introduce `isTruthy` function - as a replacement for the above ugly eslint-disable ...

src/relayclient/RelayProvider.ts
```diff
@@ -79,15 +79,19 @@ export class RelayProvider implements HttpProvider {
...
       }
       if (rpcResponse == null || rpcResponse.result == null) {
-        throw new Error('Empty JsonRpcResponse with no error message')
+        console.error('Empty JsonRpcResponse with no error message')
+        callback(error, rpcResponse)
+        return
       }
       rpcResponse.result = this._getTranslatedGsnResponseResult(rpcResponse.result)
-      callback(null, rpcResponse)
+      callback(error, rpcResponse)
```
drortirosh, line 94: true, though its known to be null (or false) at this point - and in order to protect your caller, might be better to pass up null...

helpers/src/deploy.js
```diff
@@ -66,7 +66,8 @@ async function deployRelayHub (web3, options = {}) {
     relayHubAddress: rInstance.options.address,
     stakeManagerAddress: sInstance.options.address,
     penalizerAddress: pInstance.options.address,
-    forwarderAddress: fInstance.options.address
+    forwarderAddress: fInstance.options.address,
```
drortirosh, line 69: the increasingly-inaccurately-named `deployRelayHub` ...

PR #358, opened 2020-05-02
separate GsnDevProvider

drortirosh:
to import the GsnDevProvider, you need to do:
`import { GsnDevProvider } from '@opengsn/gsn/dist/GsnDevProvider`

This prevent it from being imported automatically with `@opengsn/gsn`,
since it can't be imported at all in a browser contract

### DIFF
--- a/GsnDevProvider.ts
+++ b/GsnDevProvider.ts
@@ -1,1 +1,0 @@
-export * from './src/relayclient/GsnDevProvider'
--- a/GsnDevProvider.ts
+++ b/GsnDevProvider.ts
@@ -1,0 +1,1 @@
+export * from './src/relayclient/GsnDevProvider'
--- a/index.ts
+++ b/index.ts
@@ -1,2 +1,3 @@
 export * from './RelayProvider'
 export * from './RelayClient'
+export * from './GSNConfigurator'
--- a/index.ts
+++ b/index.ts
@@ -1,5 +1,2 @@
 export * from './RelayProvider'
 export * from './RelayClient'
-
-//not included: clients should ask for it explicitly (require '@opengsn/gsn/dist/GsnDevProvider')
-//export * from './GsDevProvider'
--- a/index.ts
+++ b/index.ts
@@ -1,3 +1,5 @@
-export * from './src/relayclient/RelayProvider'
-export * from './src/relayclient/GsnDevProvider'
-export * from './src/relayclient/GSNConfigurator'
+export * from './RelayProvider'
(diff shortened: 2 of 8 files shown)


GsnDevProvider.ts
```diff
@@ -0,0 +1 @@
+export * from './src/relayclient/GsnDevProvider'
```
forshtat, line 1: wait, all these files are in the project root? Why?

index.ts
```diff
@@ -1,3 +1,5 @@
-export * from './src/relayclient/RelayProvider'
-export * from './src/relayclient/GsnDevProvider'
-export * from './src/relayclient/GSNConfigurator'
+export * from './RelayProvider'
+export * from './RelayClient'
+
+//not included: clients should ask for it explicitly (require '@opengsn/gsn/dist/GsnDevProvider')
+//export * from './GsDevProvider'
```
forshtat, line 5: No need for this comment

GsnDevProvider.ts
drortirosh: I know. didn't want them there.
But I wanted the usage string to be the simplest
`require( 'opengsn/gsn/dist/GsnDevProvider' )`
(and actually I wanted to get rid of "dist" too...)

do you know another way to make it?
I can copy them into the root of the package during "pack", but then the paths must be broken in the git checked in version, so this seemed like a fair tradeoff..

drortirosh: removed those file (by generating them in the `prepublish`

scripts/prepublish
```diff
@@ -1,12 +1,23 @@
...
-OUTFOLDER=./helpers/compiled/
-for c in $CONTRACTS ; do cp $BUILDFOLDER/$c.json $OUTFOLDER ; done
+OUTFOLDER=./helpers/compiled
+for c in $CONTRACTS; do
+  echo copy $OUTFOLDER/$c.json
+  cp $BUILDFOLDER/$c.json $OUTFOLDER
+done
+
+
+for c in GsnDevProvider RelayProvider RelayClient; do
```
forshtat, line 18: Two questions:
1. Why do we need these files for RelayProvider and RelayClient, they are exported as index.js, I thought only GsnDevProvider creates a problem for us?
2. Could you please add these files to .gitignore as well?

drortirosh: Technically, you can "import all" and its OK. I wanted to allow "surgical import" of inner contracts, without requiring to know the entire src path.

PR #359, opened 2020-05-03
Command line tools

forshtat:


src/cli/commands/gsn-paymaster-fund.ts
```diff
@@ -0,0 +1,36 @@
...
+import { gsnCommander } from '../utils'
+
+gsnCommander(['n', 'f', 'h'])
+  .option('--paymaster <address>',
+    'address of the paymaster contract (defaults to address from build/gsn/Paymaster.json if exists')
+  .option('--amount <amount>', 'amount of funds to deposit for the paymaster contract, in wei (defaults to 1 Ether)')
+  .parse(process.argv);
+
+(async () => {
+  const nodeURL = program.ethereumNodeURL !== undefined ? program.ethereumNodeURL : 'http://localhost:8545'
```
drortirosh, line 17: nitpick (but looks much better): 
`program.ethereumNodeURL ?? 'http://localhost:8545'`

src/cli/commands/gsn-paymaster-fund.ts
```diff
@@ -0,0 +1,36 @@
...
+
+  const relayHubDeployInfo = fs.readFileSync('build/gsn/RelayHub.json').toString()
```
drortirosh, line 19: this will crash if not found, but you have a default that doesn't use it a little later..

src/cli/StatusLogic.ts
```diff
@@ -0,0 +1,70 @@
...
+  private readonly httpClient: HttpClient
+  private readonly config: StatusConfig
+
+  constructor (contractInteractor: ContractInteractor, httpClient: HttpClient, config: StatusConfig) {
+    this.contractInteractor = contractInteractor
+    this.httpClient = httpClient
+    this.config = config
+  }
+
+  async gatherStatistics (): Promise<Statistics> {
```
drortirosh, line 36: I don't think its a good approach to lump everything into one method.
- it can only be used for one purpose - the status cmd
- its too slow (many round-trips) for any other purpose (e.g. gui statistics)
- timeout of any request will break the entire structure, and you'll have to start all over (rare, but does happen)
- no way to do operations in parallel, even those unrelated.

notice that you don't return any info, until the last relay request times out (and I assume that once there are more than 1-2 relays, some relays will be "zombies" and time-out.

src/relayclient/ContractInteractor.ts
```diff
@@ -221,6 +221,22 @@ export default class ContractInteractor {
     return contract.getPastEvents('allEvents', Object.assign({}, options, { topics }))
   }
 
+  async getPastLogs (eventName: EventName, options: PastEventOptions): Promise<Log[]> {
```
drortirosh, line 224: good start, but:
- no test
- not much of use without parsing params.

tsconfig.json
```diff
@@ -33,8 +34,9 @@
     ]
   },
   "include": [
-    "**/*.ts"
-  ],
+    "**/*.ts",
+    "src/**/*"
```
drortirosh, line 38: why? are there any non `*.ts` files that are actually typescript?

patches/@truffle+hdwallet-provider+1.0.34.patch
```diff
@@ -0,0 +1,24 @@
+diff --git a/node_modules/@truffle/hdwallet-provider/dist/index.d.ts b/node_modules/@truffle/hdwallet-provider/dist/index.d.ts
```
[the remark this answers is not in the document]
forshtat: Not yet. I will create pull requests with relevant patches when I have time. There is a huge lag between submitting a PR to these packages and getting any attention anyways.

scripts/prepublish
```diff
@@ -2,11 +2,11 @@
 rm -rf ./dist/
 yarn extract_abi
 yarn generate
-mkdir -p helpers/compiled/
```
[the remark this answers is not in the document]
forshtat: This was not my decision and I did not add it. I will change the location to whatever you say it should be, I do not feel passionate about it.

helpers/gsn-balance.js
```diff
@@ -1,21 +0,0 @@
-const program = require('commander')
```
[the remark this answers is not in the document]
forshtat: * I will not attempt to salvage the git history of this particular change because I literally have something better to do
* I will attempt to cherish the git history of all files moving forward

patches/web3-eth-contract+1.2.6.patch
```diff
@@ -0,0 +1,13 @@
+diff --git a/node_modules/web3-eth-contract/types/index.d.ts b/node_modules/web3-eth-contract/types/index.d.ts
+index 6f141ab..1207009 100644
+--- a/node_modules/web3-eth-contract/types/index.d.ts
++++ b/node_modules/web3-eth-contract/types/index.d.ts
+@@ -78,7 +78,7 @@ export interface Options extends ContractOptions {
+ }
+ 
+ export interface DeployOptions {
+-    data: string;
```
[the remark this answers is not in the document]
forshtat: I now use 'patches' folder as a ad-hoc list of pull requests I want to submit once I have time. 
Also, I would need to do this weird 'null as string' on every call to '.deploy', which is not pretty.

src/cli/commands/gsn-paymaster-fund.ts
forshtat: I forgot that I created a helper for this... Fixed.

src/cli/StatusLogic.ts
forshtat: I did not want to revive the status tool initially, but then decided it was important enough to have it working. I will work on improving the code and performance of it later, at least now it works for 'one relay on ganache' use-case, which is what we target in the ongoing hackathon.

tsconfig.json
forshtat: No, but there are still .js files and they will not be copied to /dist/ unless statically imported (i.e. const a = require('runRelay.js') will not make runRelay.js be copied to 'dist'

patches/@truffle+hdwallet-provider+1.0.34.patch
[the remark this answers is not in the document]
drortirosh: a lag is a good reason to push such changes early...

patches/web3-eth-contract+1.2.6.patch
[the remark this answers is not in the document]
drortirosh: the point is that deploy requires data - its not optional, so changing the library (instead of changing your code) is not the right solution.
 its your code that says "I know my var is nullable, but I also know at this point that I checked for null"
you could actually wrap with `if data!=null ()` and it will recognize it.

src/relayclient/GsnTest.ts
```diff
@@ -0,0 +1,57 @@
...
+   *
+   * @param host
+   * @param paymaster TODO: will allow using custom paymaster (need to provide ABI file contents)
+   * @return
+   */
+  async start (host?: string, paymaster?: any): Promise<void> {
+    const _host: string = host ?? networks.get('localhost') ?? ''
+    const commandsLogic = new CommandsLogic(_host, configureGSN({}))
+    const from = await commandsLogic.findWealthyAccount()
+    if (from == null) {
```
drortirosh, line 27: findWealthyAccount already throws. redundant check.

src/relayclient/GsnTest.ts
```diff
@@ -0,0 +1,57 @@
...
+import { configureGSN } from './GSNConfigurator'
+import { DevGSNConfig } from './DevRelayClient'
+import { networks } from '../cli/utils'
+
+interface TestEnvironment {
+  deployment: DeploymentResult
+  devProvider: GsnDevProvider
+}
+
+class GsnTestClass {
```
drortirosh, line 14: not sure what it is, and why its in the relayclient. if its a test fixture - then move to "test"

src/relayclient/GsnTest.ts
```diff
@@ -0,0 +1,57 @@
...
+  private fixture?: TestEnvironment
+
+  /**
+   *
+   * @param host
+   * @param paymaster TODO: will allow using custom paymaster (need to provide ABI file contents)
+   * @return
+   */
+  async start (host?: string, paymaster?: any): Promise<void> {
```
drortirosh, line 23: I like to call this pattern `init()`: its a workaround  for Javascript's deficiency of constructor, which can't be asynchrounous. 
so you create a class with `GsnTest gsntest = await new GsnTest(...).init()`
(with/without params to init)

BTW: in this specific case, you can actually do factory function, since you don't have constructor params at all..)

test/GsnTest.test.ts
```diff
@@ -0,0 +1,17 @@
+import GsnTest from '../src/relayclient/GsnTest'
```
drortirosh, line 1: why do you use the "fixture" as instance, and not class ?
I thinkg that adding gsnTest = new GsnTest() makes it clearer (and also let you initialize different fixtures)

src/cli/CommandsLogic.ts
```diff
@@ -0,0 +1,252 @@
...
+    const response = await this.httpClient.getPingResponse(relayUrl)
+    return response.Ready
+  }
+
+  async waitForRelay (relayUrl: string): Promise<void> {
+    const timeout = 30
+    console.error(`Will wait up to ${timeout}s for the relay to be ready`)
+
+    for (let i = 0; i < timeout; ++i) {
+      await sleep(1000)
```
drortirosh, line 95: you always wait 1 second, regardless if the relay is already up.

src/cli/CommandsLogic.ts
```diff
@@ -0,0 +1,234 @@
...
+
+    try {
+      console.error(`Funding GSN relay at ${options.relayUrl}`)
+
+      const response = await this.httpClient.getPingResponse(options.relayUrl)
+      const relayAddress = response.RelayManagerAddress
+      const relayHub = await this.contractInteractor._createRelayHub(this.config.relayHubAddress)
+      const stakeManagerAddress = await relayHub.getStakeManager()
+      const stakeManager = await this.contractInteractor._createStakeManager(stakeManagerAddress)
+      const stakeTx = await stakeManager
```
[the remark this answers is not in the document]
drortirosh: everywhere the amount we specify are "target" amount, (e.g. fund, above)
why here "Stake" is the amount to ADD ?

src/cli/CommandsLogic.ts
```diff
@@ -0,0 +1,234 @@
...
+        error: `Failed to fund relay: '${error}'`
+      }
+    }
+  }
+
+  contract (file: CompiledContract): Contract {
+    return new this.web3.eth.Contract(file.abi, undefined, { data: file.bytecode })
+  }
+
+  async deployRelayHub (from: Address, gasPrice?: string, paymaster?: CompiledContract): Promise<DeploymentResult> {
```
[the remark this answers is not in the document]
drortirosh: rename method

src/cli/commands/gsn-relayer-run.ts
```diff
@@ -0,0 +1 @@
+require('../../relayserver/runServer')
```
drortirosh, line 1: english words speak order please
(`gsn-run-relayer`, otherwise, its yoda-english.. )

src/cli/commands/gsn.ts
```diff
@@ -0,0 +1,13 @@
+#!/usr/bin/env node
+
+import commander from 'commander'
+
+commander
+  .version('0.0.1')
+  .command('deploy', 'deploy RelayHub and other GSN contracts instances')
+  .command('relayer-register', 'stake for a relayer and fund it')
+  .command('relayer-run', 'launch a relayer server')
+  .command('paymaster-fund', 'fund a paymaster contract so it can pay for relayed calls')
```
drortirosh, line 10: why long name? what bad with `gsn fund --paymaster=` ?

src/relayclient/GsnTestEnvironment.ts
```diff
@@ -0,0 +1,148 @@
...
+    const _host: string = getNetworkUrl(host)
+    if (_host == null) {
+      throw new Error(`Failed to connect to ${host}`)
+    }
+    const commandsLogic = new CommandsLogic(_host, configureGSN({}))
+    const from = await commandsLogic.findWealthyAccount()
+    if (from == null) {
+      throw new Error('could not get unlocked account with sufficient balance')
+    }
+    const deploymentResult = await commandsLogic.deployRelayHub(from, undefined, paymaster)
```
drortirosh, line 43: why not keep `deploymentInfo` as a member? 
we can `save()` it, and then later `load()` and use the same contracts

src/cli/commands/gsn.ts
forshtat: Nothing is wrong, we can rename parameters at any time

src/cli/commands/gsn-relayer-run.ts
forshtat: DM

src/cli/CommandsLogic.ts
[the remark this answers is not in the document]
forshtat: Most of commands semantics are inherited from OZ

src/relayclient/GsnTestEnvironment.ts
forshtat: Added to backlog

src/cli/CommandsLogic.ts
forshtat: Fixed.

test/GsnTest.test.ts
forshtat: This is an experiment in node-controlled singletons. I want to allow only one instance to exist simultaneously.

forshtat: There is no valid reason to hold two

src/relayclient/GsnTest.ts
forshtat: Used subset of suggested API in your e-mail.

src/relayclient/GsnTest.ts
forshtat: Made 'findWealthyAccount' not throw. returning undefined is much cleaner, type-safe API imo

src/cli/CommandsLogic.ts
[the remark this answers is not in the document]
forshtat: Fixed

src/cli/CommandsLogic.ts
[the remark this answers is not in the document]
forshtat: Added to backlog

src/cli/CommandsLogic.ts
```diff
@@ -0,0 +1,234 @@
...
+import Paymaster from './compiled/TestPaymasterEverythingAccepted.json'
+import Forwarder from './compiled/TrustedForwarder.json'
+
+import { Address } from '../relayclient/types/Aliases'
+import ContractInteractor from '../relayclient/ContractInteractor'
+import { GSNConfig } from '../relayclient/GSNConfigurator'
+import HttpClient from '../relayclient/HttpClient'
+import HttpWrapper from '../relayclient/HttpWrapper'
+
+interface CompiledContract {
```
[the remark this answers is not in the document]
forshtat: I will use default parameter later.
It is not defined like that anywhere I looked.
Removed anyways.

src/relayclient/ContractInteractor.ts
forshtat: Added to backlog

src/cli/StatusLogic.ts
forshtat: Added to backlog.

src/cli/commands/gsn-paymaster-fund.ts
forshtat: Fixed

patches/web3-eth-contract+1.2.6.patch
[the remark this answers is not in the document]
forshtat: It is not that.
Data parameter is optional.
If you construct the contract object with a bytecode and ABI

PR #363, opened 2020-05-05
fix relayer webpack

drortirosh:
and added initial test for the webpacking.


test/RelayServer.webpack.test.js
```diff
@@ -0,0 +1,27 @@
...
+  })
+
+  it('should launch (and say "missing address")', () => {
+    try {
+      childProcess.execSync('node ' + oneFileRelayer, { encoding: 'ascii' })
+      assert.fail('should throw')
+    } catch (e) {
+      assert.match(e.toString(), /missing --RelayHubAddress/)
+    }
+    // just verify it launches
```
forshtat, line 23: is it a TODO or an explanation of some sort?

test/RelayServer.webpack.test.js
```diff
@@ -0,0 +1,27 @@
+// eslint-disable-next-line no-unused-vars
```
forshtat, line 1: Please use TypeScript

test/RelayServer.webpack.test.js
```diff
@@ -0,0 +1,27 @@
...
+import childProcess from 'child_process'
+import path from 'path'
+import fs from 'fs'
+
+describe.only('RelayServer-webpack', () => {
```
forshtat, line 6: Only

test/RelayServer.webpack.test.js
```diff
@@ -0,0 +1,27 @@
...
+    try {
+      childProcess.execSync('node ' + oneFileRelayer, { encoding: 'ascii' })
+      assert.fail('should throw')
+    } catch (e) {
+      assert.match(e.toString(), /missing --RelayHubAddress/)
+    }
+    // just verify it launches
+  })
+
+  it('should test it can actually work')
```
forshtat, line 26: it really should

drortirosh: That's a separate issue to repeat  server suite test with this one-file relayer.
this test only checks that the webpack was successful (in the cleanroom of CI)

the point is that we don't bundle this 4MB json file into our package (its a merge of all installed packages used by the server) - though it can be exported as a single-file-package of a server.

test/RelayServer.webpack.test.js
forshtat: Please use TypeScript for files you create. I don't want to add a single line of JavaScript to the project.

PR #364, opened 2020-05-06
Scaffold eth support

forshtat:
1. Do not lint './dist/' folder
2. Always prioritize 'sendAsync' over 'send'
3. Use 'Common.forCustomChain' method to create chain configuration for
Tranasctions objects
4. Increase log verbosity
5. In devMode server forces excecution of 'worker' every 8 seconds
6. Unconfirmed transactions logic relies on past blockchain state no more
7. Fix bug with case-sensitive address in tx store manager
8. Allow 'runServer' to read from deployment files


src/relayclient/RelayProvider.ts
```diff
@@ -40,7 +44,11 @@ export class RelayProvider implements HttpProvider {
 
     this.origProvider = origProvider
     this.config = config
-    this.origProviderSend = this.origProvider.send.bind(this.origProvider)
+    if (typeof this.origProvider.sendAsync === 'function') {
```
drortirosh, line 47: hmm. sendAsync was removed in web3-provider>1.0. so buidler implements the old protocol?

src/relayserver/RelayServer.js
```diff
@@ -286,7 +288,11 @@ class RelayServer extends EventEmitter {
         console.error('web3 subscription:', error)
       }
     }).on('data', this._workerSemaphore.bind(this)).on('error', (e) => { console.error('worker:', e) })
-    setTimeout(() => { this.web3.eth.getBlockNumber().then(blockNumber => this._workerSemaphore.bind(this)({ number: blockNumber })) }, 1)
+    const handler = () => { this.web3.eth.getBlockNumber().then(blockNumber => this._workerSemaphore.bind(this)({ number: blockNumber })) }
```
drortirosh, line 291: over-clutter the logs. 
run handler only if `blockNumber!= this.lastScannedBlock`

src/relayserver/runServer.js
```diff
@@ -13,6 +13,23 @@ function error (err) {
   process.exit(1)
 }
 
+/// TODO: remove once translated to TypeScript. This method is in ./cli/utils.ts
+function getRelayHubAddress (hub) {
```
drortirosh, line 17: you ignore the function's logic if a parameter is given?
why not `const hub = commander.hub ?? getRelayHubAddress()`

(who approved that cli.ts ?)

src/common/utils.js
```diff
@@ -2,7 +2,13 @@ const ethUtils = require('ethereumjs-util')
 const web3Utils = require('web3-utils')
 
 const abi = require('web3-eth-abi')
+
+const { default: Common } = require('ethereumjs-common')
+
 function removeHexPrefix (hex) {
+  if (hex == null || typeof hex.replace !== 'function') {
```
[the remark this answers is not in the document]
forshtat: Let's talk about it later. It is not broken.

src/relayclient/RelayProvider.ts
forshtat: web3-provider-engine enforces calling to 'sendAsync'

src/relayserver/RelayServer.js
forshtat: This is temporary. Shahaf will take over and fix these issues

src/relayserver/runServer.js
forshtat: Yes, you're right. Will fix. On next PR.

src/common/utils.js
```diff
@@ -157,5 +170,23 @@ module.exports = {
...
+  /**
+   * Ganache does not seem to enforce EIP-155 signature. Buidler does, though.
+   * This is how {@link Transaction} constructor allows support for custom and private network.
+   * @param chainId
+   * @param networkId
+   * @return {{common: Common}}
+   */
+  getRawTxOptions (chainId, networkId) {
+    return {
+      common: Common.forCustomChain(
```
shahafn, line 184: Why is it "common:" and not "chain:"?

src/relayserver/RelayServer.js
```diff
@@ -286,7 +288,11 @@ class RelayServer extends EventEmitter {
...
+    if (this.devMode) {
```
shahafn, line 292: Every 8 seconds in devMode? Are you sure?

src/relayserver/RelayServer.js
```diff
@@ -609,22 +625,41 @@ class RelayServer extends EventEmitter {
...
+      const receipt = await this.web3.eth.getTransaction(transaction.txId)
+      if (receipt == null) {
+        // I believe this means this transaction was not confirmed
+        continue
+      }
+      const txBlockNumber = receipt.blockNumber
+      const confirmations = blockHeader.number - txBlockNumber
+      if (confirmations >= confirmationsNeeded) {
+        debug(`removing tx number ${receipt.nonce} sent by ${receipt.from} with ${confirmations} confirmations`)
+        await this.txStoreManager.removeTxByNonce({
```
shahafn, line 650: This should be `await this.txStoreManager.removeTxsUntilNonce({ signer: receipt.from,  nonce: receipt.nonce })`

src/relayserver/TxStoreManager.js
```diff
@@ -78,7 +78,7 @@ class TxStoreManager {
     return this.txstore.asyncRemove({
       $and: [
         { 'nonceSigner.nonce': nonce },
-        { 'nonceSigner.signer': signer }]
+        { 'nonceSigner.signer': signer.toLowerCase() }]
```
shahafn, line 81: Actually then you need toLowerCase() all insertions as well I think.

src/relayserver/RelayServer.js
```diff
@@ -609,22 +625,41 @@ class RelayServer extends EventEmitter {
...
     // Clear out all confirmed transactions (ie txs with nonce less than the account nonce at confirmationsNeeded blocks ago)
-    await this.txStoreManager.removeTxsUntilNonce({ signer, nonce: nonce - 1 })
-
+    // await this.txStoreManager.removeTxsUntilNonce({
+    //   signer,
+    //   nonce: nonce - 1
+    // })
+
+    for (const transaction of sortedTxs) {
+      const receipt = await this.web3.eth.getTransaction(transaction.txId)
```
shahafn, line 641: You don't have to fix now, but this is technically the tx itself and not the receipt.

src/relayserver/TxStoreManager.js
drortirosh: better `web3.utils.toChecksumAddress()` on the signer itself, instead of lowering it all the time.

src/common/utils.js
forshtat: It is the name from 'ethereumjs-tx':
```
function Transaction(data, opts) {
        if (data === void 0) { data = {}; }
        if (opts === void 0) { opts = {}; }
        // instantiate Common class instance based on passed options
        if (opts.common) {
            if (opts.chain || opts.hardfork) {
                throw new Error('Instantiation with both opts.common, and opts.chain and opts.hardfork parameter not allowed!');
            }
            this._common = opts.common;
        }
        else {
            var chain = opts.chain ? opts.chain : 'mainnet';
            var hardfork = opts.hardfork ? opts.hardfork : 'petersburg';
            this._common = new ethereumjs_common_1.default(chain, hardfork);
        }
```

src/relayserver/RelayServer.js
forshtat: otherwise it just spams everything with logs

src/relayserver/RelayServer.js
forshtat: Changed.

src/relayserver/TxStoreManager.js
forshtat: Will lowercase. this is not a good place to fail on checksum

src/relayserver/RelayServer.js
forshtat: Won't fix

src/relayserver/RelayServer.js
shahafn: This is only a buidler solution, it won't work on any testnet as this doesn't fix the websocket provider issue that we have.

src/common/utils.js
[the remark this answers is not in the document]
drortirosh: I know. that's why its only a "nitpic"

src/relayserver/RelayServer.js
drortirosh: of course it spams. but if you check and call _worker only if blocknumber changes, it won't spam at all...

PR #365, opened 2020-05-07
startGsn cleanup

drortirosh:
1.  export GsnTestEnvironment. can be used as
    ```javascript
    env = require( '@opengsn/gsn/dist/GsnTestEnvironment').default
    env.startGsn('local')
    ```
2. now report an error on null, unknown network or non-url

### DIFF
--- a/package.json
+++ b/package.json
@@ -42,7 +42,7 @@
     "lint:fix": "yarn run lint:js:fix",
     "lint:js": "eslint --ext .js,.ts --cache -f unix --ignore-path .gitignore src test scripts",
     "lint:sol": "solhint -f unix \"contracts/**/*.sol\" --max-warnings 0",
-    "lint:js:fix": "eslint --ext .js,.ts --cache  -f unix --ignore-path .gitignore src test scripts helpers --fix",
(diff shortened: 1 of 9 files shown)


package.json
```diff
@@ -51,6 +51,7 @@
     "tsc": "tsc --noEmit",
     "postinstall": "patch-package",
     "coverage-if-changed": "./scripts/run-coverage-if-changed.sh npm run coverage",
+    "postpack": "./scripts/postpack",
     "prepare": "./scripts/prepublish"
```
drortirosh, line 55: do we need this "prepare" for anything other than pack? if not, it can be renamed to "prepack", and then it runs only before pack..

src/relayclient/GsnTestEnvironment.ts
```diff
@@ -70,6 +72,7 @@ class GsnTestEnvironmentClass {
     })
 
     const relayProvider = new RelayProvider(new Web3.providers.HttpProvider(_host), config)
+    console.error('== startGSN: ready.')
```
drortirosh, line 75: this is the only important log for a user - when everything is up. others logs should be silenced in normal cases.

PR #366, opened 2020-05-07
Compile in solidity 0.6.2

drortirosh:
compiles. not all tests pass.
NOTE: a contract that uses both GSN and OpenZeppelin contract (e.g. Ownable) requires some hoop to work:

```
contract TestProxy is BaseRelayRecipient, Ownable  {

    function _msgSender() internal override(Context, BaseRelayRecipient) view returns (address payable) {
        return BaseRelayRecipient._msgSender();
    }
```


drortirosh:
NOTE: requires patching LibBytes
- change package version
- they have broken soldoc which breaks compilation, so changed "@return" to "return"

(also possible to copy out just `readBytes32` and `readAddress`)

forshtat:
closed in favour of https://github.com/opengsn/gsn/pull/367

PR #367, opened 2020-05-07
Migrate GSN to solidity 0.6.2

forshtat:
Note: 0x LibBytes was checked-in here


contracts/interfaces/IRelayRecipient.sol
```diff
@@ -1,17 +1,17 @@
-pragma solidity ^0.5.16;
+pragma solidity ^0.6.2;
 
 /**
  * a contract must implement this interface in order to support relayed transaction.
  * It is better to inherit the BaseRelayRecipient as its implementation.
  */
-contract IRelayRecipient {
+abstract contract IRelayRecipient {
```
shahafn, line 7: why not interface?

package.json
```diff
@@ -30,7 +30,7 @@
   },
   "prepublish": "npm run prepare",
   "scripts": {
-    "test": "yarn run lint && yarn extract_abi && yarn generate && yarn tsc && npm run test-js",
+    "test": "yarn run lint && yarn run truffle-compile && yarn extract_abi && yarn generate && yarn tsc && npm run test-js",
```
shahafn, line 33: 1. `yarn generate` already compiles.
2. If you're at it, then it's weird that we have both `yarn run` and `npm run`, even in the same line.

contracts/interfaces/IRelayRecipient.sol
forshtat: A) it was a contract
B) I think it has to call this._msgSender() for interface

drortirosh: Interface can't declare"internal" methods

drortirosh: Interface can't declare"internal" methods

drortirosh: Interface can't declare"internal" methods

Issue #372, opened 2020-05-14
Warning: Using ".value(...)" is deprecated. Use "{value: ...}" instead.


naszam:
Hi,
I'm playing with opengsn contracts for a project at the HackMoney hackathon to implement a Paymaster contract.
When inheriting BasePaymaster.sol and trying to compile:
```
$ truffle compile
```
I get the following error:
```
    @opengsn/gsn/contracts/BasePaymaster.sol:69:9: Warning: Using ".value(...)" is deprecated. Use "{value: ...}" instead.
```
I would recommend to update to a recent solc version and update the contracts.

Using the following dependences:
Ubuntu 18.04
npm v13.13.0
truffle v5.1.25
solc v0.6.7

forshtat:
https://github.com/opengsn/gsn/pull/376

Issue #371, opened 2020-05-14
Update to OpenZeppelin v3.0 


naszam:
Hi, 
I'm playing with opengsn contracts for a project at the HackMoney hackathon to implement a Paymaster contract.
When inheriting BasePaymaster.sol and trying to compile:
```
$ truffle compile
```
I get the following error:
```
Error: Could not find openzeppelin-solidity/contracts/access/Ownable.sol from any sources; imported from @opengsn/gsn/contracts/BasePaymaster.sol
```
I would recommend to update the contracts to OpenZeppelin v3.0.

Using the following dependences:
Ubuntu 18.04
npm v13.13.0
truffle v5.1.25
solc v0.6.7

forshtat:
https://github.com/opengsn/gsn/pull/376

PR #373, opened 2020-05-15
Update to OpenZeppelin v3.0 + fix .value() with {value:...} when inheriting BasePaymaster.sol

naszam:
Updated import to OpenZeppelin v3.0  in BasePaymaster.sol and  EIP712Sig.sol.
Changed .value() with {value:...} in BasePaymaster.sol

See Issues: #371, #372

### DIFF
--- a/contracts/BasePaymaster.sol
+++ b/contracts/BasePaymaster.sol
@@ -1,7 +1,7 @@
 pragma solidity ^0.6.2;
 pragma experimental ABIEncoderV2;
 
-import "openzeppelin-solidity/contracts/access/Ownable.sol";
+import "@openzeppelin/contracts/access/Ownable.sol";
 
 import "./interfaces/IPaymaster.sol";
 import "./interfaces/IRelayHub.sol";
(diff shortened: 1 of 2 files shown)


forshtat:
@naszam is there any benefit in the changes you've made here? These seem to be the exact same package with a different name and just an alternative syntax.

naszam:
@forshtat It compiles when inheriting BasePaymaster.sol.

forshtat:
> @forshtat It compiles when inheriting BasePaymaster.sol.
> It's just a quick fix for devs that want to implement a Paymaster contract.
> It doesn't cover all the contracts.

Not sure I understand. Does it not compile for you without these changes?

naszam:
@forshtat Nope! just with this fix.

forshtat:
> @forshtat Yep! It compiles only with this fix, using truffle.
> But I guess all the contracts should be updgraded to OpenZeppelin v3.0 and obsolete syntax for .value() should be fixed with {value:...}

Got you. Like, all the ouptput of the compilation is just warnings, and some OZ contracts are still generating warnings.

Anyways, updated all contracts in https://github.com/opengsn/gsn/pull/373#issuecomment-629814943

PR #374, opened 2020-05-15
Fix everything paymaster

drortirosh:


contracts/test/TestPaymasterEverythingAccepted.sol
```diff
@@ -51,19 +51,17 @@ contract TestPaymasterEverythingAccepted is BasePaymaster {
         emit SampleRecipientPostCall(success, gasUseWithoutPost, preRetVal);
     }
 
-    // TODO: this 'overrides' existing setRelayHub method
     function setHub(IRelayHub _relayHub) public {
-        relayHub = _relayHub;
+        setRelayHub(_relayHub);
     }
```
forshtat, line 56: I don't get it. Why do we need this broken override in the first place? Can't we remove it and just keep the 'setRelayHub'?

drortirosh: I wanted a minimal commit. setHub is spread all over the place (and luckily it is always called by the owner, so it works)

PR #378, opened 2020-05-20
Bug fixes

shahafn:


src/relayserver/RelayServer.ts
```diff
@@ -108,43 +106,42 @@ export interface RelayServerParams {
...
+  readonly hubAddress: PrefixedHexString
+  readonly baseRelayFee: number
+  readonly pctRelayFee: number
+  readonly gasPriceFactor: number
+  readonly url: string
   private readonly workerMinBalance: number
   private readonly workerTargetBalance: number
   private readonly devMode: boolean
-  private readonly web3: Web3
   private workerTask: any
```
forshtat, line 145: Side note: this is an insane amount of fields for a class of this size. No way these do not deserve being split into groups like 'config', 'network', 'stuff' etc.

Also, why did you need to make so many of them public here?

PR #379, opened 2020-05-24
GSN-10 Remove 'canRelay' from the public API of RelayHub

forshtat:
There is no need for 'canRelay' as 'relayCall' can be called directly in
a view-only mode and this is equivalent and does not require any extra
roundrips.

Also, separates 'getTrustedForwarder' into a separate API.


contracts/BaseRelayRecipient.sol
```diff
@@ -23,8 +23,8 @@ contract BaseRelayRecipient is IRelayRecipient {
         _;
     }
 
-    function getTrustedForwarder() public override view returns(address) {
-        return trustedForwarder;
+    function isTrustedForwarder(address forwarder) public override view returns(bool) {
```
drortirosh, line 26: do you insist on bundling  "remove canRelay" with "remove getTrustedForwarder" ? I thnik they are separate issues.

contracts/Penalizer.sol
```diff
@@ -84,7 +84,7 @@ contract Penalizer {
         if (decodedTx.to == address(hub)) {
             bytes4 selector = GsnUtils.getMethodSig(decodedTx.data);
             require(
-                selector != hub.relayCall.selector,
+                selector != IRelayHub.relayCall.selector,
```
drortirosh, line 87: excelent. I was trying to look for a way to reference a method without having an instance! is it new in sol6?

contracts/RelayHub.sol
```diff
@@ -232,50 +203,41 @@ contract RelayHub is IRelayHub {
...
-                    relayRequest.relayData.senderAddress,
-                    relayRequest.target,
-                    relayRequest.relayData.paymaster,
-                    functionSelector,
-                    recipientContext);
-                return;
-            }
+        if (!success) {
+            revertReason = GsnUtils.getError(vars.recipientContext);
+            emit UnpaidPaymasterRejection(
```
drortirosh, line 232: success is "TransactionRelayed".
so failure should be "TxFailed", or TransactionRejectedByPaymaster"

contracts/TrustedForwarder.sol
```diff
@@ -28,15 +28,14 @@ contract TrustedForwarder is ITrustedForwarder {
...
+    {
         _verify(req, sig);
         _updateNonce(req);
 
         // solhint-disable-next-line avoid-low-level-calls
-        return req.target.call{gas:req.gasData.gasLimit}
-        (abi.encodePacked(req.encodedFunction, req.relayData.senderAddress));
+        (bool success, bytes memory returnValue) = req.target.call{gas:req.gasData.gasLimit}(abi.encodePacked(req.encodedFunction, req.relayData.senderAddress));
+        // TODO: use assembly to prevent double-wrapping of the revert reason (part of GSN-37)
+        require(success, string(returnValue));
```
drortirosh, line 38: `GsnUtils.getError(returnValue)`

contracts/interfaces/IRelayHub.sol
```diff
@@ -41,12 +41,12 @@ interface IRelayHub {
     // The actual relayed call was not executed, and the recipient not charged.
     // The reason field contains an error code: values 1-10 correspond to CanRelayStatus entries, and values over 10
     // are custom recipient error codes returned from acceptRelayedCall.
-    event CanRelayFailed(
+    event UnpaidPaymasterRejection(
```
drortirosh, line 44: This event is matched again the TransactionRelayed. they should matching indexed params.

contracts/interfaces/IKnowForwarderAddress.sol
```diff
@@ -0,0 +1,12 @@
+// SPDX-License-Identifier:MIT
+pragma solidity ^0.6.2;
+
+interface IKnowForwarderAddress {
```
drortirosh, line 4: Do you think its such an esoteric case that a recipient defines its own forwarder? I think it is part of the recipient, maybe rename it to getDefaultForwarder()

contracts/interfaces/IRelayRecipient.sol
```diff
@@ -8,11 +8,11 @@ pragma solidity ^0.6.2;
 abstract contract IRelayRecipient {
 
     /**
-     * return the forwarder we trust to forward relayed transactions to us.
+     * return if the forwarder is trusted to forward relayed transactions to us.
      * the forwarder is required to verify the sender's signature, and verify
      * the call is not a replay.
      */
-    function getTrustedForwarder() public virtual view returns(address);
```
drortirosh, line 15: I'd leave here the getter method, maybe just named `getDefaultForwarder()`

src/relayclient/ContractInteractor.ts
```diff
@@ -256,8 +257,9 @@ export default class ContractInteractor {
     return this.web3.eth.getGasPrice()
   }
 
-  async getTransactionCount (address: string, defulatBlock?: BlockNumber): Promise<number> {
-    return this.web3.eth.getTransactionCount(address)
+  async getTransactionCount (address: string, defaultBlock?: BlockNumber): Promise<number> {
+    // @ts-ignore (web3 does not define 'defaultBlock' as optional)
+    return this.web3.eth.getTransactionCount(address, defaultBlock)
```
drortirosh, line 262: nitpic: I'd prefer doing `defaultBlock as BlockNumber` instead of "ts-ignore" - since "ts-ignore" ignores everything, not only one error type

src/relayclient/RelayClient.ts
```diff
@@ -157,8 +158,14 @@ export default class RelayClient {
...
       await this._prepareRelayHttpRequest(relayInfo, gsnTransactionDetails)
     const acceptRelayCallResult = await this.contractInteractor.validateAcceptRelayCall(relayRequest, signature, approvalData)
-    if (!acceptRelayCallResult.success) {
-      return { error: new Error(`canRelay failed: ${acceptRelayCallResult.returnValue}`) }
+    if (!acceptRelayCallResult.paymasterAccepted) {
+      let message: string
+      if (acceptRelayCallResult.reverted) {
+        message = 'local view call to \'relayCall()\' reverted'
+      } else {
+        message = 'paymaster rejected in local view call to \'relayCall()\' '
```
drortirosh, line 166: lost reject reason... should report it as well.

src/relayclient/RelaySelectionManager.ts
```diff
@@ -74,6 +74,7 @@ export default class RelaySelectionManager {
             relayInfo: events[0]
           }
         } else {
+          // TODO: do not throw! The preffered relay may be removed since.
```
drortirosh, line 77: I don't understand the sentence...

src/relayclient/types/GsnTransactionDetails.ts
```diff
@@ -14,9 +14,9 @@ export default interface GsnTransactionDetails {
   gas?: PrefixedHexString
   gasPrice?: PrefixedHexString
 
-  // Required parameters for GSN:
-  readonly forwarder: Address
-  readonly paymaster: Address
+  // Required parameters for GSN, but assigned later
```
drortirosh, line 17: not required per transaction... can be set in the provider

test/relayclient/RelayClient.test.ts
```diff
@@ -211,7 +223,8 @@ contract('RelayClient', function (accounts) {
 
       // @ts-ignore (sinon allows spying on all methods of the object, but TypeScript does not seem to know that)
       sinon.spy(dependencyTree.knownRelaysManager)
-      await relayClient._attemptRelay(relayInfo, optionsWithGas)
+      const attempt = await relayClient._attemptRelay(relayInfo, optionsWithGas)
+      assert.equal(attempt.error?.message, 'some error describing how timeout occurred somewhere')
```
drortirosh, line 227: huh? where did this error came from?

contracts/BaseRelayRecipient.sol
forshtat: I don't, but the 'canRelay' used to call 'get forwarder' so it seemed to be related

contracts/Penalizer.sol
forshtat: Yes, this is solc v6

contracts/RelayHub.sol
forshtat: We agreed on this name some time ago in the chat.

contracts/TrustedForwarder.sol
forshtat: Should I?

contracts/interfaces/IRelayHub.sol
forshtat: For idealogical reasons? Failure and success may need to highlight different aspects of a transaction. In this case, it may be critical to be able to easily get all transactions rejected by a specific paymaster.

contracts/interfaces/IKnowForwarderAddress.sol
forshtat: Discussed over email. Cannot use all existing channels for the same discussion.

src/relayclient/ContractInteractor.ts
forshtat: I dream of a day when I will go over our ts-ignores and open PRs to relevant repos to fix their typings. This day is near, I just need to keep faith.

src/relayclient/RelayClient.ts
forshtat: no I didn't, look literally 1 line below

src/relayclient/RelaySelectionManager.ts
forshtat: If you define a preferred relay in a client but relay is removed we want to handle this situation gracefully. Now this will throw (not sure what happens next)

forshtat: GSN-38

src/relayclient/types/GsnTransactionDetails.ts
forshtat: We used to allow overrides

test/relayclient/RelayClient.test.ts
forshtat: This is from a stubbed method, don't worry!

src/relayclient/ContractInteractor.ts
drortirosh: no, I meant how to write a code with cast instead of ts-ignore. but its a nitpic - better writing int he future..

contracts/interfaces/IRelayRecipient.sol
forshtat: I realized your point. I really did.

contracts/RelayHub.sol
drortirosh: The reason for this event is matching TransactionRelayed and TransactionFailed to tell if a there's a possibility of attack, by someone. That's why its important to make these events close to each other (both in names, and with indexed params)

src/relayclient/ContractInteractor.ts
forshtat: ok, I will. But ```var_A: Something | undefined as Something``` is equivalent to ```var_A!```, which we forbid in linter for being not good.

contracts/TrustedForwarder.sol
drortirosh: yes, you should. assembly is optimization. using getError to actually get the error is bugfix.

forshtat: Fixed.

contracts/interfaces/IRelayRecipient.sol
drortirosh: yes, I think `getDefaultForwarder()` should be here, instead of introducing yet another single-function interface.

contracts/RelayHub.sol
forshtat: TransactionRelayed is a wrong name in such case, as it indicates success but this is not the case. Will not rename now. Can be streamlined in a separate commit.

contracts/Penalizer.sol
shahafn: Yes I tried but solc 5 didn't let me when writing it...

contracts/RelayHub.sol
drortirosh: (also, I don't remember seeing this name agreed upon in any chat)
@shahafn ?

src/relayclient/RelaySelectionManager.ts
shahafn: If you (dapp developer) defined a preferred relay, then it most likely means that this is your relay. If you remove it, then you should also update the client's code.

src/relayclient/ContractInteractor.ts
```diff
@@ -137,65 +151,52 @@ export default class ContractInteractor {
   }
 
   async getForwarder (recipientAddress: Address): Promise<Address> {
-    const recipient = await this._createRecipient(recipientAddress)
+    const recipient = await this._createKnowsForwarder(recipientAddress)
```
drortirosh, line 154: this function will crash if the client doesn't support the interface.
if you require clients to opt-in for interface to support it, at least give a proper error incase it doesn't.

src/relayclient/RelayClient.ts
```diff
@@ -260,4 +260,28 @@ export default class RelayClient {
...
+      try {
+        if (this.config.verbose) {
+          console.log(`will attempt to get trusted forwarder from: ${gsnTransactionDetails.to}`)
+        }
+        forwarderAddress = await this.contractInteractor.getForwarder(gsnTransactionDetails.to)
+        if (this.config.verbose) {
+          console.log(`on-chain forwarder for: ${gsnTransactionDetails.to} is ${forwarderAddress}`)
+        }
+      } catch (e) {
+        throw new Error('No forwarder address configured and fetching from Recipient failed')
```
drortirosh, line 281: Please update the error: `No forwarder address configured and no getTrustedForwarder in target contract`

forshtat: Done

PR #380, opened 2020-05-24
Update server version

shahafn:
Also removing unused tmp legacy client, adding workerSemaphore test, needs rename since it's not really a semaphore...

Need to update npm package since the server version and .env were updated. Also, to fix the docs to include the gsn-helpers for staking and funding the server, we need post PR #378 npm package, which is currently not the case...


test/RelayServer.test.ts
```diff
@@ -751,8 +749,8 @@ contract('RelayServer', function (accounts) {
     beforeEach(function () {
       origWorker = relayServer._worker
       started = false
-      // @ts-ignore
-      relayServer._worker = function () {
+      relayServer._worker = async function () {
+        await Promise.resolve()
```
drortirosh, line 753: what does `await Promise.resolve()` does here?

shahafn: Only prevents linter from complaining that it's async. But it has to be async, otherwise `.then` get thrown in the server.

PR #381, opened 2020-05-26
Cleanup repo from deprecated files and dependencies

forshtat:
solves #382


docs/protocol.md
```diff
@@ -1,230 +0,0 @@
-# The GSN Protocol
```
drortirosh, line 1: why remove, and not update?

package.json
```diff
@@ -36,21 +36,14 @@
     "extract_abi": "./scripts/extract_abi.js",
     "truffle-test": "yarn tsc && npx truffle test",
     "truffle-compile": "truffle compile",
-    "webtools-pack": "webpack-cli --config ./webtools.webpack.js",
```
drortirosh, line 39: how can this run? we have a test to pack the jsrelay.

package.json
```diff
@@ -59,89 +52,72 @@
...
-    "@0x/typescript-typings": "5.0.2",
-    "@machinomy/types-ethereumjs-wallet": "^0.0.12",
+    "@typescript-eslint/parser": "3.0.0",
     "@openeth/truffle-typings": "0.0.5",
     "@truffle-types/abi-decoder": "^0.1.1",
     "@typescript-eslint/eslint-plugin": "2.23.0",
-    "bignumber.js": "9.0.0",
-    "browser-request": "0.3.3",
     "chai-as-promised": "7.1.1",
+    "cors": "2.8.5",
```
drortirosh, line 104: can startGsn work? this is used by relay, so if you remove it, relays won't be able to start

src/relayserver/TxStoreManager.ts
```diff
@@ -66,7 +65,7 @@ export const TXSTORE_FILENAME = 'txstore.db'
 export class TxStoreManager {
   private readonly txstore: AsyncNedb<any>
   constructor ({ workdir = '/tmp/test/', inMemory = false }) {
-    this.txstore = new Nedb({
+    this.txstore = new AsyncNedb({
```
drortirosh, line 68: not realy equivalent. one is default, the other is named member

package.json
```diff
@@ -59,89 +52,72 @@
...
     "web3-eth": "1.2.6",
     "web3-eth-abi": "1.2.6",
     "web3-eth-contract": "1.2.6",
-    "web3-utils": "1.2.6",
-    "webpack-bundle-analyzer": "3.3.2"
+    "web3-utils": "1.2.6"
   },
   "devDependencies": {
-    "@0x/sol-compiler": "4.0.8",
     "@0x/sol-coverage": "4.0.8",
```
drortirosh, line 95: if remove, the remove all unused 0x packages.

docs/protocol.md
forshtat: It will never be well-maintained. We have https://github.com/opengsn/docs.opengsn.org for documentation in version control

package.json
forshtat: this is not jsrelay, this is webtools

package.json
forshtat: this is just adding a dependency that we used but did not declare directly (it just happened to be there because some other project used it)

src/relayserver/TxStoreManager.ts
forshtat: Not sure about that 

import AsyncNedb from 'nedb-async'
and 
const Nedb = require('nedb-async').AsyncNedb

seem to use the same class to me. If not, this is extremely confusing.

Also, tests pass.

package.json
forshtat: these were indicated to be unused by depcheck

PR #388, opened 2020-05-31
Rebased gas limits

forshtat:


contracts/RelayHub.sol
```diff
@@ -179,46 +166,44 @@ contract RelayHub is IRelayHub {
...
     }
+
     function relayCall(
     // TODO: msg.sender used to be treated as 'relay' (now passed in a struct),
     //  make sure this does not have security impl
         ISignatureVerifier.RelayRequest calldata relayRequest,
         bytes calldata signature,
-        bytes calldata approvalData
+        bytes calldata approvalData,
+        uint externalGasLimit
```
forshtat, line 182: Probably cannot be left unconstrained. Will add require(externalGasLimit  < blockGasLimit) to prevent some relay draining more than it's stake

test/RelayHubGasCalculations.test.ts
```diff
@@ -23,28 +23,18 @@ const TestRecipient = artifacts.require('TestRecipient')
...
-  for (let i = 0; i < buffer.length; i++) {
-    if (buffer[i] === 0) {
-      gasCost += zerocost
-    } else {
-      gasCost += nonzerocost
-    }
-  }
-  return gasCost
-}
+require('source-map-support').install({ errorFormatterForce: true })
```
forshtat, line 26: put it in truffle.js, otherwise tests that run before this will probably still have distorted stack traces

test/RelayHubGasCalculations.test.ts
```diff
@@ -281,39 +261,50 @@ contract('RelayHub gas calculations', function ([_, relayOwner, relayWorker, rel
...
+              const workerWeiGasUsed = beforeBalances.relayWorkers.sub(afterBalances.relayWorkers)
+
+              // sanity: worker executed and paid this tx
               // @ts-ignore (this types will be implicitly cast to correct ones in JavaScript)
-              assert.equal((res.receipt.gasUsed * gasPrice).toString(), weiGasUsed.toString(), 'where else did the money go?')
+              assert.equal((res.receipt.gasUsed * gasPrice).toString(), workerWeiGasUsed.toString(), 'where else did the money go?')
+
+              // const txRelayed:any = res.logs.find(e=>e.event=='TransactionRelayed')
+              // assert.equal(res.receipt.gasUsed, txRelayed.args.gasUsed.toNumber(),
+              //     'GAS_OVERHEAD: add='+(res.receipt.gasUsed-txRelayed.args.gasUsed.toNumber()))
```
forshtat, line 285: \--

test/RelayHubPenalizations.test.ts
```diff
@@ -40,7 +40,7 @@ contract('RelayHub Penalizations', function ([_, relayOwner, relayWorker, otherR
   before(async function () {
```
forshtat, line 40: Missing test for wrong gas limit parameter penalization... and penalizer change as well

test/RelayHubPenalizations.test.ts
```diff
@@ -250,7 +250,7 @@ contract('RelayHub Penalizations', function ([_, relayOwner, relayWorker, otherR
           const baseFee = new BN('300')
           const fee = new BN('10')
           const gasPrice = new BN('1')
-          const gasLimit = new BN('1000000')
+          const gasLimit = new BN('5000000')
```
forshtat, line 253: why?

contracts/RelayHub.sol
drortirosh: how can it drain anything? it doesn't really matter, as far as the EVM cares. its the real gas limit that determine the upper bound.

contracts/RelayHub.sol
```diff
@@ -228,29 +213,34 @@ contract RelayHub is IRelayHub {
                 msg.sender,
                 vars.functionSelector,
                 revertReason);
-            return (success, revertReason);
+            return (vars.success, revertReason);
         }
 
-        // From this point on, this transaction will not revert nor run out of gas, and the recipient will be charged
+        // From this point on, this transaction will not revert nor run out of gas, and the paymaster will be charged
```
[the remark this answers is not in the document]
drortirosh: that's exactly why I like them - you can look at the code and at the comment together (yes, you batch-renamed it in the past and missed it, but the code is alive, and others are reading it.
embedded docs has greater chance of getting fixed...

contracts/RelayHub.sol
```diff
@@ -228,29 +213,34 @@ contract RelayHub is IRelayHub {
...
+        (, bytes memory relayCallStatus) = address(this).call{gas:innerGasLimit}(
+            abi.encodeWithSelector(RelayHub.innerRelayCall.selector, relayRequest, signature, vars.gasLimits,
+                innerGasLimit + externalGasLimit-gasleft(),
+                abi.decode(vars.recipientContext, (bytes)))
         );
+    
         vars.status = abi.decode(relayCallStatus, (RelayCallStatus));
-
+    }
+    {
```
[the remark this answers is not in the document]
drortirosh: not all. there are 2 local variables, one in each scope... I didn't want to move really local vars into the struct, which is essentially global to the entire method's scope

test/RelayHubPenalizations.test.ts
drortirosh: the penalizer itself is still missing.

test/RelayHubPenalizations.test.ts
drortirosh: i think it was too low.

contracts/RelayHub.sol
forshtat: I am a relay server
I set externalGasLimit = 1e20 or whatever

uint256 gasUsed = (externalGasLimit - gasleft()) + GAS_OVERHEAD

charge will be enormous

contracts/RelayHub.sol
```diff
@@ -179,46 +166,44 @@ contract RelayHub is IRelayHub {
     }
 
     struct RelayCallData {
-        uint256 initialGas;
+        bool success;
```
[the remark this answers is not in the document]
drortirosh: adding the success into the structure? apparently yes... I hate it as much as you.
AFAIK, it is possible to remove it, with yet another `{}` closure ...

contracts/RelayHub.sol
drortirosh: indeed. needs `require(externalGasLimit<tx.gaslimit)` (upper "attack" limit near 8M gas)
(I wanted `require(externalGasLimit-gasleft()<val)`, but we depend on current calldata pricing)

forshtat: Done

PR #391, opened 2020-06-02
Disable contracts as workers

forshtat:


shahafn:
This is disturbing. I added this check a long time ago. When was it removed and why?

PR #392, opened 2020-06-02
Generic forwarder

drortirosh:
First stage of generic forwarder:
- changed the RelayRequest structure to be "generic-forwarder-friendly": all major fields are flat in the RR struct (from,to,func,gaslimit,nonce)


contracts/SignatureVerifier.sol
```diff
@@ -23,13 +23,13 @@ contract SignatureVerifier is ISignatureVerifier{
     );
 
     // solhint-disable-next-line max-line-length
-    bytes32 public constant RELAY_REQUEST_TYPEHASH = keccak256("RelayRequest(address target,bytes encodedFunction,GasData gasData,RelayData relayData)GasData(uint256 gasLimit,uint256 gasPrice,uint256 pctRelayFee,uint256 baseRelayFee)RelayData(address senderAddress,uint256 senderNonce,address relayWorker,address paymaster,address forwarder)");
+    bytes32 public constant RELAY_REQUEST_TYPEHASH = keccak256("RelayRequest(address target,bytes encodedFunction,address senderAddress,uint256 senderNonce,uint256 gasLimit,address forwarder,GasData gasData,RelayData relayData)GasData(uint256 gasPrice,uint256 pctRelayFee,uint256 baseRelayFee)RelayData(address relayWorker,address paymaster)");
 
     // solhint-disable-next-line max-line-length
-    bytes32 public constant CALLDATA_TYPEHASH = keccak256("GasData(uint256 gasLimit,uint256 gasPrice,uint256 pctRelayFee,uint256 baseRelayFee)");
+    bytes32 public constant CALLDATA_TYPEHASH = keccak256("GasData(uint256 gasPrice,uint256 pctRelayFee,uint256 baseRelayFee)");
```
forshtat, line 29: Let's move gas price to the Forwarder responsibility as well. Then all remaining fields will be clearly GSN-specific

contracts/Eip712Forwrader.sol
```diff
@@ -0,0 +1,109 @@
...
+import "./utils/GsnUtils.sol";
+
+// a Generic EIP712 forwarder.
+// actual struct has to START with known fields, but may contain other fields
+contract Eip712Forwarder {
+    using ECDSA for bytes32;
+
+    //all valid requests must start with this prefix.
+    // (note that the prefix is not a valid type by itself: it doesn't end with ")" (or ",", if more fields are added)
+    string constant typehashPrefix = "Forward(address sender, address target, bytes encodedFunction, uint256 gasLimit, address forwarder";
```
forshtat, line 15: Can we extract it to be a struct? So instead of 'gasData' and 'relayData' we have 'ForwarderData' and 'GSNData'?

contracts/Eip712Forwrader.sol
```diff
@@ -0,0 +1,109 @@
...
+    }
+
+    function verifyAndCall(Forward memory req,
+        bytes32 domainSeparator, bytes32 requestTypeHash, bytes memory suffixData, bytes memory sig) public {
+        _verifyNonce(req);
+        _verifySig(req, domainSeparator, requestTypeHash, suffixData, sig);
+        _updateNonce(req);
+
+        // solhint-disable-next-line avoid-low-level-calls
+        (bool success, bytes memory returnValue) = req.target.call{gas : req.gasLimit}(abi.encodePacked(req.encodedFunction, req.sender));
```
forshtat, line 49: I think it is forwarder's responsibility to make sure exactly 'gasLimit' gas is available to the recipient code. So should verify that `gasleft - gaslimit > 1/64` or whatever, don't you think this is a reasonable expectation from the forwarder with signed gas limit?

contracts/Eip712Forwrader.sol
forshtat: Also, missing nonce, right?

forshtat: And, use original Ethereum names. Together, this structure is 
```from to data gas gasPrice nonce```, only missing ```value```

Also, we can include 'value' in this forwarder. If the paymaster sends ether here, it should accept it and make it available for 'verifyAndCall', what do you think?

Comment (forshtat) [reply contracts/Eip712Forwrader.sol:15]: Also, we could close that bracket if we make it a suffix, not a prefix (requires it to be last alphabetically)
```RelayRequest(ForwarderData fd, GSNData gd)GSNData(pctRelayFee,baseRelayFee)ZZZForwarderData(from,to,data,gas,gasPrice,nonce)```

Comment (drortirosh) [reply contracts/Eip712Forwrader.sol]: 1. its deliberately prefix, since these are supposed to be the most important fields. putting it last, means you can add as much garbage at the beginning.
2. making it a struct is problematic: the parsing will become much more complex. the prefix is "RelayData data" and then the "RelayData" is defined somewhere - in a sorted list of types...
AAATypename and ZZZTypename are not good either...

I want to use the original names, I just dislike "data" as it seems to me more of an anachronism than intended name. "method" or "function" is better.

File: contracts/Eip712Forwrader.sol (hunk 3 above)
Comment (drortirosh) [reply contracts/Eip712Forwrader.sol]: yes, gas limit should be here. but I don't think we should put into it the EVM-specific calculation. passing a value as-is is good enough

File: contracts/Eip712Forwarder.sol [hunk 4]
```diff
@@ -0,0 +1,150 @@
...
+        //sanity: parameters should not end parameters block
+        for (uint i = 0; i < bytes(extraParams).length; i++) {
+            require(bytes(extraParams)[i] != ")", "invalid extraParams");
+        }
+        require(!typeHashes[requestTypehash], "typehash already registered");
+        typeHashes[requestTypehash] = true;
+        emit RequestTypeRegistered(requestTypehash, string(requestType));
+    }
+
+    function isRegisteredTypehash(bytes32 typehash) public view returns (bool) {
```
forshtat, line 105: How is this method expected to be used?

forshtat: Should it be part of '_verifySig' on-chain?

drortirosh: its there for completeness. you can re-register at will, and it will ignore previous run (since they both do the same: save boolean in a hash) I added it because I don't like to rely on public members - these are exposing internal structs to tests at best, and can't be considered an API.

patches/@openzeppelin+contracts+3.0.1.patch
```diff
@@ -0,0 +1,63 @@
+diff --git a/node_modules/@openzeppelin/contracts/GSN/Context.sol b/node_modules/@openzeppelin/contracts/GSN/Context.sol
```
forshtat, line 1: they released 3.0.2

contracts/test/TestPaymasterPreconfiguredApproval.sol
```diff
@@ -13,12 +13,12 @@ contract TestPaymasterPreconfiguredApproval is TestPaymasterEverythingAccepted {
...
-        ISignatureVerifier.RelayRequest calldata relayRequest,
-        bytes calldata signature,
-        bytes calldata approvalData,
+        ISignatureVerifier.RelayRequest memory relayRequest,
+        bytes memory signature,
+        bytes memory approvalData,
         uint256 maxPossibleGas
     )
-    external
+    public
```
forshtat, line 21: Why?

drortirosh: solc abiv2...
(I'll check if solc0.6.9 can compile, and if so, I'll revert it)

drortirosh: fixed

test/RelayHubGasCalculations.test.ts
```diff
@@ -14,16 +14,18 @@ import {
...
+const Eip712Forwarder = artifacts.require('Eip712Forwarder')
 const StakeManager = artifacts.require('StakeManager')
 const Penalizer = artifacts.require('Penalizer')
 const TestRecipient = artifacts.require('TestRecipient')
 const TestPaymasterVariableGasLimits = artifacts.require('TestPaymasterVariableGasLimits')
 const TestPaymasterConfigurableMisbehavior = artifacts.require('TestPaymasterConfigurableMisbehavior')
 
-contract('RelayHub gas calculations', function ([_, relayOwner, relayWorker, relayManager, senderAddress, other]) {
+// TMP: skip until we finish the new Forwarder
+contract.skip('RelayHub gas calculations', function ([_, relayOwner, relayWorker, relayManager, senderAddress, other]) {
```
forshtat, line 28: Don't think we should merge with these skipped. This test turned out to be an extreme PITA though

test/Utils.test.ts
```diff
@@ -64,9 +86,62 @@ contract('Utils', function (accounts) {
...
+      assert.equal(ret.domainSeparator, bufferToHex(TypedDataUtils.hashStruct('EIP712Domain', dataToSign.domain, dataToSign.types)))
+
+      // const digest = keccak256(bufferToHex(Buffer.concat([
+      //   Buffer.from("\x19\x01"), domainSeparator, keccak256(encodedForSig)
+      // ].map(toBuffer))))
+
+      // possible exceptions:
+      //  "invalid request typehash" - missing register type with relayHub.registerRequestType(forwarder)
+      //  "invalid nonce"
+      // "signature mismatch" signature error - wrong signer/domain/struct
```
forshtat, line 108: this is a lot of commenting-out

contracts/utils/GsnEip712Library.sol
```diff
@@ -0,0 +1,124 @@
...
+
+    /**
+     * call forwarder.verifyAndCall()
+     * note that we call it with address.call, and return (success,ret): this helper is a library
+     * function, and library function can't be wrapped with try/catch... (or called with address.call)
+     */
+    function callForwarderVerifyAndCall(GsnTypes.RelayRequest memory req, bytes memory sig) internal returns (bool success, bytes memory ret) {
+        (Eip712Forwarder.ForwardRequest memory fwd, bytes memory suffixData) = splitRequest(req);
+        // Eip712Forwarder forwarder = Eip712Forwarder(req.extraData.forwarder);
+        // forwarder.verifyAndCall(fwd, req.extraData.domainSeparator, RELAY_REQUEST_TYPEHASH, suffixData, sig);
```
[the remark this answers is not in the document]
drortirosh: removed.

contracts/utils/GsnEip712Library.sol
```diff
@@ -0,0 +1,124 @@
...
+        return req.extraData.forwarder.call(abi.encodeWithSelector(IForwarder.verifyAndCall.selector,
+            fwd, req.extraData.domainSeparator, RELAY_REQUEST_TYPEHASH, suffixData, sig
+        ));
+    }
+
+    function domainSeparator(address forwarder) internal pure returns (bytes32) {
+        return hashDomain(EIP712Domain({
+            name : "GSN Relayed Transaction",
+            version : "2",
+            chainId : 1234, // getChainID(),
```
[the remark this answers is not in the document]
drortirosh: it works if I hard-code the same value everywhere (both sol and ts)...
removed.

contracts/utils/GsnEip712Library.sol
```diff
@@ -0,0 +1,124 @@
...
+            verifyingContract : forwarder
+            }));
+    }
+
+    function getChainID() internal pure returns (uint256) {
+        uint256 id;
+        /* solhint-disable no-inline-assembly */
```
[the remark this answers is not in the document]
drortirosh: fixed (though I thnik you added to RelayHub all those solhint-disable at the beginning of the file ..)

migrations/2_deploy_contracts.js
```diff
@@ -2,10 +2,12 @@ var RelayHub = artifacts.require('./RelayHub.sol')
...
 var SampleRecipient = artifacts.require('./test/TestRecipient.sol')
+var Eip712Forwarder = artifacts.require('Eip712Forwarder')
 
 module.exports = async function (deployer) {
   await deployer.deploy(StakeManager)
   await deployer.deploy(Penalizer)
   await deployer.deploy(RelayHub, StakeManager.address, Penalizer.address)
-  await deployer.deploy(SampleRecipient)
+  await deployer.deploy(Eip712Forwarder)
+  await deployer.deploy(SampleRecipient, Eip712Forwarder.address)
```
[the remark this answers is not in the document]
drortirosh: its just a refactor of SampleRecipient, which now receives a forwrader instead of creating one inside the constructor (which is a bad habbit anyway)
yes, we should remove them - and have some proper deployment utility (e.g. buidler?!?!)

patches/@openzeppelin+contracts+3.0.1.patch
drortirosh: removed (though I don't remember I ever touched this file..)

scripts/extract_abi.js
```diff
@@ -72,7 +72,9 @@ function compileFile (contractFile, c) {
 }
 */
 
-fs.readdirSync(contractsFolderToExtract).forEach(file => {
+const files = fs.readdirSync(contractsFolderToExtract)
+files.push('IForwarder.sol')
+files.forEach(file => {
```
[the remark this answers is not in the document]
drortirosh: get used to disappointments..

src/relayserver/RelayServer.ts
```diff
@@ -25,6 +25,7 @@ import { configureGSN } from '../relayclient/GSNConfigurator'
 import { defaultEnvironment } from '../relayclient/types/Environments'
 import VersionsManager from '../common/VersionsManager'
 import { calculateTransactionMaxPossibleGas } from '../common/Utils'
+import { extraDataWithDomain } from '../common/EIP712/ExtraData'
```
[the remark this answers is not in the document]
drortirosh: think of a better way. we must pass in the forwarder param, and also the domainSeparator (which is not constant - depends on forwader itself, and on chain-id).
so it was logical to have a method to create the entire 2-field struct..

test/Eip712Forwarder.test.ts
```diff
@@ -0,0 +1,336 @@
...
+//  given a complete EIP712TypedData, validate it extends the ForwardRequest type,
+//  and extract the strings required for registering this type.
+function getRegisterParams (data: EIP712TypedData, genericParams: string, genericType: string): RegisterTypeParams {
+  // before placing the string into a regex, we need to escape special chars
+  function escaped (s: string): string {
+    return s.replace(/([()])/g, '\\$1')
+  }
+
+  const type = TypedDataUtils.encodeType(data.primaryType, data.types)
+  const m = new RegExp(`(\\w+)\\(${genericParams}(?:,(.+))?\\)(.*)${escaped(genericType)}(.*)`).exec(type)
```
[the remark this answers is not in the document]
drortirosh: I verify that a given EIP712 type is valid extension for a Generic forwarder - following the rules of extending the type. I first create the full EIP712 string, then I make sure it starts with our "genericParams", and has "genericType" in the typelist.

Sorry. I don't have any other language to express such expression better (maybe YACC or ANTLR, but I won't bring in a compiler-compiler to parse a string)
Its a regular expression that looks for a sequence of:
- word
- open-parenthesis
- "generic params",
- some optional extra stuff (more params) before close parenthesis - either "?" and stuff, or empty string.
- close parenthesis
- "generic type" (with optional data before/after)
- m is "matcher" - the output of regex execution.

contracts/RelayHub.sol
```diff
@@ -169,6 +170,9 @@ contract RelayHub is IRelayHub {
             relayRequest.relayData.paymaster.staticcall{gas:gasLimits.acceptRelayedCallGasLimit}(encodedTx);
     }
 
+    function registerRequestType(Eip712Forwarder forwarder) public override {
+        GsnEip712Library.registerRequestType(forwarder);
+    }
```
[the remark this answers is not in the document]
forshtat: You don't need to call library, you need to call forwarder. I don't like adding a method that is basically one-time usage only.

contracts/RelayHub.sol
```diff
@@ -229,17 +233,17 @@ contract RelayHub is IRelayHub {
         // errors in the recipient. In either case (revert or regular execution) the return data encodes the
         // RelayCallStatus value.
         (, bytes memory relayCallStatus) = address(this).call{gas:innerGasLimit}(
-            abi.encodeWithSelector(RelayHub.innerRelayCall.selector, relayRequest, signature, vars.gasLimits,
+            abi.encodeWithSelector(this.innerRelayCall.selector, relayRequest, signature, vars.gasLimits,
```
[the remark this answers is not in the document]
forshtat: a. No it doesn't

contracts/RelayHub.sol
```diff
@@ -268,13 +272,13 @@ contract RelayHub is IRelayHub {
...
-        IPaymaster.GasLimits calldata gasLimits,
+        GsnTypes.RelayRequest memory relayRequest,
+        bytes memory signature,
+        IPaymaster.GasLimits memory gasLimits,
         uint256 totalInitialGas,
-        bytes calldata recipientContext
+        bytes memory recipientContext
     )
-    external
+    public
```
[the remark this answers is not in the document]
forshtat: where exactly? let's debug that. you are passing calldata to memory or vice versa. not a reason to change API.

contracts/forwarder/Eip712Forwarder.sol
```diff
@@ -0,0 +1,145 @@
...
+    string public constant GENERIC_TYPE = "_ForwardRequest(address to,bytes data,address from,uint256 nonce,uint256 gas)";
+    bytes32 public constant GENERIC_TYPEHASH = keccak256(bytes(GENERIC_TYPE));
+
+    mapping(bytes32 => bool) public typeHashes;
+
+    // Nonces of senders, used to prevent replay attacks
+    mapping(address => uint256) private nonces;
+
+    function versionForwarder() external view virtual override returns (string memory) {
+        return "2.0.0-alpha.2+opengsn.forwarder.eip712";
```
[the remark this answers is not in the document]
forshtat: Nobody is handling these strings on-chain, Dror.

contracts/test/TestPaymasterPreconfiguredApproval.sol
forshtat: No it's not fixed. ARC should be external.

contracts/RelayHub.sol
```diff
@@ -188,15 +192,15 @@ contract RelayHub is IRelayHub {
     returns (bool paymasterAccepted, string memory revertReason)
     {
         RelayCallData memory vars;
-        vars.functionSelector = LibBytesV06.readBytes4(relayRequest.encodedFunction, 0);
+        vars.functionSelector = LibBytesV06.readBytes4(relayRequest.request.data, 0);
```
shahafn, line 195: I don't think that 'data' is a better name than encodedFunction

contracts/RelayHub.sol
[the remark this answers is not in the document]
shahafn: Why did you change it from calldata to memory?

contracts/forwarder/Eip712Forwarder.sol
```diff
@@ -0,0 +1,145 @@
...
+    //all valid requests must start with this prefix.
+    // request name is arbitrary, but the parameter block must match exactly these parameters
+    string public constant GENERIC_PARAMS = "_ForwardRequest request";
+    string public constant GENERIC_TYPE = "_ForwardRequest(address to,bytes data,address from,uint256 nonce,uint256 gas)";
+    bytes32 public constant GENERIC_TYPEHASH = keccak256(bytes(GENERIC_TYPE));
+
+    mapping(bytes32 => bool) public typeHashes;
+
+    // Nonces of senders, used to prevent replay attacks
+    mapping(address => uint256) private nonces;
```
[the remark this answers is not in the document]
shahafn: Make it public. No reason other than inheritance to have private members.

package.json
```diff
@@ -30,10 +30,10 @@
   },
   "prepublish": "npm run prepare",
   "scripts": {
-    "solc": "../gsn-sponsor/scripts/solc contracts/*.sol 2>&1",
-    "test": "yarn run lint && yarn run truffle-compile && yarn generate && yarn tsc && npm run test-js",
+    "test": "yarn run lint && yarn generate && yarn tsc && npm run test-js",
     "ganache": "ganache-cli --hardfork 'istanbul' --gasLimit 100000000 --defaultBalanceEther 1000 --deterministic --keepAliveTimeout 2147483647",
     "test-js": "run-with-testrpc --hardfork 'istanbul' --port 8544 --gasLimit 10000000 --defaultBalanceEther 1000 --deterministic 'npx truffle --network npmtest test'",
+    "test-bail": "run-with-testrpc --hardfork 'istanbul' --port 8544 --gasLimit 10000000 --defaultBalanceEther 1000 --deterministic 'npx truffle --network npmtest test --bail'",
```
shahafn, line 36: why?

src/common/EIP712/TypedRequestData.ts
```diff
@@ -9,33 +12,46 @@ const EIP712DomainType = [
...
-  GasData: EIP712TypeProperty[]
   RelayData: EIP712TypeProperty[]
+  _ForwardRequest: EIP712TypeProperty[]
+}
+
+export function getDomainSeparator (verifier: Address, chainId: number): any {
+  return {
+    name: 'GSN Relayed Transaction',
+    version: '2',
+    chainId: 1234, // chainId,
```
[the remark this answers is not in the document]
shahafn: I don't understand the hard coded chainId here...

contracts/RelayHub.sol
drortirosh: It isn't a better name. but I wanted to be as "standardized" as possible: we're simply a transport layer for ethereum transaction, and `from`/`to`/`data`/`gas`/`nonce`/`value` are the standard names transaction fields

test/Eip712Forwarder.test.ts
```diff
@@ -0,0 +1,336 @@
...
+  if (m == null) { throw new Error(`type "${type}" doesn't contain "${genericParams}" or "${genericType}"`) }
+
+  const [, name, extraParams, subTypes, subTypes2] = m
+  return {
+    name, extraParams, subTypes, subTypes2
+  }
```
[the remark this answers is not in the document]
drortirosh: what "nonono" ? this is the way you break out a regex. each () in the expression is an array element. position zero is the full match (which we don't care)

contracts/forwarder/Eip712Forwarder.sol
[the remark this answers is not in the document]
drortirosh: I told you that I talked with @shahafn , and we think that at least for some versions, the versionOracle will look them up. e.g. matching a paymaster to a relayhub. this matching is going to be based on the versions, to determine if they are compatible or not - and we want one versioning scheme, so if _some_ versions are going to be checked on-chain, then the format of _all_ version strings should be changed.

contracts/RelayHub.sol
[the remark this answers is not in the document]
drortirosh: Go ahead, help me debug. I removed the public and changed it back to "calldata" and got "unimplemented feature", so I left it this way.

contracts/RelayHub.sol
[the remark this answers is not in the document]
drortirosh: a. I initially wrote "you're right. changing it back" , then I did, got an error - and updated my comment.
b. this is still relevant, so there...

test/RelayServer.webpack.test.ts
```diff
@@ -7,8 +7,11 @@ describe('RelayServer-webpack', () => {
   before('create webpack', function () {
     this.timeout(15000)
     const jsrelayDir = path.join(__dirname, '..', 'jsrelay')
+    console.log('jsrelayDir=', jsrelayDir)
+    // const webpackEntry = require( jsrelayDir+'/webpack.config').entry
+    // assert.ok( fs.existsSync(jsrelayDir+'/'+webpackEntry), 'missing entry file: '+webpackEntry)
     fs.rmdirSync(path.join(jsrelayDir, 'dist'), { recursive: true })
-    childProcess.execSync('npx webpack', { cwd: jsrelayDir })
+    childProcess.execSync('sh -c "find . ../dist ; npx webpack"', { cwd: jsrelayDir, stdio: 'inherit' })
```
shahafn, line 14: Why? it does nothing other than print the current folder and ../dist

contracts/RelayHub.sol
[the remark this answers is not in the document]
drortirosh: a constructor is also a one-time usage. 
I prefer initialization code to be on-chain, then rely on off-chain proper initialization.
That is,  I can write the javascript initialization code - but where should I put it? the RelayClient library?

contracts/test/TestPaymasterPreconfiguredApproval.sol
drortirosh: right (I fixed the code as I wrote the comments, but then had to revert since it didn't compile)

shahafn:
This PR is huge. I don't mind solving several issues in one, but both refactoring and renaming files such that git doesn't show diff other than files removed/added, and having a new more complicated forwarder to review, make it impossible.

test/RelayServer.webpack.test.ts
drortirosh: it was added to troubleshoot a bug in the webpack. (it passed locally, and failed on the server. I don't really remember what fixed it..)

package.json
drortirosh: for testing, it is sometimes better to stop after the first failure - instead of waiting 3-5 minutes for all tests to finish - before seeing the error reason...
(If I could see the stack trace of the already failed error, I wouldn't mind. but truffle (mocha) keeps you in suspense, and show everything at the end..

drortirosh:
> This PR is huge.
I tried to keep separate commits, and most of them (at least in the beginning) were purpose-built (just renaming, just restructuring)
its too bad that github/git compare ignores all sub-commit renames, and show you just the final diff.
You can workaround by walking through the commits.

contracts/RelayHub.sol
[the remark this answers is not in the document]
forshtat: I am sure we can work something out

contracts/RelayHub.sol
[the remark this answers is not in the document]
forshtat: are you saying it did not work like that or what?

contracts/forwarder/Eip712Forwarder.sol
```diff
@@ -0,0 +1,145 @@
...
+            types
+        );
+        bytes32 requestTypehash = keccak256(bytes(requestType));
+        uint len = bytes(subTypes).length;
+        //sanity: avoid redefining our type, e.g.: subType="_ForwardRequest(whatever)_z"
+        require(len == 0 || bytes(subTypes)[len - 1] == ")", "invalid subType");
+        //sanity: parameters should not end parameters block
+        for (uint i = 0; i < bytes(extraParams).length; i++) {
+            require(bytes(extraParams)[i] != ")", "invalid extraParams");
+        }
```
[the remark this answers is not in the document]
forshtat: I am not sure this is equivalent, here we are talking about a malicious JS code trying to trick user into signing something, not an executable code. So, the severity is lover.

contracts/interfaces/GsnTypes.sol
```diff
@@ -0,0 +1,29 @@
...
+    struct RelayData {
+        uint256 gasPrice;
+        uint256 pctRelayFee;
+        uint256 baseRelayFee;
+        address relayWorker;
+        address paymaster;
+    }
+
+    struct ExtraData {
+        address forwarder;
```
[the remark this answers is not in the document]
forshtat: don't. just sign.

contracts/utils/GsnEip712Library.sol
```diff
@@ -0,0 +1,124 @@
...
+
+    bytes32 public constant RELAYDATA_TYPEHASH = keccak256(RELAYDATA_TYPE);
+    bytes32 public constant RELAY_REQUEST_TYPEHASH = keccak256(RELAY_REQUEST_TYPE);
+
+    //must call this method exactly once to register the GSN type.
+    // (note that its a public method: anyone can register this GSN version)
+    function registerRequestType(Eip712Forwarder forwarder) internal {
+        forwarder.registerRequestType(
+            RELAY_REQUEST_NAME, RELAY_REQUEST_PARAMS, string(RELAYDATA_TYPE), "" );
+        require(forwarder.isRegisteredTypehash(RELAY_REQUEST_TYPEHASH), "Fatal: registration failed");
```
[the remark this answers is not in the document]
forshtat: Yes. The downside is more solidity code.

contracts/utils/GsnEip712Library.sol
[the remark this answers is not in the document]
forshtat: yes. relay hub is pure magic, no linter can comprehend it's perfection 😁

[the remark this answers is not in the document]
forshtat: why do you keep saying fixed if you did not change the string?

test/Utils.test.ts
forshtat: I was hinting on not committing  entire paragraphs of comments in general

package.json
forshtat: oh, cool. we should run this in CI as well.

test/RelayServer.webpack.test.ts
forshtat: I suggest removing things you cannot remember being useful.

forshtat:
> > This PR is huge.
> > I tried to keep separate commits, and most of them (at least in the beginning) were purpose-built (just renaming, just restructuring)
> > its too bad that github/git compare ignores all sub-commit renames, and show you just the final diff.
> > You can workaround by walking through the commits.

You realize this PR is 38 commits long? 😀 You said my PRs were big when there were like 5 commits

contracts/forwarder/Eip712Forwarder.sol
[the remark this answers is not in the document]
drortirosh: the only reason for using a structure, is make sure the user sees what we meant him to see,
e.g. if someone can swap the "from" and "to" fields, the user might think the tx means different thing.
That's why its important there is no way to abuse the structure.
Since there is no "trusted authority", we had to check it on chain.
we can simplify by far the check if instead of struct we have plain fields.

drortirosh:
> You realize this PR is 38 commits long? 😀 You said my PRs were big when there were like 5 commits

you're right. just used your approach, and stopped rebasing all the time... and its also not a PR to be squashed either...

contracts/forwarder/Eip712Forwarder.sol
```diff
@@ -64,47 +62,22 @@ contract Eip712Forwarder is IForwarder {
...
-            require(bytes(extraParams)[i] != ")", "invalid extraParams");
+    function registerRequestType(string calldata typeName, string calldata typeSuffix) external override {
+
+        for (uint i = 0; i < bytes(typeName).length; i++) {
+            bytes1 c = bytes(typeName)[i];
+            require(c != "(" && c != ")", "invalid typename");
         }
+
+        bytes1 separator = ")";
+        if (bytes(typeSuffix).length > 0) separator = ",";
```
forshtat, line 73: Let's ban empty type suffix, and pre-register pure forwarder type in constructor. I don't like this conditional code in 'register' function.

src/common/EIP712/RelayRequest.ts
```diff
@@ -8,20 +8,7 @@ export default interface RelayRequest {
...
-    relayData: {
-      gasPrice: relayRequest.relayData.gasPrice,
-      pctRelayFee: relayRequest.relayData.pctRelayFee,
-      baseRelayFee: relayRequest.relayData.baseRelayFee,
-      paymaster: relayRequest.relayData.paymaster,
-      forwarder: relayRequest.relayData.forwarder,
-      relayWorker: relayRequest.relayData.relayWorker
-    }
+    request: { ... relayRequest.request },
+    relayData: { ... relayRequest.relayData }
```
forshtat, line 12: I honestly prefer the verbose, type-aware way...

src/common/EIP712/TypedRequestData.ts
```diff
@@ -57,7 +56,7 @@ export default class TypedRequestData implements EIP712TypedData {
   readonly types: Types
   readonly domain: EIP712Domain
   readonly primaryType: string
-  readonly message: RelayRequest
+  readonly message: any
```
forshtat, line 59: mmmhhhh....

src/common/EIP712/TypedRequestData.ts
```diff
@@ -67,17 +66,19 @@ export default class TypedRequestData implements EIP712TypedData {
...
+    }
   }
 }
 
 export const GsnRequestType = {
   typeName: 'RelayRequest',
-  extraParams: 'RelayData relayData',
-  subTypes: 'RelayData(uint256 gasPrice,uint256 pctRelayFee,uint256 baseRelayFee,address relayWorker,address paymaster)',
-  subTypes2: ''
+  typeSuffix: 'RelayData relayData)RelayData(uint256 gasPrice,uint256 pctRelayFee,uint256 baseRelayFee,address relayWorker,address paymaster)',
```
forshtat: Why could it not stay as 'extraParams' and 'subTypes'? You changed the way it signs, did you have to change the API into passing a half-truncated string?

src/common/Utils.ts
```diff
@@ -36,13 +36,14 @@ export function event2topic (contract: any, names: any): any {
     .map(abi.encodeEventSignature)
 }
 
+//caller provider's eth_signTypedData
```
forshtat, line 39: I don't know why, but your comments actually make me understand less. Could you please stop putting comments in code?

test/RelayHubGasCalculations.test.ts
```diff
@@ -16,14 +16,14 @@ import {
...
-const Forwarder = artifacts.require('Forwarder')
+const Eip712Forwarder = artifacts.require('Eip712Forwarder')
 const StakeManager = artifacts.require('StakeManager')
 const Penalizer = artifacts.require('Penalizer')
 const TestRecipient = artifacts.require('TestRecipient')
 const TestPaymasterVariableGasLimits = artifacts.require('TestPaymasterVariableGasLimits')
 const TestPaymasterConfigurableMisbehavior = artifacts.require('TestPaymasterConfigurableMisbehavior')
 
-contract('RelayHub gas calculations', function ([_, relayOwner, relayWorker, relayManager, senderAddress, other]) {
+contract.skip('RelayHub gas calculations', function ([_, relayOwner, relayWorker, relayManager, senderAddress, other]) {
```
forshtat, line 26: Please unskip gas calculations before this can go in.

contracts/RelayHub.sol
forshtat: I agree with @drortirosh and resolve the comment.

test/RelayServer.webpack.test.ts
```diff
@@ -7,8 +7,11 @@ describe('RelayServer-webpack', () => {
   before('create webpack', function () {
     this.timeout(15000)
     const jsrelayDir = path.join(__dirname, '..', 'jsrelay')
+    console.log('jsrelayDir=', jsrelayDir)
+    // const webpackEntry = require( jsrelayDir+'/webpack.config').entry
+    // assert.ok( fs.existsSync(jsrelayDir+'/'+webpackEntry), 'missing entry file: '+webpackEntry)
     fs.rmdirSync(path.join(jsrelayDir, 'dist'), { recursive: true })
-    childProcess.execSync('npx webpack', { cwd: jsrelayDir })
+    childProcess.execSync('sh -c "npx webpack"', { cwd: jsrelayDir, stdio: 'inherit' })
```
forshtat, line 14: What is even going on here? How come this file needs changes at all?

contracts/utils/GsnEip712Library.sol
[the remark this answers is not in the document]
forshtat: Forget it

test/RelayServer.webpack.test.ts
drortirosh: removed (was historical reason: change in dpeloy script caused this to fail, so I added check code, including file listings.
then found a better way to debug: in circleci, you can launch with ssh: after a [failed] launch, you can ssh into the build machine to look around - which is much better than repeatedly pushing build script with debugging stuff...

src/common/EIP712/TypedRequestData.ts
drortirosh: if I pass 2 separate items, then I need to make sure they indeed each mean what they meant to mean - but that's exactly the tests I wanted to remove.

src/common/EIP712/TypedRequestData.ts
drortirosh: yes - it ISN'T a relayRequest. it is a member to match the format for signing - and relayRequest is of different structure.

src/common/EIP712/RelayRequest.ts
drortirosh: it caused so much hassle to fix on each change. I honestly want a "deep-clone" here. I settled on using one-level clone.

src/common/EIP712/TypedRequestData.ts
forshtat: Can they mean something else with a flattened structure? Honest question.

src/common/EIP712/TypedRequestData.ts
forshtat: Could still use a type declaration, just saying...

src/common/EIP712/RelayRequest.ts
forshtat: This PR signifies an API freeze, so... Kinda...

contracts/forwarder/Eip712Forwarder.sol
drortirosh: I don't really understand why you want to ban no-added-args.
The net result is that there is  exactly one allowed message name without extra parameters - and whatever name you like if you add a single field.
Why is that make sense?

forshtat: what does not make sense is code like ```if (unnecessaryCheck) { doUnnecessaryStuff() }``` in solidity. 
No extra parameters - no rename of a structure for you.

Want a custom name? Add field 'hello', set value to 'world'. I don't want to care about this use-case

contracts/forwarder/Eip712Forwarder.sol
```diff
@@ -0,0 +1,123 @@
...
+        _verifyNonce(req);
+        _verifySig(req, domainSeparator, requestTypeHash, suffixData, sig);
+        _updateNonce(req);
+
+        // solhint-disable-next-line avoid-low-level-calls
+        return req.to.call{gas : req.gas}(abi.encodePacked(req.data, req.from));
+    }
+
+    function _verifyNonce(ForwardRequest memory req) internal view {
+        require(nonces[req.from] == req.nonce, "nonce mismatch");
```
wighawag, line 58: Not planning to support different nonce strategy (provided by signer, like https://github.com/wighawag/eip-2585/blob/c9769ad0b4260b4f08f981527bb1162ecce570d3/contracts/src/EIP712Forwarder.sol#L50) ?

contracts/forwarder/Eip712Forwarder.sol
```diff
@@ -0,0 +1,123 @@
...
+
+import "@openzeppelin/contracts/cryptography/ECDSA.sol";
+import "./IForwarder.sol";
+// TODO now: move getError to the forwarder folder
+import "../utils/GsnUtils.sol";
+
+contract Eip712Forwarder is IForwarder {
+    using ECDSA for bytes32;
+
+    string public constant GENERIC_PARAMS = "address to,bytes data,address from,uint256 nonce,uint256 gas";
```
wighawag, line 13: I would remove gas from the forwarder. This can be handled by forwarder extension.
This is important because EVM is not great for gas handling right now

contracts/forwarder/Eip712Forwarder.sol
```diff
@@ -0,0 +1,123 @@
...
+    public
+    override
+    returns (bool, bytes memory) {
+
+        _verifyNonce(req);
+        _verifySig(req, domainSeparator, requestTypeHash, suffixData, sig);
+        _updateNonce(req);
+
+        // solhint-disable-next-line avoid-low-level-calls
+        return req.to.call{gas : req.gas}(abi.encodePacked(req.data, req.from));
```
wighawag, line 54: no guarantee that `req.gas` will be passed in here.
Better to handle that in the Forwarder extension. Hence why my comment above about removing the gas parameter from the base forwarder. This is also how I handled in EIP-2585 : https://github.com/wighawag/eip-2585/blob/c9769ad0b4260b4f08f981527bb1162ecce570d3/contracts/src/EIP712Forwarder.sol#L226

contracts/forwarder/Eip712Forwarder.sol
```diff
@@ -0,0 +1,123 @@
...
+    internal
+    view
+    {
+
+        require(typeHashes[requestTypeHash], "invalid request typehash");
+        bytes32 digest = keccak256(abi.encodePacked(
+                "\x19\x01", domainSeparator,
+                keccak256(_getEncoded(req, requestTypeHash, suffixData))
+            ));
+        require(digest.recover(sig) == req.from, "signature mismatch");
```
wighawag, line 98: No plan to support smart contract wallet with EIP-1271 and EIP-1654 ?
I think it would be great for wallet so they do not need special handling for each smart contract wallet metatx support

contracts/forwarder/Eip712Forwarder.sol
```diff
@@ -0,0 +1,123 @@
...
+
+    function getNonce(address from)
+    public view override
+    returns (uint256) {
+        return nonces[from];
+    }
+
+    function verify(
+        ForwardRequest memory req,
+        bytes32 domainSeparator,
```
wighawag, line 28: why is the domain separator not hardcoded in the forwarder?
Allowing Forwarder extension to change it does not seem like a good idea as this might confuse user

contracts/forwarder/Eip712Forwarder.sol
drortirosh: If we move the gas into the extension, then the forwarder will not be able to apply it to the call, since (by definition) the generic forwarder knows only the generic fields.
The relayer calling the forwarder must have a way to ensure an "upper bound" on the gas limit of the call, and the "execute" is the place to handle that gas limits.
Its true that EVM's gas handling is problematic, and even more when handled in nested calls. That's why it has to be the "leaf" call to enforce the gas limit (that is, it is very hard to estimate the value to apply when calling "forwarder.execute" itself.

contracts/forwarder/Eip712Forwarder.sol
drortirosh: The purpose of a forwarder is to forward requests from an EOA. Why would you use it if the sender is another contract?

contracts/forwarder/Eip712Forwarder.sol
drortirosh: Depends what we define as "domain". 
I was thinking of "Domain" as the relaying service in use (in our case, its "GSN"). It has nothing to do with the client's signature on the request.

contracts/forwarder/Eip712Forwarder.sol
drortirosh: Its quite possible to add such nonce policy.
We're leaving it to a separate pull request, as this one is already too big (and also it doesn't require an API change)
The nice thing adding batch/nonce can be done not only without solidity API change, but even without client change: the standard logic of latest-nonce-plus-1 is good even if the current batch was changed.

contracts/forwarder/Eip712Forwarder.sol
wighawag: ok, I see

contracts/forwarder/Eip712Forwarder.sol
wighawag: to simplify wallet integration. they would not need to have special case for smart contract account (which come in different variation)

contracts/forwarder/Eip712Forwarder.sol
wighawag: The current way you apply gas do not protect the signer so I do not see the point of having it there.
Plus you only need to protect the signer if it is the one who pays the gas. hence why it should be outside of the forwarder, that have no need to bother about gas repayment

Gas limit can be easily handled in the forwarder caller as I did in my EIP-1776 demo : https://github.com/wighawag/eip-2585/blob/c9769ad0b4260b4f08f981527bb1162ecce570d3/contracts/src/EIP1776ForwarderWrapper.sol#L102

if better gas handling appears, we can keep using the same forwarder and extension can use better handling

contracts/forwarder/Eip712Forwarder.sol
drortirosh: Relayed transaction has its overhead. I don't see a reason to go through it unless it has some benefit.

wighawag: the benefit is that smart contract wallet can use GSN or any other native meta tx capable relayer
I agree that there is an overhead (checking if sender is a contract). but it might be worth it. 
The point is that if the forwarder do not support it, then we cannot add it later, without having a new forwarder

contracts/forwarder/Eip712Forwarder.sol
```diff
@@ -0,0 +1,131 @@
...
+    public
+    override
+    returns (bool, bytes memory) {
+
+        _verifyNonce(req);
+        _verifySig(req, domainSeparator, requestTypeHash, suffixData, sig);
+        _updateNonce(req);
+
+        // solhint-disable-next-line avoid-low-level-calls
+        return req.to.call{gas : req.gas}(abi.encodePacked(req.data, req.from));
```
wighawag, line 60: msg.value should be passed in, allowing metatx to also be able to paay with ETH

contracts/forwarder/Eip712Forwarder.sol
```diff
@@ -0,0 +1,131 @@
+// SPDX-License-Identifier:MIT
```
wighawag, line 1: General comment :
I don't see the benefit of braking down the forwarder in multiple inheritance contracts

We cannot expect receiver to choose which forwarder they support. The forwarder should be supporting features that we deemed necessary.

In my opinion, these are
- batch calls
- msg.value forwarding
- flexible nonce protection

And I am on the fence with :
- smart contract wallet support 

If the overhead is minimal (checking if sender is a smart contract)
I would go with it as it enable all smart contract wallet that support EIP-1271 and EIP-1654, to be supported by wallet that support GSN or other native metatx  enabledrelayer service

contracts/forwarder/Eip712Forwarder.sol
drortirosh: I'm not talking about the overhead of checking if the sender is a contract.
I'm asking why should a contract call `forwarder.execute( ForwardRequest(target,func, ...) )`
instead of `target.call(func)` ?

wighawag: that's not how it would work
see : https://github.com/wighawag/gsn-playground/blob/2a4a4652ecd72efd233c2442a26db97adea67da9/contracts/src/MetaTransaction/ForwarderRegistry.sol#L77-L88

the overhead for non-contract is really just checking if it is a contract

wighawag: if you talk about the overhead for smart contract wallet, this is the relay overhead for the forwarder  (+ eip1271/eip1654 call) vs the relay overhead used by the smart contract wallet

PR #395, opened 2020-06-03
paymaster-server integration, relay workers rebalancing by manager hub balance, balance to owner fixes, registration fixes

shahafn:
GSN-7
GSN-70
GSN-71
GSN-72
GSN-73


jsrelay/docker-compose.yml
```diff
@@ -20,12 +20,12 @@ services:
...
-      Url: https://${HOST}
-      Port: 80
-      EthereumNodeUrl: ${NODE_URL}
-      RelayHubAddress: ${RELAY_HUB}
-      GasPricePercent: ${GAS_PRICE_PERCENT}
-      BaseFee: ${BASE_FEE}
-      PercentFee: ${PERCENT_FEE}
+      url: https://${HOST}
+      port: 80
+      ethereumNodeUrl: ${NODE_URL}
```
drortirosh, line 25: the whole point of config file is use a config file. why still pass everything as cmdline args?
maybe url, workdir still "deserve"  cmdline, but the rest aren't.

src/relayserver/RelayServer.ts
```diff
@@ -594,50 +614,97 @@ export class RelayServer extends EventEmitter {
...
+      console.log(`Sending workers' eth balance ${workerBalance.toString()} to owner`)
+      receipts.push((await this._sendTransaction({
+        signerIndex: 1,
+        destination: this.owner as string,
+        gasLimit: gasLimit.toString(),
+        gasPrice,
+        value: toHex(workerBalance.sub(txCost))
+      })).receipt)
+    } else {
+      console.log(`balance too low: ${managerBalance.toString()}, tx cost: ${gasLimit * parseInt(gasPrice)}`)
```
drortirosh, line 703: wrong error message: its worker, not manager.

src/relayserver/runServer.js
```diff
@@ -22,36 +22,35 @@ const envDefaults = Object.entries(process.env)
...
-const port = argv.Port || 8090
-const relayHubAddress = getRelayHubAddress(argv.RelayHubAddress) || error('missing --RelayHubAddress')
-// const defaultGasPrice = argv.DefaultGasPrice || 1e9 // 1 Gwei
-const gasPricePercent = argv.GasPricePercent || 10
-// const registrationBlockRate = argv.RegistrationBlockRate || 6000 - 200
-const ethereumNodeUrl = argv.EthereumNodeUrl || 'http://localhost:8545'
-const workdir = argv.Workdir || error('missing --Workdir')
-const devMode = argv.DevMode || false
-const Debug = argv.Debug || false
+const config = argv.config ? require(argv.config) : error('missing --config')
```
drortirosh, line 43: I thought a config file was optional. now I see its mandatory..

test/server-config.json
```diff
@@ -0,0 +1,12 @@
...
+  "baseRelayFee": 70,
+  "pctRelayFee": 0,
+  "url": "http://localhost:8090",
+  "port":  8090,
+  "relayHubAddress": "",
+  "gasPricePercent": 10,
+  "ethereumNodeUrl": "http://localhost:8545",
+  "workdir": "",
+  "devMode": "true",
+  "Debug": "false"
```
drortirosh, line 11: lowercase debug

src/relayserver/runServer.js
```diff
@@ -22,36 +22,35 @@ const envDefaults = Object.entries(process.env)
...
+      'relayHubAddress',
+      'defaultGasPrice',
+      'gasPricePercent',
+      'ethereumNodeUrl',
+      'workdir'
     ],
-  boolean: ['DevMode', 'Debug'],
+  boolean: ['devMode', 'debug'],
   alias: {},
   default: envDefaults
```
drortirosh, line 37: current **envDefaults** will not work: it requires env var to be CamelCase (starting with capital and then small letter). 
This won't work with new names, that some are all lowercase..
simplest fix is accept any env.var starting with lowercase (all standard env.vars are all caps)
e.g.: `/^[a-z]\w*$/`

src/relayserver/RelayServer.ts
```diff
@@ -429,31 +435,40 @@ export class RelayServer extends EventEmitter {
...
+        const message = `== replenishWorker: can't replenish: mgr balance too low ${balance.div(toBN(1e18)).toString()} refill=${refill.div(
+          toBN(1e18)).toString()}`
+        this.emit('fundingNeeded', message)
+        console.log(message)
       }
     }
   }
 
-  async _worker (blockHeader: BlockHeader): Promise<TransactionReceipt | void> {
+  async _worker (blockHeader: BlockHeader): Promise<TransactionReceipt[]> {
```
drortirosh, line 471: I saw tests that use `_worker()`'s return value - but they assume its a single receipt.
how can they work?

src/relayserver/RelayServer.ts
```diff
@@ -96,7 +96,8 @@ export interface RelayServerParams {
   readonly web3provider: provider
   readonly keyManager: KeyManager
   readonly contractInteractor: ContractInteractor
-  readonly hubAddress: PrefixedHexString
+  readonly hubAddress: Address
+  readonly trustedPaymasters?: Address[]
```
drortirosh, line 100: no new tests for trustedPaymasters. separate issue?

src/relayserver/RelayServer.ts
```diff
@@ -429,31 +435,40 @@ export class RelayServer extends EventEmitter {
...
           gasLimit: mintxgascost.toString()
         })
-        await this.getManagerBalance()
       } else {
-        console.log(
-          `== replenishWorker: can't replenish: mgr balance too low ${balance.div(toBN(1e18)).toString()} refill=${refill.div(
-            toBN(1e18)).toString()}`)
+        const message = `== replenishWorker: can't replenish: mgr balance too low ${balance.div(toBN(1e18)).toString()} refill=${refill.div(
+          toBN(1e18)).toString()}`
+        this.emit('fundingNeeded', message)
```
forshtat, line 465: To the point of our conversation earlier today: I believe here is an example of an 'exception' you might have caught and know exactly what you want to say to the higher-tier process - you need more money. It is kind of an exception, but a well anticipated one - money go fast. This can be emitted as an event, but is there another way?
What if what you did was calling a purpose-built callback function?

```this.callbacks.onFundingNeeded()```? You do not interrupt the execution flow, do not lose the stack trace or step in functionality, and you know for sure that this will be handled. 

HttpServer's behaviour on 'fundingNeeded' can easily be tested or mocked, and the API for the method can be easily declared in TypeScript, like:
```function onFundingNeeded(message: string, address: Address )```

File: src/relayserver/RelayServer.ts [hunk 9]
```diff
@@ -478,23 +493,23 @@ export class RelayServer extends EventEmitter {
...
     const decodedLogs = abiDecoder.decodeLogs(logs).map(this._parseEvent)
-    let receipt
+    let receipts: TransactionReceipt[] = []
     // TODO: what about 'penalize' events? should send balance to owner, I assume
     // TODO TODO TODO 'StakeAdded' is not the event you want to cat upon if there was no 'HubAuthorized' event
     for (const dlog of decodedLogs) {
       switch (dlog.name) {
         case 'HubAuthorized':
-          receipt = await this._handleHubAuthorizedEvent(dlog)
+          receipts = receipts.concat(await this._handleHubAuthorizedEvent(dlog))
```
forshtat, line 502: Please do not use 'await' in this manner. Please create variables for them, it greatly improves readability (at least for me)

src/relayserver/RelayServer.ts
```diff
@@ -594,50 +614,97 @@ export class RelayServer extends EventEmitter {
...
+      filter: { relayManager: this.managerAddress }
     })
-    debug(`Relay ${this.managerAddress} registered on hub ${this.relayHubContract?.address}. `)
-
+    if (relayRegisteredEvents.find(
+      (e: any) =>
+        e.returnValues.relayManager.toLowerCase() === this.managerAddress.toLowerCase() &&
+        e.returnValues.baseRelayFee.toString() === this.baseRelayFee.toString() &&
+        e.returnValues.pctRelayFee.toString() === this.pctRelayFee.toString() &&
+        e.returnValues.relayUrl.toString() === this.url.toString()) == null) {
```
forshtat, line 632: maybe create something (function, lambda, idk) for this, it is a long ```find```

contracts/RelayHub.sol
```diff
@@ -55,14 +55,14 @@ contract RelayHub is IRelayHub {
...
+    mapping(address => address) public workerToManager;
 
     // maps relay managers to the number of their workers
-    mapping(address => uint256) private workerCount;
+    mapping(address => uint256) public workerCount;
 
     uint256 constant public MAX_WORKER_COUNT = 10;
 
-    mapping(address => uint256) private balances;
+    mapping(address => uint256) public balances;
```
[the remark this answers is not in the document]
forshtat: We can remove balanceOf now that we use solidity 0.6 and can use mapping to implement an interface

src/relayserver/runServer.js
forshtat: What is envDefaults?

src/relayserver/RelayServer.ts
forshtat: In my view, event-driven approach is like if postmen threw all mail of a street to a single trash can, so that anyone expecting mail would be able to sort through it to find his and ignore the rest of mail and trash. And in case there is only a single recipient for messages, it becomes even less clear why not use a post box?
Sorry for bad metaphores

src/relayserver/RelayServer.ts
shahafn: There is a test

src/relayserver/RelayServer.ts
shahafn: fixed

contracts/RelayHub.sol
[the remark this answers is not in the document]
shahafn: I don't care, we shouldn't have private fields for other reasons than inheritance.

src/relayserver/RelayServer.ts
shahafn: I changed it in this PR. No tests assume a single receipt.

jsrelay/.env
```diff
@@ -6,9 +6,10 @@ RELAY_HUB=0x79D1160DFDc24326a21fDC280D8FEc341c85D9Af
 #can't use localhost: must specify device IP
 #NODE_URL=http://192.168.1.241:8545
 
-GAS_PRICE_PERCENT=70
-BASE_FEE=
-PERCENT_FEE=
+#GAS_PRICE_PERCENT=70
+SERVER_CONFIG_FILE=server-config.json
+#BASE_FEE=
+#PERCENT_FEE=
```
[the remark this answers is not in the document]
shahafn: I talked with @drortirosh about that, and he believes that having multiple ways to configure it might be beneficial for using other systems than docker/docker-compose, and I don't mind it either now.

src/relayserver/RelayServer.ts
shahafn: "You do not interrupt the execution flow, do not lose the stack trace or step in functionality, and you know for sure that this will be handled." - I don't understand what you mean here, but this is certainly not the place to do it.

test/server-config.json
shahafn: fixed

src/relayserver/runServer.js
shahafn: fixed

src/relayserver/RelayServer.ts
shahafn:
```
      (e: any) =>
        e.returnValues.relayManager.toLowerCase() === this.managerAddress.toLowerCase() &&
        e.returnValues.baseRelayFee.toString() === this.baseRelayFee.toString() &&
        e.returnValues.pctRelayFee.toString() === this.pctRelayFee.toString() &&
        e.returnValues.relayUrl.toString() === this.url.toString()
```
Isn't it obvious that it compares all of the event entries to see if it was already registered with the same params? It seems pretty obvious to me

jsrelay/docker-compose.yml
shahafn: This should be fixed in a PR after also testing docker imo

src/relayserver/RelayServer.ts
drortirosh: where is it? you added a `trustedPaymasters` parameter to RelayServer, but the string doesn't appear in any test.

PR #396, opened 2020-06-03
Remove canrelay strings

forshtat:


contracts/interfaces/IRelayHub.sol
```diff
@@ -114,19 +112,19 @@ interface IRelayHub {
...
+    ///  - the Paymaster must have enough balance to pay the Relay Worker for the scenario when all gas is spent
     ///
     /// If all conditions are met, the call will be relayed and the recipient charged.
     ///
     /// Arguments:
-    /// @param relayRequest - all details of the requested relayWorker call
-    /// @param signature - client's signature over all previous params, plus the relayWorker and RelayHub addresses
+    /// @param relayRequest - all details of the requested relayed call
+    /// @param signature - client's EIP-712 signature over the relayRequest struct
     /// @param approvalData: dapp-specific data forwarded to acceptRelayedCall.
```
drortirosh, line 126: nitpick: its Paymaster-specific data, not dapp-specific

forshtat: this string was not even touched by this commit

drortirosh: exactly... you changed all the documents around it - but it not (still, its a nitpick only: I'd accept it if that was the only issue: this commit is based on another commit to be fixed

PR #398, opened 2020-06-04
Fix prepack-postpack-generate

drortirosh:
the `prepublish` scripts runs also after each "yarn install", which is excessive.
instead, we use now `prepack`/`postpack`, which run only on "pack" and "publish", as expected.
also: `generate` is now a script that now does all preparation (compiles contracts, extract ABIs and cli-required contracts, generate types)

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,7 +28,6 @@
   "bin": {
     "gsn": "dist/src/cli/commands/gsn.js"
   },
-  "prepublish": "npm run prepare",
   "scripts": {
     "test": "yarn run lint && yarn generate && yarn tsc && npm run test-js",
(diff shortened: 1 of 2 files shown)


scripts/generate
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash -e
+
+rm -rf ./build/contracts
+truffle compile
+rm -rf types/truffle-contracts/
+typechain --target trufflehotfix './build/contracts/**/*.json'
+rm -rf src/common/interfaces/
+yarn extract_abi
```
forshtat, line 8: do we still need 'extract_abi'?

forshtat:
Is this abandoned? It did not pass tests 4 days ago.

PR #402, opened 2020-06-11
GSN-78: Add 'trustedForwarder' to BasePaymaster

forshtat:


shahafn:
TODO move back tests from gsn-paymasters that are relevant to base paymsters!

PR #405, opened 2020-06-14
Forwarder supports "value"

drortirosh:
forwarder now has a "value" parameter, which is the amount of ether to send with the transaction.
The {{verifyAndCall}} function is now payable, to be able to receive this value.
Note that it is possible to pass the value into the forwarder in a call just before verifyAndCall (still, under the same transaction)

The {{verifyAndCall}} would fail the request if it is given a **value** parameter, but doesn't have enough value to fulfill it.


forshtat:
Closed in favour of https://github.com/opengsn/gsn/pull/410

PR #407, opened 2020-06-16
Simplified forwarder

drortirosh:
Instead of inner type, put the generic fields directly into the main struct.
This way, extending the type is much simpler - and registerType also gets simpler., as it no longer needs to verify the type - no way create valid  type which override existing fields.

### DIFF
--- a/contracts/forwarder/Eip712Forwarder.sol
+++ b/contracts/forwarder/Eip712Forwarder.sol
@@ -12,21 +12,15 @@ contract Eip712Forwarder is IForwarder {
 
     //all valid requests must start with this prefix.
     // request name is arbitrary, but the parameter block must match exactly these parameters
-    string public constant GENERIC_PARAMS = "_ForwardRequest request";
-    string public constant GENERIC_TYPE = "_ForwardRequest(address to,bytes data,address from,uint256 nonce,uint256 gas)";
-    bytes32 public constant GENERIC_TYPEHASH = keccak256(bytes(GENERIC_TYPE));
+    string public constant GENERIC_PARAMS = "address to,bytes data,address from,uint256 nonce,uint256 gas";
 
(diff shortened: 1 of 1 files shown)


contracts/forwarder/Eip712Forwarder.sol
```diff
@@ -12,23 +12,17 @@ contract Eip712Forwarder is IForwarder {
 
     //all valid requests must start with this prefix.
     // request name is arbitrary, but the parameter block must match exactly these parameters
-    string public constant GENERIC_PARAMS = "_ForwardRequest request";
-    string public constant GENERIC_TYPE = "_ForwardRequest(address to,bytes data,address from,uint256 nonce,uint256 gas)";
-    bytes32 public constant GENERIC_TYPEHASH = keccak256(bytes(GENERIC_TYPE));
+    string public constant GENERIC_PARAMS = "address to,bytes data,address from,uint256 nonce,uint256 gas";
```
forshtat, line 15: I swear I have no idea why, but these parameters come in a different order in my head 😂
For me it is 'from, to, gas, value, nonce, data'? I know there is no standard or anything...
Also, if you see it in RLP hex having variable size parameter last is more... clear. Or maybe I am just slowly going crazy.

contracts/forwarder/Eip712Forwarder.sol
```diff
@@ -67,33 +61,14 @@ contract Eip712Forwarder is IForwarder {
      * @param subTypes - subtypes used by the extraParams
      * @param subTypes2 - more subtypes, if sorted after _ForwardRequest (e.g. if type starts with lowercase)
```
forshtat, line 62: I can see the old comments

drortirosh:
alreayd merged into generic-forwarder

PR #408, opened 2020-06-17
Forwarder alex rework

forshtat:
1. Change the way GsnEip712Library handles calls to externall addresses
to minimize the risk of revert. Cannot rely on try/catch, cannot rely on
return values being decodable.

2. Make ARC external again. Also 'innerRelayCall'. Generaly I believe it
is preferable to keep methods external to avoid copying data to memory
etc.

3. Remove 'registerRequestType' from the RelayHub. Can be done off-chain
with a oneliner.

4. Add 'relayedCallReturnValue' field. Not used yet, but should be
emitted.

5. Made forwarder part of RelayData. NOTE: did not change signature yet.

6. Removed 'domain separator' from RelayRequest. It is cheaper to
calculate it on each request, and it is a cleaner API.

7. Renamed 'callForwarderVerify' and 'callForwarderVerifyAndCall' to
'verify' and 'execute'


contracts/forwarder/Eip712Forwarder.sol
```diff
@@ -36,22 +34,28 @@ contract Eip712Forwarder is IForwarder {
...
+    returns (bool, string memory) {
         _verifyNonce(req);
         _verifySig(req, domainSeparator, requestTypeHash, suffixData, sig);
         _updateNonce(req);
 
         // solhint-disable-next-line avoid-low-level-calls
-        return req.to.call{gas: req.gas}(abi.encodePacked(req.data, req.from));
+        (bool executeSuccess, bytes memory ret) = req.to.call{gas: req.gas}(abi.encodePacked(req.data, req.from));
+        if (!executeSuccess){
+            return (false, GsnUtils.getError(ret));
```
drortirosh, line 54: the problem with getError is that it adds dependency on LibBytes, which I didn't want to have in the forwarder, and leave it to its caller. why not redefine `ret` as `revertBytes`, and let the caller handle it.

PR #409, opened 2020-06-21
Server fixes

shahafn:
GSN-106
GSN-107
GSN-110


src/relayserver/RelayServer.ts
```diff
@@ -16,7 +16,7 @@ import { KeyManager } from './KeyManager'
 import ContractInteractor from '../relayclient/ContractInteractor'
 import PingResponse from '../common/PingResponse'
 import { Address, IntString } from '../relayclient/types/Aliases'
-import GsnTransactionDetails from '../relayclient/types/GsnTransactionDetails'
+import TmpRelayTransactionJsonRequest from '../relayclient/types/TmpRelayTransactionJsonRequest'
```
forshtat, line 19: a reminder that this class was  supposed to be temporary until we unify client and server request types

contracts/StakeManager.sol
```diff
@@ -54,16 +54,19 @@ contract StakeManager is IStakeManager {
...
-        StakeInfo storage info = stakes[relayManager];
-        require(info.owner == msg.sender, "not owner");
+    modifier ownerOrManagerOnly(address relayManager) {
+        StakeInfo storage infoByManager = stakes[relayManager];
+        StakeInfo storage infoBySender = stakes[msg.sender];
+        require(infoByManager.owner == msg.sender || infoBySender.owner != address(0), "not owner or manager");
+        _;
+    }
+
+    function authorizeHub(address relayManager, address relayHub) external override ownerOrManagerOnly(relayManager) {
```
forshtat, line 64: This has a security implication that manager account of any hub can now authorise any other hub (therefore can withdraw the stake through the penalisation).
Meaning, there is no longer semantical difference between 'owner' and 'manager'.
Why would manager need a way to authorize a new hub?

src/relayserver/RelayServer.ts
```diff
@@ -65,21 +65,7 @@ function spam (...args: any): void {
...
-  signature: PrefixedHexString
-  senderNonce: IntString
-  relayMaxNonce: IntString
-  baseRelayFee: IntString
-  pctRelayFee: IntString
-  relayHubAddress: Address
-  paymaster: Address
-  forwarder: Address
-}
+export type CreateTransactionDetails = TmpRelayTransactionJsonRequest
```
forshtat, line 68: exactly

src/relayserver/RelayServer.ts
```diff
@@ -607,14 +612,10 @@ export class RelayServer extends EventEmitter {
...
     }
-    const workersAddedEvents = await this.relayHubContract?.contract.getPastEvents('RelayWorkersAdded', {
-      fromBlock: 1,
-      filter: { relayManager: this.managerAddress }
-    })
 
     // add worker only if not already added
-    if (workersAddedEvents.find((e: any) => e.returnValues.newRelayWorkers
-      .map((a: string) => a.toLowerCase()).includes(this.getAddress(1).toLowerCase())) == null) {
+    const workersAdded = await this._areWorkersAdded()
```
forshtat, line 617: what's going on here? why _isRegistered and _areWorkersAdded need to be async?
is there a risk of checking and mining race condition then?

src/relayserver/RelayServer.ts
shahafn: Yes it should be removed, I didn't think that I should refactor it in this PR

contracts/StakeManager.sol
shahafn: If we don't have it, no way to auto upgrade server to support new hub. Each upgrade now requires owner manual tx authorization

src/relayserver/RelayServer.ts
shahafn: This should be removed when removing tmpRelayTransactionJsonRequest

src/relayserver/RelayServer.ts
shahafn: I just separated it to a function, they are async since `getPastEvents()` is called

contracts/StakeManager.sol
```diff
@@ -54,16 +54,19 @@ contract StakeManager is IStakeManager {
         emit StakeWithdrawn(relayManager, msg.sender, amount);
     }
 
-    function authorizeHub(address relayManager, address relayHub) external override {
-        StakeInfo storage info = stakes[relayManager];
-        require(info.owner == msg.sender, "not owner");
+    modifier ownerOrManagerOnly(address relayManager) {
+        StakeInfo storage infoByManager = stakes[relayManager];
+        StakeInfo storage infoBySender = stakes[msg.sender];
+        require(infoByManager.owner == msg.sender || infoBySender.owner != address(0), "not owner or manager");
```
forshtat, line 60: Note the bug discussed here (for the record)

shahafn: Fixed

src/relayserver/RelayServer.ts
```diff
@@ -486,52 +469,33 @@ export class RelayServer extends EventEmitter {
...
+    await this._resendUnconfirmedTransactions(blockHeader)
     if (this.stake.eq(toBN(0))) {
       throw new StateError('Waiting for stake')
     }
-    // todo check if registered!!
-    // TODO: now even more todo then before. This is a hotfix.
-    if (!this.isAddressAdded) {
+
+    const registered = await this._isRegistered()
+    if (!registered) {
```
drortirosh, line 479: different state errors handled differently... why?

src/relayserver/RelayServer.ts
```diff
@@ -589,8 +575,22 @@ export class RelayServer extends EventEmitter {
     return await this._registerIfNeeded()
   }
 
+  async _handleHubUnauthorizedEvent (dlog: DecodeLogsEvent): Promise<TransactionReceipt[]> {
+    if (dlog.name !== 'HubUnauthorized' || dlog.args.relayManager.toLowerCase() !== this.managerAddress.toLowerCase()) {
+      // eslint-disable-next-line @typescript-eslint/restrict-template-expressions
+      throw new Error(`PANIC: handling wrong event ${dlog.name} or wrong event relay ${dlog.args.relay}`)
```
drortirosh, line 581: error doesn't match parameter you check...

test/StakeManager.test.ts
```diff
@@ -173,28 +173,48 @@ contract('StakeManager', function ([_, relayManager, anyRelayHub, owner, nonOwne
...
+    describe('should not allow not owner to call to', function () {
+      it('authorize hub by owner', async function () {
+        await expectRevert(stakeManager.authorizeHubByOwner(relayManager, anyRelayHub, { from: nonOwner }), 'not owner')
+      })
+      it('unauthorize hub by owner', async function () {
+        await expectRevert(stakeManager.unauthorizeHubByOwner(relayManager, anyRelayHub, { from: nonOwner }), 'not owner')
+      })
+    })
+
+    describe('should not allow not manager or to call to', function () {
```
drortirosh, line 210: title typo..

PR #413, opened 2020-06-23
Paymaster data

drortirosh:


contracts/interfaces/GsnTypes.sol
```diff
@@ -10,6 +10,8 @@ interface GsnTypes {
         uint256 baseRelayFee;
         address relayWorker;
         address paymaster;
+        bytes32 paymasterData;
```
forshtat, line 13: bytes

bytes32 may be not enough for some use-cases

contracts/utils/GsnEip712Library.sol
```diff
@@ -14,7 +14,7 @@ library GsnEip712Library {
     //copied from Forwarder (can't reference string constants even from another library)
     string public constant GENERIC_PARAMS = "address to,bytes data,uint256 value,address from,uint256 nonce,uint256 gas";
 
-    bytes public constant RELAYDATA_TYPE = "RelayData(uint256 gasPrice,uint256 pctRelayFee,uint256 baseRelayFee,address relayWorker,address paymaster)";
+    bytes public constant RELAYDATA_TYPE = "RelayData(uint256 gasPrice,uint256 pctRelayFee,uint256 baseRelayFee,address relayWorker,address paymaster,bytes32 paymasterData,uint256 clientId)";
```
forshtat, line 17: add forwarder please please please please

src/relayclient/GSNConfigurator.ts
```diff
@@ -98,20 +100,30 @@ export function getDependencies (config: GSNConfig, provider?: HttpProvider, ove
...
     transactionValidator,
     pingFilter,
     relayFilter,
-    asyncApprovalData,
+    asyncPaymasterData,
     scoreCalculator,
     config
   }
+
+  // sanity check: overrides must on contain unknown fields.
```
forshtat, line 121: 1. Why not? It can not be accessed by our TypeScript code anyways
2. TypeScript compiler does run such checks for you

src/relayclient/types/Aliases.ts
```diff
@@ -12,8 +12,15 @@ export type IntString = string
  * For legacy reasons, to filter out the relay this filter has to throw.
  * TODO: make ping filtering sane!
  */
+
+export interface PaymasterData {
+  approvalData? : PrefixedHexString
+  paymasterData?: PrefixedHexString
+}
```
forshtat, line 19: PaymasterData.paymasterData? I don't think so

src/relayclient/RelayClient.ts
```diff
@@ -15,9 +15,14 @@ import RelayedTransactionValidator from './RelayedTransactionValidator'
...
-export const EmptyApprovalData: AsyncApprovalData = async (): Promise<PrefixedHexString> => {
-  return await Promise.resolve('0x')
+export const ZERO_BYTES32 = '0x'.padEnd(66, '0')
+
+// generate "approvalData" and "paymasterData" for a request.
+// paymasterData must be valid bytes32 value (defaults to zero-bytes)
+// approvalData is string-encoded bytes array, (defaults to empty array)
+// either field can be missing, and filled with default
+export const EmptyPaymasterData: AsyncPaymasterData = async (): Promise<PaymasterData> => {
+  return await Promise.resolve({})
```
forshtat, line 25: There are exactly 0 reasons to put these 2 in one object in my view

src/relayclient/RelayClient.ts
```diff
@@ -226,12 +231,19 @@ export default class RelayClient {
...
         paymaster,
+        paymasterData: '', // temp value. filled in by asyncPaymasterData, below.
+        clientId: this.config.clientId,
         forwarder: forwarderAddress,
         relayWorker
       }
     }
+    const pmData = await this.asyncPaymasterData(relayRequest)
+    const paymasterData = pmData.paymasterData ?? ZERO_BYTES32
+    const approvalData = pmData.approvalData ?? '0x'
```
forshtat, line 242: meh

test/RelayHubGasCalculations.test.ts
```diff
@@ -32,12 +33,14 @@ contract('RelayHub gas calculations', function ([_, relayOwner, relayWorker, rel
...
 
   const senderNonce = new BN('0')
   const magicNumbers = {
-    arc: 928,
-    pre: 1552,
-    post: 1591
+    arc: 907 + 22,
+    pre: 1486 + 66,
+    post: 1613 - 22
   }
```
forshtat, line 44: https://www.calculator.net/

src/relayclient/RelayClient.ts
```diff
@@ -15,9 +15,14 @@ import RelayedTransactionValidator from './RelayedTransactionValidator'
 import { configureGSN, getDependencies, GSNConfig, GSNDependencies } from './GSNConfigurator'
 import { RelayInfo } from './types/RelayInfo'
 
-// generate "approvalData" for a request. must return string-encoded bytes array
-export const EmptyApprovalData: AsyncApprovalData = async (): Promise<PrefixedHexString> => {
-  return await Promise.resolve('0x')
+export const ZERO_BYTES32 = '0x'.padEnd(66, '0')
```
forshtat, line 18: Aaand you won't be needing this of course

contracts/interfaces/GsnTypes.sol
drortirosh: if you want more signed data that bytes32, then put the rest into approvalData, and just put a hash here. 
No need to make the basic structure overly complex.

src/relayclient/GSNConfigurator.ts
drortirosh: 1. typescript compiler missed it in my test
2. your clients might be using javascript, not typescript, and you hide them from a breaking change (they can set fields that silently gets completely ignored)
3. need to find a library that does runtime schema check, and sanitize input params. can't rely on transpiled language.

src/relayclient/types/Aliases.ts
drortirosh: `interface PaymasterInfo` ?

src/relayclient/RelayClient.ts
drortirosh: they are both dynamic parameters, per-request that are only visible by paymaster's `acceptRelayedCall`.
Why separate them?

src/relayclient/RelayClient.ts
drortirosh: mah ?

src/relayclient/RelayClient.ts
drortirosh: might want to move it to some constants file, but we don't have any in client, only in test.

contracts/interfaces/GsnTypes.sol
forshtat: No need to make client logic overly complex. 
It makes total sense to have bytes and it has zero cost, code or otherwise.

contracts/utils/GsnEip712Library.sol
forshtat: Any response here?

src/relayclient/GSNConfigurator.ts
forshtat: 1. Check your code, do not introduce "sanity checks" to the code you are working with
2. If clients are using JavaScript their lives will be filled with pain and misery and I won't take on myself a mission to make their lives any better.
3. Not as part of Paymaster Data PR, ok? On your free time.

src/relayclient/types/Aliases.ts
forshtat: no interface whatsoever

src/relayclient/RelayClient.ts
forshtat: why unite them in interface?

test/RelayHubGasCalculations.test.ts
forshtat: please use

src/relayclient/RelayClient.ts
forshtat: remove

contracts/interfaces/GsnTypes.sol
drortirosh: fixed

contracts/utils/GsnEip712Library.sol
drortirosh: I still see no reason to expose it - even though we already thrown a lot of garbage into the struct, its no reason to throw out more.

src/relayclient/GSNConfigurator.ts
drortirosh: added it because *I* hit this issue. added from the egoistic attempt to make my life easier (true, in earlier format of this PR, where I changed the asyncApprovalData to something else.)
Still, the minimum support we can give users is that if they use our lib, and later we change the API, then at least we should trigger an error if we notice they use an old code (I'm not saying we should make the lib backward-compatible, which is something that sometimes is useful, but at least fail fast.)

src/relayclient/types/Aliases.ts
drortirosh: removed

src/relayclient/RelayClient.ts
drortirosh: removed

PR #416, opened 2020-06-28
GSN-119 Test calling payable function

forshtat:


test/relayclient/RelayProvider.test.ts
```diff
@@ -145,7 +147,32 @@ contract('RelayProvider', function (accounts) {
...
-        realSender: gasLess
+        realSender: gasLess,
+        msgValue: '0',
+        balance: '0'
+      })
+    })
+
+    it('should relay transparently with value', async function () {
+      const value = 1e18.toString()
+      await web3.eth.sendTransaction({
```
drortirosh, line 158: at a minimum, at a comment explaining that it should be a paymaster logic to pass value into the forwrader...
otherwise, this is a WTF-test

forshtat: Well, it tests the behavior of provider, client and forwarder. I came to really hate in-code comments explaining things.

drortirosh: I don't agree. you write "proper" test (e.g paymaster , etc) - or explain why you don't.
Our test code is usage sample

drortirosh: right now, you read and understand it. few days from now, someone will look and really say WTF - why moving money like that, security-racing-whatever. 
At a minimum, say that actual transfer should be done by paymaster (probably by exchanging user's tokens)

drortirosh: I still think this WTF code deserves at least a comment (if we're not adding a proper paymaster)

forshtat: You did add a comment yourself, didn't you, Dror?)

PR #418, opened 2020-06-30
Remove @openzeppelin/test-helpers from relayclient code

dmihal:
The @openzeppelin/test-helpers package is not optimized for client-side projects and has many dependencies. The inclusion of this package causes a number of build issues in webpack.

Furthermore, this package is only used for the ZERO_ADDRESS constant. So I've removed this dependency and hardcoded the constant.


forshtat:
Hello @dmihal 
This is obviously a mistake and should be fixed. 
Correct me if I'm wrong, but there are a few more places across the code where zero-address gets imported from test-helpers. Should those be cleaned as well for webpack to work better? 
But also a question, what are the dependencies that create issues for the webpack? For context, we barely use webpack internally but it would be great if we could make GSN library as lightweight as possible in terms of dependencies.
Thanks!

PR #422, opened 2020-07-06
OG-14 GsnTestEnvironment is not exported properly for TypeScript depe…

forshtat:
…ndency

TypeScript/ES6 is missing support for wildcard export of default import.
This means we cannot use default export for the top-level components.

### DIFF
--- a/src/relayclient/RelayClient.ts
+++ b/src/relayclient/RelayClient.ts
@@ -42,7 +42,7 @@ export interface RelayingResult {
   relayingErrors: Map<string, Error>
 }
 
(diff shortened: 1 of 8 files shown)


src/relayclient/RelayClient.ts
```diff
@@ -42,7 +42,7 @@ export interface RelayingResult {
   relayingErrors: Map<string, Error>
 }
 
-export default class RelayClient {
+export class RelayClient {
```
shahafn, line 45: Finally

PR #425, opened 2020-07-08
OG-113 Implement VersionsManager and use it in client

forshtat:


src/relayclient/ContractInteractor.ts
```diff
@@ -28,6 +29,7 @@ import { GSNConfig } from './GSNConfigurator'
 import GsnTransactionDetails from './types/GsnTransactionDetails'
 import { BlockTransactionString } from 'web3-eth'
 import Common from 'ethereumjs-common'
+import { constants } from '@openzeppelin/test-helpers'
```
shahafn, line 32: Aren't we in the process of removing these?

PR #426, opened 2020-07-08
OG-105 Remove OZ/TestHelpers from production dependencies

forshtat:
fully resolves #418


dmihal:
@forshtat Do you know when this change will get published to NPM?

Looks like there was new release 14 days ago, but this change wasn't included

forshtat:
> @forshtat Do you know when this change will get published to NPM?
> 
> Looks like there was new release 14 days ago, but this change wasn't included

I think this will be released soon, but if this is blocking your workflow we can prioritise this for the next week.

PR #429, opened 2020-07-12
OG-100: New Paymaster API

drortirosh:
- removed acceptRelayedCall
- use preRelayedCall instead.


contracts/BasePaymaster.sol
```diff
@@ -40,23 +43,20 @@ abstract contract BasePaymaster is IPaymaster, Ownable {
...
-    )
+    // this method must be called from preRelayedCall to validate that the forwarder
+    // is approved by the paymaster as well as by the recipient contract.
+    function _verifyForwarder(GsnTypes.RelayRequest calldata relayRequest)
     public
     view
     {
         require(address(trustedForwarder) == relayRequest.relayData.forwarder, "Forwarder is not trusted");
-        GsnEip712Library.verify(relayRequest, signature);
+        GsnEip712Library.verifyForwarderTrusted(relayRequest);
```
shahafn, line 59: Why don't we need signature verification here?

contracts/BasePaymaster.sol
```diff
@@ -26,10 +26,13 @@ abstract contract BasePaymaster is IPaymaster, Ownable {
...
     }
 
-    // Gas stipends for acceptRelayedCall, preRelayedCall and postRelayedCall
-    uint256 constant private ACCEPT_RELAYED_CALL_GAS_LIMIT = 50000;
-    uint256 constant private PRE_RELAYED_CALL_GAS_LIMIT = 100000;
-    uint256 constant private POST_RELAYED_CALL_GAS_LIMIT = 110000;
+    //Paymaster is commited to pay for any reverted transaction above the commitment. it covers both preRelayedCall aod
+    // forwrader's check for nonce/signature.
+    // we assume 50k is more than enough for forwarder (10k bytes request takes ~30kgas)
+    uint256 constant public COMMITMENT_GAS_LIMIT = 150000;
```
shahafn, line 32: Please document the importance of the diff between commitment and pre gas. Also add tests for that.

contracts/RelayHub.sol
```diff
@@ -138,7 +134,7 @@ contract RelayHub is IRelayHub {
             initialGas >= maxPossibleGas,
             "Not enough gas left for innerRelayCall to complete");
 
-        uint256 maxPossibleCharge = calculateCharge(
+        uint maxPossibleCharge = calculateCharge(
```
shahafn, line 137: uint256 for consistency?

contracts/RelayHub.sol
```diff
@@ -292,41 +303,53 @@ contract RelayHub is IRelayHub {
             // acceptRelayedCall that this will not happen.
             (success, retData) = relayRequest.relayData.paymaster.call{gas:gasLimits.preRelayedCallGasLimit}(atomicData.data);
             if (!success) {
-                revertWithStatus(RelayCallStatus.PreRelayedFailed, GsnEip712Library.getTruncatedData(retData));
+                revertWithStatus(RelayCallStatus.PreRelayedFailed, retData);
             }
-            atomicData.preReturnValue = abi.decode(retData, (bytes32));
+            (atomicData.recipientContext, atomicData.isTrustedRecipient) = abi.decode(retData, (bytes,bool));
```
shahafn, line 308: Maybe rename to revertOnRecipientRevert? since that's what distinguishes between "pre" mode and "recipient" mode per the doc

contracts/RelayHub.sol
```diff
@@ -292,41 +303,53 @@ contract RelayHub is IRelayHub {
             // acceptRelayedCall that this will not happen.
             (success, retData) = relayRequest.relayData.paymaster.call{gas:gasLimits.preRelayedCallGasLimit}(atomicData.data);
             if (!success) {
-                revertWithStatus(RelayCallStatus.PreRelayedFailed, GsnEip712Library.getTruncatedData(retData));
+                revertWithStatus(RelayCallStatus.PreRelayedFailed, retData);
```
shahafn, line 306: Why don't we keep truncating it?

contracts/RelayHub.sol
```diff
@@ -292,41 +303,53 @@ contract RelayHub is IRelayHub {
...
         }
 
         // The actual relayed call is now executed. The sender's address is appended at the end of the transaction data
-        (atomicData.relayedCallSuccess, atomicData.relayedCallReturnValue) = GsnEip712Library.execute(relayRequest, signature);
 
+        {
+            bool forwarderSuccess;
+            (forwarderSuccess, atomicData.relayedCallSuccess, atomicData.relayedCallReturnValue) = GsnEip712Library.execute(relayRequest, signature);
+            if ( !forwarderSuccess ) {
+                revertWithStatus(RelayCallStatus.ForwarderFailed, atomicData.relayedCallReturnValue);
```
shahafn, line 317: Should be truncated as well?

contracts/RelayHub.sol
```diff
@@ -292,41 +303,53 @@ contract RelayHub is IRelayHub {
...
             relayRequest.relayData
         );
 
-        (bool successPost,) = relayRequest.relayData.paymaster.call{gas:gasLimits.postRelayedCallGasLimit}(atomicData.data);
+        {
+        (bool successPost,bytes memory ret) = relayRequest.relayData.paymaster.call{gas:gasLimits.postRelayedCallGasLimit}(atomicData.data);
 
         if (!successPost) {
-            revertWithStatus(RelayCallStatus.PostRelayedFailed, atomicData.relayedCallReturnValue);
+            revertWithStatus(RelayCallStatus.PostRelayedFailed, ret);
```
shahafn, line 340: Should truncate here as well?

contracts/RelayHub.sol
```diff
@@ -292,41 +303,53 @@ contract RelayHub is IRelayHub {
...
+        }
         }
 
         if (balances[relayRequest.relayData.paymaster] < atomicData.balanceBefore) {
-            revertWithStatus(RelayCallStatus.RecipientBalanceChanged, atomicData.relayedCallReturnValue);
+            revertWithStatus(RelayCallStatus.RecipientBalanceChanged, "");
```
shahafn, line 345: Todo: Should change to "paymasterBalanceChanged

contracts/test/TestPaymasterConfigurableMisbehavior.sol
```diff
@@ -32,17 +32,16 @@ contract TestPaymasterConfigurableMisbehavior is TestPaymasterEverythingAccepted
         overspendAcceptGas = val;
     }
 
-    function acceptRelayedCall(
+    function preRelayedCall(
```
shahafn, line 35: Should add _verifyForwarder check

contracts/test/TestPaymasterOwnerSignature.sol
```diff
@@ -12,22 +12,20 @@ contract TestPaymasterOwnerSignature is TestPaymasterEverythingAccepted {
     /**
      * This demonstrates how dapps can provide an off-chain signatures to relayed transactions.
      */
-    function acceptRelayedCall(
+    function preRelayedCall(
```
shahafn, line 15: _verifyForwarder()

contracts/test/TestPaymasterVariableGasLimits.sol
```diff
@@ -19,47 +18,31 @@ contract TestPaymasterVariableGasLimits is TestPaymasterEverythingAccepted {
         uint256 gasUseWithoutPost
     );
 
-    function acceptRelayedCall(
+    function preRelayedCall(
```
shahafn, line 21: _verifyForwarder

test/PaymasterCommitment.test.ts
```diff
@@ -0,0 +1,336 @@
...
+  }
+}
+
+// verify the paymaster's commitment:
+// - PM always pay for non-reverted TXs (either high or low gas use)
+// - if preRelayedCall reverts: PM always pay (=as long as commitment>preRelayedCallGasLimit)
+// - standard forwarder reverts: PM always pay (since commitment > gas of (preRelayedCall,forwarder))
+// - nonstandard forwardeR: PM pays above commitment
+// - trusted recipient: PM pays above commitment.
+contract('Paymaster Commitment', function ([_, relayOwner, relayManager, relayWorker, senderAddress, other]) { // eslint-disable-line no-unused-vars
```
shahafn, line 66: Should add forwarder that overspends the commitment gas

test/RelayHubGasCalculations.test.ts
```diff
@@ -264,9 +269,9 @@ contract('RelayHub gas calculations', function ([_, relayOwner, relayWorker, rel
       await relayHub.depositFor(relayOwner, { value: (1).toString() })
     });
 
-    [0, 100, 1000, 50000]
+    [0, /* 100, */ 1000, 50000]
```
shahafn, line 272: Unrelated to this PR

contracts/interfaces/IPaymaster.sol
```diff
@@ -34,65 +36,53 @@ interface IPaymaster {
     );
 
     /**
-     * Called by Relay (and RelayHub), to validate if this recipient accepts this call.
+     * Called by Relay (and RelayHub), to validate if the paymaster accepts paying this call.
```
forshtat, line 39: 'agrees to pay for' ?

contracts/interfaces/IPaymaster.sol
```diff
@@ -34,65 +36,53 @@ interface IPaymaster {
...
      *  @param approvalData - extra dapp-specific data (e.g. signature from trusted party)
      *  @param maxPossibleGas - based on values returned from {@link getGasLimits},
      *         the RelayHub will calculate the maximum possible amount of gas the user may be charged for.
      *         In order to convert this value to wei, the Paymaster has to call "relayHub.calculateCharge()"
-     *  @return a context to be passed to preRelayedCall and postRelayedCall.
+     *  return:
+     *      a context to be passed to postRelayedCall
+     *      revertOnRecipientRevert - TRUE if paymaster want to reject the TX if the recipient reverts.
+     *          FALSE means that rejects by the recipient will be completed on chain, and paid by the paymaster.
+     *          (note that in this case, the preRelayedCall and postRelayedCall are not reverted).
```
forshtat, line 58: why not stick to `@return` natspec syntax?

contracts/interfaces/IRelayHub.sol
```diff
@@ -73,12 +74,16 @@ interface IRelayHub {
...
+    /// @param ForwarderFailed - the transaction was not relayed due to forwrader check (signature,nonce)
+    /// @param TrustedRecipientFailed - the transaction was not relayed since Paymaster refuse to relay reverted recipient TX.
     /// @param PostRelayedFailed - the transaction was relayed and reverted due to postRelatedCall reverting
     /// @param RecipientBalanceChanged - the transaction was relayed and reverted due to the recipient balance change
     enum RelayCallStatus {
         OK,
         RelayedCallFailed,
         PreRelayedFailed,
+        ForwarderFailed,
+        RecipientFailed,
```
forshtat, line 86: What is the difference between `RelayedCallFailed` and `RecipientFailed`?

contracts/RelayHub.sol
drortirosh: renamed.

test/RelayHubGasCalculations.test.ts
drortirosh: removed

contracts/BasePaymaster.sol
```diff
@@ -26,10 +26,13 @@ abstract contract BasePaymaster is IPaymaster, Ownable {
...
+    //any revert in preRelayedCall is within "commitment"
+    uint256 constant public PRE_RELAYED_CALL_GAS_LIMIT = 100000;
+    uint256 constant public POST_RELAYED_CALL_GAS_LIMIT = 110000;
```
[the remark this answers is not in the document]
drortirosh: we move all documentation to the IPaymaster.getGasLimits(), and reference it here .

contracts/BasePaymaster.sol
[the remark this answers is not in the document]
drortirosh: verify used to execute the forwarder.verify, but verofyForwraderTrusted is not.
Actually, with the new paymaster model, we can completely remove the "verify()" method of the forwrader.

contracts/RelayHub.sol
```diff
@@ -187,42 +182,54 @@ contract RelayHub is IRelayHub {
...
         // Calls to the recipient are performed atomically inside an inner transaction which may revert in case of
         // errors in the recipient. In either case (revert or regular execution) the return data encodes the
         // RelayCallStatus value.
-        (, bytes memory relayCallStatus) = address(this).call{gas:innerGasLimit}(
-            abi.encodeWithSelector(RelayHub.innerRelayCall.selector, relayRequest, signature, vars.gasLimits,
-                innerGasLimit + externalGasLimit-gasleft() + gasOverhead + postOverhead, /*totalInitialGas*/
-                abi.decode(vars.recipientContext, (bytes)))
+        (bool success, bytes memory relayCallStatus) = address(this).call{gas:innerGasLimit}(
+            abi.encodeWithSelector(RelayHub.innerRelayCall.selector, relayRequest, signature, approvalData, vars.gasLimits,
+                innerGasLimit + vars.externalGasLimit - gasleft() + gasOverhead + postOverhead, /*totalInitialGas*/
```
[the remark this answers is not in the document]
drortirosh: in the outer call, we do: externalGasLimit-gasLeft() + gasOverhead
here we want to do the same, with 2 exceptions:
- we don't know how much postRelayedCall will cost, so we dont' want to have it in the calculation
- we're one level deeper in the stack, so we want to add also the gas usage just before the call
  - externalGasLimit-gasLeft() - this is the gas used so far.
  - innerGasLimit-gasleft() (when gasleft() called INSIDE InternalRelayedCall) - is gas used by internalRelayedCall

so we just sum them up, and send a single var (totalInitialGas) into innerRelayedCall)

the overhead can be replaced with a single prePostOveread, instead of summing them up each time (maybe it will be clearer, since the relation of the 2 overheads is not clear here)

contracts/RelayHub.sol
```diff
@@ -187,42 +182,54 @@ contract RelayHub is IRelayHub {
...
+                vars.maxPossibleGas
+                )
         );
+        vars.success = success;
+        vars.innerGasUsed = vars.gasBeforeInner-gasleft();
         (vars.status, vars.relayedCallReturnValue) = abi.decode(relayCallStatus, (RelayCallStatus, bytes));
     }
     {
+        if (!vars.success) {
+//            (vars.status, relayCallStatus) = abi.decode(relayCallStatus, (RelayCallStatus, bytes));
```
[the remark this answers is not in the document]
drortirosh: removed

contracts/RelayHub.sol
```diff
@@ -187,42 +182,54 @@ contract RelayHub is IRelayHub {
...
+            //Failure cases where the PM doesn't pay
+            if ( (vars.innerGasUsed < vars.gasLimits.commitmentGasLimit ) && (
+                    vars.status == RelayCallStatus.PreRelayedFailed ||
+                    vars.status == RelayCallStatus.ForwarderFailed ||
+                    vars.status == RelayCallStatus.RecipientFailed  //can only be thrown if our preRelayedCall trusted it
```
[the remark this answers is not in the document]
drortirosh: //can only be thrown if revertOnRecipientRevert==true

contracts/interfaces/IPaymaster.sol
drortirosh: thanks. sounds better.

contracts/interfaces/IPaymaster.sol
drortirosh: because @return doesn't work for me when returning a tupple.

contracts/interfaces/IRelayHub.sol
drortirosh: there is, and I wonder if we should keep it:
- RecipientFailed is returned on `revertOnRecipientRevets` (gas calculation is done outside, but the "intent" is inside innerRelayedCall)
- RecipientCallFailed is returned if the paymaster want to pay anyway (that is, both pre and post are executed and not reverted)

I guest that even if we leave them both in, they need better naming..

contracts/RelayHub.sol
forshtat: @drortirosh you resolved issue but did not change code

contracts/RelayHub.sol
```diff
@@ -113,22 +113,18 @@ contract RelayHub is IRelayHub {
         emit Withdrawn(account, dest, amount);
     }
 
-    function canRelay(
+    function checkGasLimits(
```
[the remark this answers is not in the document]
forshtat: verifyGasLimits?

contracts/RelayHub.sol
```diff
@@ -189,43 +184,55 @@ contract RelayHub is IRelayHub {
...
-            return (vars.success, revertReason);
-        }
-
-        // From this point on, this transaction will not revert nor run out of gas, and the paymaster will be charged
-        // for the gas spent.
+        (vars.gasLimits, vars.maxPossibleGas) =
+            checkGasLimits(relayRequest, externalGasLimit);
 
     {
+        vars.externalGasLimit = externalGasLimit;
```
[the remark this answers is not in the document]
forshtat: But this variable is already defined inside the block, you don't free up anything on the stack, do you?

contracts/RelayHub.sol
```diff
@@ -113,22 +113,18 @@ contract RelayHub is IRelayHub {
...
-    {
+    returns (IPaymaster.GasLimits memory gasLimits, uint256 maxPossibleGas) {
         gasLimits =
             IPaymaster(relayRequest.relayData.paymaster).getGasLimits();
-        uint256 maxPossibleGas =
+        maxPossibleGas =
             gasOverhead +
-            gasLimits.acceptRelayedCallGasLimit +
-            gasLimits.preRelayedCallGasLimit +
+            gasLimits.commitmentGasLimit +
```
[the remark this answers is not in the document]
forshtat: There is no other side to this. Nothing forces you to make commitmentGL > preRelayCallGL, so in general case the maxPossibleGas is calculated incorrectly here.

contracts/interfaces/IRelayHub.sol
forshtat: OMG, this is extremely confusing. Like, this flag is exposed upwards via a new value for the enum. Can we find a better solution?

contracts/RelayHub.sol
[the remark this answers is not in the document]
forshtat: By the way, even if you set commitmentGL to 20'000'000 or whatever, no method is given that much, so it is not part of the maxPossibleGas

contracts/interfaces/IPaymaster.sol
```diff
@@ -6,12 +6,40 @@ import "./GsnTypes.sol";
...
-        uint256 acceptRelayedCallGasLimit;
+        //paymaster is committed to pay reverted transactions above this gas limit.
+        // This limit should cover both preRelayedCall gaslimit AND forwarder's nonce and signature validation.
+        uint256 paymasterPaysAbove;
         uint256 preRelayedCallGasLimit;
         uint256 postRelayedCallGasLimit;
     }
 
+    /**
+     * The RelayHub will call accept-, pre-, and post-, RelayCall methods with these values for their gas limits.
```
forshtat, line 34: accept- ? This (new) comment is outdated

contracts/BasePaymaster.sol
```diff
@@ -26,37 +25,37 @@ abstract contract BasePaymaster is IPaymaster, Ownable {
...
+    uint256 constant public FORWARDER_HUB_OVERHEAD = 50000;
+
+    //These parameters are documented in IPaymaster.GasLimits
+    uint256 constant public PRE_RELAYED_CALL_GAS_LIMIT = 100000;
+    uint256 constant public POST_RELAYED_CALL_GAS_LIMIT = 110000;
+    uint256 constant public PAYMASTER_PAYS_ABOVE = PRE_RELAYED_CALL_GAS_LIMIT + FORWARDER_HUB_OVERHEAD;
 
     function getGasLimits()
-    external
+    public
```
drortirosh, line 37: changed to public to allow easy accessing from a subclass:
https://github.com/opengsn/gsn/pull/429/files#diff-77990bc92cd4dea5097e507e088f3048R92
```
IPaymaster.GasLimits private limits = super.getGasLimits();
```

contracts/RelayHub.sol
```diff
@@ -174,6 +168,7 @@ contract RelayHub is IRelayHub {
     override
     returns (bool paymasterAccepted, string memory revertReason)
     {
+        (signature);
```
forshtat, line 171: signature is used in the method body, no need to silence it

contracts/RelayHub.sol
```diff
@@ -186,43 +181,54 @@ contract RelayHub is IRelayHub {
...
+                )
         );
+        vars.success = success;
+        vars.innerGasUsed = vars.gasBeforeInner-gasleft();
         (vars.status, vars.relayedCallReturnValue) = abi.decode(relayCallStatus, (RelayCallStatus, bytes));
     }
     {
+        if (!vars.success) {
+            //Failure cases where the PM doesn't pay
+            if ( (vars.innerGasUsed < vars.gasLimits.acceptanceBudget ) && (
```
forshtat, line 212: So is it 'acceptanceBudget' or 'paymasterPaysAbove'? The only thing worse then a bad name is two bad names for the same variable.

PR #435, opened 2020-07-15
OG-70 Emit the return value/revert reason in TransactionRelayed event

forshtat:


contracts/RelayHub.sol
```diff
@@ -294,7 +295,7 @@ contract RelayHub is IRelayHub {
             // acceptRelayedCall that this will not happen.
             (success, retData) = relayRequest.relayData.paymaster.call{gas:gasLimits.preRelayedCallGasLimit}(atomicData.data);
             if (!success) {
-                revertWithStatus(RelayCallStatus.PreRelayedFailed);
+                revertWithStatus(RelayCallStatus.PreRelayedFailed, bytes(""));
```
drortirosh, line 298: any special reason you *don't* want to show the revert reason of preRelayedCall ?

contracts/RelayHub.sol
```diff
@@ -317,21 +318,21 @@ contract RelayHub is IRelayHub {
...
+            revertWithStatus(RelayCallStatus.PostRelayedFailed, atomicData.relayedCallReturnValue);
         }
 
         if (balances[relayRequest.relayData.paymaster] < atomicData.balanceBefore) {
-            revertWithStatus(RelayCallStatus.RecipientBalanceChanged);
+            revertWithStatus(RelayCallStatus.RecipientBalanceChanged, atomicData.relayedCallReturnValue);
         }
 
-        return atomicData.relayedCallSuccess ? RelayCallStatus.OK : RelayCallStatus.RelayedCallFailed;
+        return (atomicData.relayedCallSuccess ? RelayCallStatus.OK : RelayCallStatus.RelayedCallFailed, atomicData.relayedCallReturnValue);
```
drortirosh, line 328: I still think the "success" check should also be applied to the returned value:
e.g.
```
if ( !atomicData.relayedCallSuccess ) {
  return (RelayCallStatus.RelayedCallFailed, atomicData.relayedCallReturnValue);
} else {
  return (RelayCallStatus.OK, "");
}
```

contracts/RelayHub.sol
forshtat: Not really. Fixing)

contracts/RelayHub.sol
forshtat: I know, but that would be burning the 'relayedCallReturnValue' for idealogical reasons mostly

PR #436, opened 2020-07-15
 OG-128 Make 'gsn deploy' use flattened source tree compilation result

forshtat:


.circleci/config.yml
```diff
@@ -8,6 +8,10 @@ jobs: # a collection of steps
     steps: # a collection of executable commands
       - checkout # special step to check out source code to working directory
 
+      - run: # run tests
+          name: global-install
```
drortirosh, line 12: the build tool should also run locally. you make the tool depend on global packages... not good.

scripts/compile_flattened.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash -ex
+
+PRAGMA="// SPDX-License-Identifier:MIT\npragma solidity ^0.6.10;\npragma experimental ABIEncoderV2;"
+CONTRACTS=("RelayHub" "StakeManager" "Penalizer" "Forwarder" "TestPaymasterEverythingAccepted")
+
+# ok, I have no idea how to do it right; if it's working it ain't stupid though
+PATHS=("./" "./" "./" "./forwarder/" "./test/")
```
drortirosh, line 7: I agree. you proved your point - you're not good at writing shell scripts....

but seriously, I don't think we should use this method as main compile-for-deployment, so it doesn't really matter.

drortirosh: FWIW, I'd give the paths tot he contract names themselves, instead of having two arrays scanned by index, and merge them...
```
for c in RelayHub StakeManager forwarder/Forwarder test/TestPaymaster; do echo $c.sol; done
```

scripts/compile_flattened.sh
```diff
@@ -0,0 +1,22 @@
...
+OUTFOLDER=./src/cli/compiled
+
+mkdir -p $TMPFOLDER
+
+END=${#CONTRACTS[@]}-1
+for ((i=0;i<=END;i++)); do
+  c=${CONTRACTS[$i]}
+  c_flat=$TMPFOLDER"${c}"_flat.sol
+  echo -e $PRAGMA > $c_flat
+  truffle-flattener $INFOLDER/${PATHS[$i]}$c.sol | grep -v pragma | grep -v SPDX >> $c_flat
```
drortirosh, line 20: if flattener is that broken that it doesn't handle the source code well, then we should't depend on it.
(e.g. we KNOW the code changes because of the use of pragma... so why do we test one thing, and deploy another?)

forshtat:
Closed in favour of JSON input verification

PR #444, opened 2020-07-23
OG-125 configure registration block rate in server

shahafn:


src/relayserver/runServer.ts
```diff
@@ -68,6 +69,8 @@ const ethereumNodeUrl: string = argv.ethereumNodeUrl ?? config.ethereumNodeUrl ?
 const workdir: string = argv.workdir ?? config.workdir ?? error('missing --workdir')
 const devMode: boolean = argv.devMode ?? config.devMode ?? error('missing --devMode')
 const debug: boolean = argv.debug ?? config.debug ?? error('missing --debug')
+const registrationBlockRate: string = argv.registrationBlockRate ?? config.registrationBlockRate?.toString()
+console.log('wtf is registrationBlockRate?', registrationBlockRate)
```
drortirosh, line 73: please be respectful with "wtf". if you do need it, put it in "debug"

drortirosh: or better: when the server starts, dump its complete active configuration.

test/RelayServer.test.ts
```diff
@@ -1109,43 +1111,72 @@ contract('RelayServer', function (accounts) {
...
-        assertRelayAdded(receipts, newServer, false)
-        // @ts-ignore
-        newServer.pctRelayFee++
-        receipts = await newServer._registerIfNeeded()
-        assertRelayAdded(receipts, newServer, false)
-        // @ts-ignore
-        newServer.url = 'fakeUrl'
-        receipts = await newServer._registerIfNeeded()
-        assertRelayAdded(receipts, newServer, false)
+      describe('with re-registration', function () {
```
drortirosh, line 1160: I'm trying to understand why you re-run all tests again, where the only test that can be affected is the added one.
why not just modify its configuration to see that (say) after 900 blocks i doens't re-register, and after 1100 it does ?

src/relayserver/runServer.ts
shahafn: Or better, have a git hook not to commit it

src/relayserver/RelayServer.ts
```diff
@@ -86,9 +88,12 @@ export interface RelayServerParams {
   readonly baseRelayFee: number | undefined
   readonly pctRelayFee: number | undefined
   readonly gasPriceFactor: number
+  readonly registrationBlockRate?: number
   readonly url?: string
   readonly workerMinBalance: number | undefined // = defaultWorkerMinBalance,
   readonly workerTargetBalance: number | undefined // = defaultWorkerTargetBalance,
+  readonly managerTargetBalance: number | undefined // = defaultManagerTargetBalance,
+  readonly minHubWithdrawalBalance: number | undefined // = defaultManagerTargetBalance,
```
forshtat, line 96: You use  '?:' and '| undefined' interchangeably. I think these are equivalent and it's better to stick to one syntax in general...

src/relayserver/RelayServer.ts
```diff
@@ -471,19 +488,19 @@ export class RelayServer extends EventEmitter {
       throw new StateError('Could not get gasPrice from node')
     }
     const balance = await this.getManagerBalance()
-    if (balance.lt(toBN(minimumRelayBalance))) {
+    if (balance.lt(toBN(defaultMinManagerBalance))) {
       throw new StateError(
         // eslint-disable-next-line @typescript-eslint/restrict-template-expressions
-        `Server's balance too low ( ${balance}, required ${minimumRelayBalance}). Waiting for funding...`)
+        `Server's balance too low ( ${balance}, required ${defaultMinManagerBalance}). Waiting for funding...`)
```
forshtat, line 494: Why did you change it to 'default'? Doesn't it break the sole idea of having it be configurable?

src/relayserver/RelayServer.ts
```diff
@@ -644,17 +664,18 @@ export class RelayServer extends EventEmitter {
     return receipts
   }
 
-  async _isRegistered (): Promise<boolean> {
+  async _getRegistrationBlock (): Promise<number> {
```
forshtat, line 667: I was under impression the idea was for the server to send a re-register transaction in case no other transaction was made for a 'registrationBlockRate', so that the server does not become invisible. 
Here it will re-register even if active because it only checks the lates register event.
Is this on purpose?

src/relayserver/runServer.ts
```diff
@@ -68,6 +69,7 @@ const ethereumNodeUrl: string = argv.ethereumNodeUrl ?? config.ethereumNodeUrl ?
 const workdir: string = argv.workdir ?? config.workdir ?? error('missing --workdir')
 const devMode: boolean = argv.devMode ?? config.devMode ?? error('missing --devMode')
 const debug: boolean = argv.debug ?? config.debug ?? error('missing --debug')
+const registrationBlockRate: string = argv.registrationBlockRate ?? config.registrationBlockRate?.toString()
```
forshtat, line 72: `null?.toString()` will return 'undefined' and this should break the types system here...

src/relayserver/RelayServer.ts
shahafn: Fixed

src/relayserver/runServer.ts
shahafn: Wouldn't that crash?

src/relayserver/RelayServer.ts
shahafn: Actually I didn't change it, it was never configurable and I forgot to add it to configuration. Fixed.

src/relayserver/RelayServer.ts
```diff
@@ -949,4 +986,8 @@ export class RelayServer extends EventEmitter {
     }
     return 1000
   }
+
+  address2topic (address: string): string {
```
forshtat, line 990: We already have 'event2topic' in the Utils.ts
I believe it will be beneficial to keep these utility functions in the same utility file instead of inside the server

src/relayserver/runServer.ts
forshtat: And it will!..

test/RelayServer.test.ts
forshtat: I don't understand the question but I don't see an answer either 😀

src/relayserver/RelayServer.ts
forshtat: Nothing changed here

shahafn: This is true in general in the code, since it's not only me writing it. Consistency is important but should be fixed in another PR imo.

src/relayserver/RelayServer.ts
shahafn: Fixed

test/RelayServer.test.ts
shahafn: Fixed

forshtat:
Unresolved issues migrated to OG-174

PR #447, opened 2020-07-23
Og 121 remove 0x

drortirosh:
- use minimal LibBytes of readBytes32/readBytes4
- use "truncate" in-place (so we don't need complex "slice")
- GsnUtils.revertWithData: to revert with raw bytes data (taken from internal call revert()), instead of reason string.
- API: don't extract on-chain revert reason: return entire revert byte array (in TrasnactionRelayed and relayCall return value)
- added decodeRevertReason() in js


contracts/RelayHub.sol
```diff
@@ -302,7 +300,8 @@ contract RelayHub is IRelayHub {
             bytes memory retData;
             (success, retData) = relayRequest.relayData.paymaster.call{gas:gasLimits.preRelayedCallGasLimit}(vars.data);
             if (!success) {
-                revertWithStatus(RelayCallStatus.RejectedByPreRelayed, GsnEip712Library.getTruncatedData(retData));
+                GsnEip712Library.truncatedData(retData);
```
shahafn, line 303: Why is it here then?

contracts/utils/MinLibBytes.sol
```diff
@@ -0,0 +1,99 @@
...
+// a minimal subset from 0x/LibBytes
+/* solhint-disable no-inline-assembly */
+pragma solidity ^0.6.2;
+
+library MinLibBytes {
+
+    //truncate the given parameter if its length is above the given maximum length
+    // do nothing otherwise.
+    //NOTE: the truncation is done in-place
+    function truncatedData(bytes memory data, uint256 maxlen) internal pure {
```
shahafn, line 12: truncateInPlace maybe?

contracts/RelayHub.sol
drortirosh: yes, renamed.

contracts/utils/MinLibBytes.sol
drortirosh: yes, renamed.

PR #448, opened 2020-07-26
Do not use a field in a struct for a local variable

forshtat:


forshtat:
The failed test is just post overhead, will need to fix for all cheanges anyway

PR #450, opened 2020-07-27
OG135: Emit revert reason

drortirosh:
emit revert reason in a separate transaction (TransactionResult), not in TransactionRelayed

### DIFF
--- a/contracts/RelayHub.sol
+++ b/contracts/RelayHub.sol
@@ -206,7 +206,7 @@ contract RelayHub is IRelayHub {
         vars.innerGasUsed = vars.gasBeforeInner-gasleft();
         (vars.status, vars.relayedCallReturnValue) = abi.decode(relayCallStatus, (RelayCallStatus, bytes));
         if ( vars.relayedCallReturnValue.length>0 ) {
-            emit TransactionResult(vars.status, vars.relayedCallReturnValue );
+            emit TransactionResult(vars.status, vars.relayedCallReturnValue);
         }
     }
     {
--- a/contracts/RelayHub.sol
+++ b/contracts/RelayHub.sol
@@ -205,6 +205,9 @@ contract RelayHub is IRelayHub {
         vars.success = success;
         vars.innerGasUsed = vars.gasBeforeInner-gasleft();
         (vars.status, vars.relayedCallReturnValue) = abi.decode(relayCallStatus, (RelayCallStatus, bytes));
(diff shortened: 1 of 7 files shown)


contracts/RelayHub.sol
```diff
@@ -205,6 +205,9 @@ contract RelayHub is IRelayHub {
         vars.success = success;
         vars.innerGasUsed = vars.gasBeforeInner-gasleft();
         (vars.status, vars.relayedCallReturnValue) = abi.decode(relayCallStatus, (RelayCallStatus, bytes));
```
forshtat, line 207: Can it be a local variable now?

contracts/RelayHub.sol
```diff
@@ -205,6 +205,9 @@ contract RelayHub is IRelayHub {
...
+        if ( vars.relayedCallReturnValue.length>0 ) {
+            emit TransactionResult(vars.status, vars.relayedCallReturnValue );
+        }
```
forshtat, line 210: @drortirosh, please, get your code style together. I know solhint does not care about that, I know it is nitpicking, but these spaces! 
Use js standard when in doubt)

test/RelayHubGasCalculations.test.ts
```diff
@@ -280,7 +280,10 @@ contract('RelayHub gas calculations', function ([_, relayOwner, relayWorker, rel
               const beforeBalances = await getBalances()
               const pctRelayFee = requestedFee.toString()
               const senderNonce = (await forwarderInstance.getNonce(senderAddress)).toString()
-              const encodedFunction = recipient.contract.methods.emitMessage('a'.repeat(messageLength)).encodeABI()
+              let encodedFunction = recipient.contract.methods.emitMessage('a'.repeat(messageLength)).encodeABI()
+              if (messageLength === 0) {
+                encodedFunction = recipient.contract.methods.emitMessageNoParams().encodeABI()
```
forshtat, line 285: what's this?

contracts/RelayHub.sol
drortirosh: nope. adding a single stack variable  (even unused) before this line breaks it..

test/RelayHubGasCalculations.test.ts
drortirosh: this is a test that uses emitMEssageNoParams() instead of emitMessage("")
this way, there is no return value - and still, the gas calculation is correct.

PR #454, opened 2020-07-30
Use SafeMath for value-related calculations

forshtat:


contracts/RelayHub.sol
```diff
@@ -94,7 +95,7 @@ contract RelayHub is IRelayHub {
         uint256 amount = msg.value;
         require(amount <= maximumRecipientDeposit, "deposit too big");
 
-        balances[target] = SafeMath.add(balances[target], amount);
+        balances[target] = balances[target].add(amount);
```
drortirosh, line 98: not really needed just after a check that it is actual below maxDeposit...

contracts/RelayHub.sol
```diff
@@ -237,8 +238,8 @@ contract RelayHub is IRelayHub {
         // We've already checked that the paymaster has enough balance to pay for the relayed transaction, this is only
         // a sanity check to prevent overflows in case of bugs.
         require(balances[relayRequest.relayData.paymaster] >= charge, "Should not get here");
```
drortirosh, line 240: your safemath actually make this strange "require" line useless:

contracts/RelayHub.sol
```diff
@@ -362,7 +363,7 @@ contract RelayHub is IRelayHub {
     }
 
     function calculateCharge(uint256 gasUsed, GsnTypes.RelayData calldata relayData) public override virtual view returns (uint256) {
-        return relayData.baseRelayFee + (gasUsed * relayData.gasPrice * (100 + relayData.pctRelayFee)) / 100;
+        return relayData.baseRelayFee.add((gasUsed.mul(relayData.gasPrice).mul(100 + relayData.pctRelayFee)).div(100));
```
drortirosh, line 366: we're not protecting against any invalid values of the relayData. 
but instead of check we can limit them in range, as  none of these parameters require 256 bits.

pctRelayFee can safely use 32bit
baseRelayFee, gasPrice can fit <112bit
and thus they are protected against overflows, reduce bit-pollution (i.e reduce request size by 2 words)

src/common/Environments.ts
```diff
@@ -12,7 +12,7 @@ interface Environment {
 }
 
 const defaultRelayHubConfiguration: RelayHubConfiguration = {
-  gasOverhead: 34936,
+  gasOverhead: 37054,
```
drortirosh, line 15: the overhead of these tests is by far not neglible: its total of 2000

contracts/StakeManager.sol
```diff
@@ -39,7 +40,7 @@ contract StakeManager is IStakeManager {
         StakeInfo storage info = stakes[relayManager];
         require(info.owner == msg.sender, "not owner");
         require(info.withdrawBlock == 0, "already pending");
-        info.withdrawBlock = block.number + info.unstakeDelay;
+        info.withdrawBlock = block.number.add(info.unstakeDelay);
```
drortirosh, line 43: unstakeDelay is time (expressed in 15-second block time). it shouldn't be uint256, but uint128

contracts/RelayHub.sol
drortirosh: and even if not - they should be validated on registerRelay, not on each usage.

drortirosh:
missing test is validation of constructor parameters 
though the simplest way is use non-uint256 types (e.g. uint96 - since all these numbers represent "real world" numbers, not arbitrary

drortirosh:
one thing that is missing is constructor param validation. the simplest way is change number types to uint96, as all numbers are "real world" values, and none can/should use insane huge values (yes, it limits the max stake to 79 Billion ether...)

contracts/RelayHub.sol
forshtat: It is not about being needed

contracts/RelayHub.sol
forshtat: Should I remove this line?

contracts/RelayHub.sol
forshtat: A client can send a request with whatever value if the server agrees to it/the server can be the one manipulating with these values. Again, I don't think there is anything wrong with this code - but better safe then sorry

forshtat: And limiting uint size is something I really don't like as an idea

src/common/Environments.ts
forshtat: yeah, that is weird, we make like 5 checks, how is it so expensive?

contracts/StakeManager.sol
forshtat: Playing with int size is not security IMHO

contracts/RelayHub.sol
drortirosh: a. you left *two* unchecked values: `pctRelayFee` and `baseRelayFee` (remember: `add(a,b)` only verify that `a+b > b`, and make no assumption about a)
b. if you change it do `add(100)`, the expression bcms cmpltly uncpmrhnsibl -
   all for the sake of making it "correct".
c. So a better way is validating the parameters (e.g. `require(pctGasPrice<1e10 && baseRelayFee<1e10)`) - but I still think that using a bounded types are better.
d. side-note: `div(100)` is useless, since it only verifies that `100>0`

contracts/StakeManager.sol
drortirosh: a better option is `require( unstakeDelay<1e10)` - when adding it. if the number is known to be in valid range, no need to check it over again (both here and in "unauthorize")

contracts/RelayHub.sol
drortirosh: This is the wrong check, in the wrong place.
The method is called multiple times (any TokenPaymaster-style paymaster), so the params should be checked, 
liek other params, at the beginning of `relayCall()`.
We do check some params there, but neglect to check others.

should add a line there:

```
require(relayRequest.relayData.baseRelayFee<MAXINT && relayRequest.relayData.pctRelayFee < MAXINT, "Invalid fees");
```

(and outside: `uint constant MAXINT = uint96(-1);`)

note that no need to check other params of this function, since there is already check for `gasPrice` in relayCall, and `gasused` is also a bound value.

PR #456, opened 2020-08-02
OG-172 Add helper to read the GSN deployment addresses form the Paymaster

forshtat:


src/relayclient/GSNConfigurator.ts
```diff
@@ -33,6 +33,22 @@ const defaultGsnConfig: GSNConfig = {
   clientId: '1'
 }
 
+export async function resolveGSNDeploymentFromPaymaster (provider: provider, partialConfig: Partial<GSNConfig>): Promise<Partial<GSNConfig>> {
```
drortirosh, line 36: something like this, except that you want to call it automatically when there is no RelayHub configured

PR #457, opened 2020-08-03
OG-170 docker-based deployment for multiple servers on host

drortirosh:
specific relay instances are accessed as https://host/gsn1, https://host/gsn2
the relay-server instances are unmodified, and are unaware of the change (except that in their configuration they have the same  "gsn1" suffix)

moved all docker images to /dockers
* main image - relaydc
* inner relay server - jsrelay
* multi-server router - prefixrouter.

default sample configuration in relaydc/config-sample

relaydc - exposes "docker-compose" API. 
requires the "relaydc" alias 
Uses `gsndata`, `config` folders and `.env` file from current working dir


forshtat:
actionable:
rebase to master - then can be approved

dockers/relaydc/add-instance
```diff
@@ -0,0 +1,30 @@
+#!/usr/bin/perl
```
shahafn, line 1: @drortirosh this means you're the only one going to work with this file

drortirosh: It's a sample of how to add an entry to the dinner-compose file.
It duplicates the gsn1 entry, and rename it to the next number. 
It's not in use, and will never be used automatically. Only when we explicitly define a configuration with me server, which will required manual editing anyways.

drortirosh:
If it's that problematic I'll remove it. I can't think of writing it in any other language. Is pure text manipulation

PR #461, opened 2020-08-06
OG-122 Protect client fetching too many transactions in the "lookup window"

forshtat:
The current client logic for finding active relays:
look for all events in the "lookup window" (one month's worth of Transactions), and treat all relays found there as "active".
However, if there are too many transactions in that window, the client request may fail.

The mitigation for that should be to query the window in smaller chunks, so that a provider-imposed limit on # of returned events is not hit.
Note that the fix doesn't change the client's logic on the window size, and "active relay" definition is the same, regardless of network load.

### DIFF
--- a/src/relayclient/GSNConfigurator.ts
+++ b/src/relayclient/GSNConfigurator.ts
@@ -13,12 +13,13 @@ import { constants } from '../common/Constants'
 const GAS_PRICE_PERCENT = 20
 const MAX_RELAY_NONCE_GAP = 3
 const DEFAULT_RELAY_TIMEOUT_GRACE_SEC = 1800
+const BLOCKS_PER_MONTH = 30 * 24 * 60 * 6 // 259200, assuming 10 seconds block time
 
 const defaultGsnConfig: GSNConfig = {
   preferredRelays: [],
-  relayLookupWindowBlocks: 6000,
+  relayLookupWindowSliceSize: BLOCKS_PER_MONTH,
   gasPriceFactorPercent: GAS_PRICE_PERCENT,
-  lookupWindowsMaxCount: 3,
+  lookupWindowSliceCount: 1,
   minGasPrice: 0,
   maxRelayNonceGap: MAX_RELAY_NONCE_GAP,
   sliceSize: 3,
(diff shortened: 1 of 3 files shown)


drortirosh:
It is very unclear (from the user prespective) what the lookup window is.
why increase it gradually, instead of using bigger window to begin with?
if there is a technical problem with large window, maybe its better to configure a large window, and the code break it down to "chunks" to avoid node/API limitations (e.g. we already know that rpc can't return more than 10000 results, so we know that in the future, to handle such cases we would need to break the request down to ranges. maybe this is the case here too.)

drortirosh:
@forshtat - either update it, or close

drortirosh:
I see one problem with the solution I propose above:
our current implementation of `KnownRelaysManager` "forgot" all the information about a relay, and when pinging, all it has the URL. this is unfortunate, since the client does know the relayManagerAddress of each relay - it only that the data structure in KnownRelaysManager doesn't contain it.
So its only a matter of adding it back, and then the "hijacking" relay can be dropped during ping, without causing any issues.

drortirosh:
This is the fix to "hijacking" relayer: 
(we have all the info in **relayInfo**, even though at this place it is casted to a struct-with-url-only - since this place handles both preferred relays and normal ones, and the check applies only to non-preferred relays, which by definition, don't attempt to attack anyone)


```
 async _getRelayAddressPing (relayInfo: RelayInfoUrl): Promise<PartialRelayInfo> {
    if (this.config.verbose) {
      console.log(`getRelayAddressPing URL: ${relayInfo.relayUrl}`)
    }
    const pingResponse = await this.httpClient.getPingResponse(relayInfo.relayUrl)

    if (!pingResponse.Ready) {
      throw new Error(`Relay not ready ${JSON.stringify(pingResponse).slice(0,500)}`)
    }
//<<<<<
    const relayManager = (relayInfo as any).relayManager
    if ( relayManager!=null && pingResponse.RelayServerAddress.toLowerCase() !== relayManager.toLowerCase() ) {
      throw new Error( `Invalid relay configuration (wrong manager) ${JSON.stringify(pingResponse).slice(0, 500)}`)
    }
//>>>>>
    this.pingFilter(pingResponse, this.gsnTransactionDetails)
    return {
      pingResponse,
      relayInfo
    }
  }
```

drortirosh:
the title of this PR (and the jira issue) is broken - it still describes a possible solution, instead of the problem (=handle reject by a node if too many responses are returned)

PR #462, opened 2020-08-06
OG-63 Use 'relayManager' as a key to the mapping

forshtat:
Relays without known relay manager (preferred relays) are left with 0
score anyways.


drortirosh:
I don't really understand the issue, nor the fix.
The issue OG-63 was to prevent hijaking of URLs, and that can be done without any mapping or score changes. preferred relays don't need scoring: they are always used first, before any other relays, in the exact order they are defined.

The fix to the possible "hijacking" is:
when pinging a relay, the client knows the url and manager address. if the pinged relay return any other manager - simply ignore this response, as if it returned "ready:false": if relayA decide to register with urlB (of relayB), even if it is with lower gas fee, we ignore this response, since its not really relayB - we will ping relayB with its URL when we reach it in the list of relays, based on its relayManager address.

forshtat:
@drortirosh there is a misconception in your proposed solution:
Sorting is done before pinging, not after. 
This means that as a malicious actor, I can re-register (hijack) the URL of a _very good_ relay with a very low stake/high fee.
In this case, the client will not ping this _very good_ relay because it may use a wrong event to calculate its score.

I prevent this attack here by only keeping the scores by the manager address. This one is guaranteed to be unique.

For preferred relays I don't know it but I don't rank preferred relays anyways.

drortirosh:
1. client query events. get relayA and relayB (both use the same URL)
2. client sorts the relays - regardless of URL
3. client start pinging the relays. it drops any relay with invalid DNS, invalid response - and my suuggestion here, any relay with relayManager address not matching the expected relayManager (because "hijacker" must use its own relayManager address)

for this dropped entry, I don't care where it is placed in the sorted list. 
What is the problem here?

drortirosh:
This is the fix to "hijacking" relayer: 
(we have all the info in **relayInfo**, even though at this place it is casted to a struct-with-url-only - since this place handles both preferred relays and normal ones, and the check applies only to non-preferred relays, which by definition, don't attempt to attack anyone)


```
 async _getRelayAddressPing (relayInfo: RelayInfoUrl): Promise<PartialRelayInfo> {
    if (this.config.verbose) {
      console.log(`getRelayAddressPing URL: ${relayInfo.relayUrl}`)
    }
    const pingResponse = await this.httpClient.getPingResponse(relayInfo.relayUrl)

    if (!pingResponse.Ready) {
      throw new Error(`Relay not ready ${JSON.stringify(pingResponse).slice(0,500)}`)
    }
//<<<<<
    const relayManager = (relayInfo as any).relayManager
    if ( relayManager!=null && pingResponse.RelayServerAddress.toLowerCase() !== relayManager.toLowerCase() ) {
      throw new Error( `Invalid relay configuration (wrong manager) ${JSON.stringify(pingResponse).slice(0, 500)}`)
    }
//>>>>>
    this.pingFilter(pingResponse, this.gsnTransactionDetails)
    return {
      pingResponse,
      relayInfo
    }
  }
```

forshtat:
@drortirosh 
Only #2 is sufficient to fix the issue
There is no big advantage in also implementing #3
It will cause a client to drop a relay server that actually pinged, and the attack surface here is negligible in my opinion.

drortirosh:
see my suggested implementation above: a relay that returns invalid json, connection-refused and wrong relayhub should be treated the same.
as "#2" above - this is what it is today (sort is based on scoring, never on URLs), and there is no need to change it

drortirosh:
My bad.. you're right. I knew our scoring was based on server, NOT on URL, and I didn't understand why you wanted to change it.
but you actually DID fix it to be based on manager's address, which is of course correct.
Sorry for not reading your code change completely.
You're completely right for changing it to use relayManager address.


Still, dropping a server with broken relayManager during ping is better than return it as "success", and start a "ping" session again.

drortirosh:
@forshtat , there is one thing that misses in this PR: you're fixing a potential bug, but there is no test for it.
Can you add a test for it? 
(but if its too difficult, we can leave it as is, as it took too long..)

shahafn:
Added a check in an existing test that's supposed to fail on this line if it's changed, but that's not a full flow test of a relayer trying to shadow other relayer's url. I believe it's sufficient for now, though.

shahafn (approved):

PR #465, opened 2020-08-07
OG-168 VersionRegistry

drortirosh:


drortirosh:
An implementation of VersionOracle contract, as described in the https://docs.google.com/document/d/1f2xJSktgaPWwagT8UzUFTn7mdHbRdMfYccmtX_yIlqs/edit# document.

Note that it doesn't depend on events, and keep entire storage on chain.
its ~120k gas for a 32 byte "value" string, 250k for 200-byte string.
(event-based client-side implementation takes ~31kgas)

contracts/interfaces/IVersionOracle.sol
```diff
@@ -0,0 +1,54 @@
...
+    //struct defining a single version entry
+    struct VersionInfo {
+        uint64 time;
+        bool canceled;
+        bytes32 version;
+        string value;
+    }
+
+    //event emitted whenever a version is added
+    event VersionAdded(bytes32 indexed id, bytes32 version, string value);
```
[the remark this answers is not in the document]
drortirosh: value is string. client-side define its usage, not on-chain.
we will parse it as json, so there's room to whatever metadata we want.

contracts/interfaces/IVersionOracle.sol
```diff
@@ -0,0 +1,54 @@
...
+    /**
+     * return all (up to maxVersions) version entries from the registry for the given id, in reverse order
+     * (latest to earliest)
+     * @param id the id to get the version for.
+     * @param maxVersions the size of the returned array. up to that many entries can be returned.
+     * returns: count number of returned entries. if count==maxVersions, it is quite possible there are more versions,
+     *  so a new call should be attempted, with higher maxVersions.
+     *  ret array of VersionInfo
+     */
+    function getAllVersions(bytes32 id, uint maxVersions) external view returns (uint count, VersionInfo[] memory ret);
```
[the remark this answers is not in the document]
drortirosh: 1. its off-chain scanning. it takes ~1Mil gas to get 100 entries of 200 bytes.
2. we control the list. we keep the total number down
3. even with history, I think you can use the latest X versions. I actually added "getAllVersions" for completeness, not that we really need it (and we do have "last resort" event-based scanning.

contracts/utils/VersionOracle.sol
```diff
@@ -0,0 +1,65 @@
...
+        values[id][version].time = uint64(block.timestamp);
+        values[id][version].version = version;
+        values[id][version].value = value;
+        bytes32 lastVersion = idToVersion[id];
+        idToVersion[id] = version;
+        versionHistory[id][version] = lastVersion;
+        emit VersionAdded(id, version, value);
+    }
+
+    function cancelVersion(bytes32 id, bytes32 version) external override onlyOwner {
```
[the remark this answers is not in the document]
drortirosh: good idea. adding.

contracts/utils/VersionOracle.sol
```diff
@@ -0,0 +1,65 @@
...
+        values[id][version].canceled = true;
+        emit VersionCanceled(id, version);
+    }
+
+    function getAllVersions(bytes32 id, uint maxVersions) external override view returns (uint count, VersionInfo[] memory ret) {
+        ret = new VersionInfo[](maxVersions);
+        count = 0;
+        bytes32 ver = idToVersion[id];
+
+        while (ver != bytes32(0) && count < maxVersions) {
```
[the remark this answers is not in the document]
drortirosh: I wonder if it worth the effort. it is not that important (you can always use events)
latest events are the only one important (last "mature" version, and all pending "young" versions - but not much farther back.

.eslintrc.js
```diff
@@ -50,6 +50,8 @@ module.exports = {
           {
             allowNumber: true,
             allowBoolean: true,
+            allowAny: true,
+            allowNullish: true,
```
forshtat, line 54: At this point we can just disable this rule. I agree it is somewhat annoying, but it does very well prevent you from getting logs like `actual value: [Object object], expectedValue: undefined`

contracts/interfaces/IVersionOracle.sol
```diff
@@ -0,0 +1,54 @@
...
+pragma solidity ^0.6.2;
+pragma experimental ABIEncoderV2;
+
+interface IVersionOracle {
+
+    //struct defining a single version entry
+    struct VersionInfo {
+        uint64 time;
+        bool canceled;
+        bytes32 version;
```
forshtat, line 11: Meanings of 'canceled' and 'version' fields are not clear to me at this point. Could you please elaborate? Thanks.

contracts/interfaces/IVersionOracle.sol
```diff
@@ -0,0 +1,54 @@
...
+
+    /**
+     * return actual version to use.
+     * @param id the id to get the version for.
+     * @param delayPeriod - don't return entries added during the delayPeriod.
+     * @param optInVersion - override delayPeriod: return this entry even if it is within the delayPeriod (but not canceled)
+     * NOTE: in any case, "canceled" versions are never returned
+     * reverts if no entry can be returned (no versions, or all canceled or within the delayPeriod
+     */
+    function getVersion(bytes32 id, bytes32 optInVersion, uint delayPeriod) external view returns (VersionInfo memory);
```
forshtat, line 53: Why is there a need for a delay to be on-chain? How is it better than a client code making the timing decision here?

contracts/utils/VersionOracle.sol
```diff
@@ -0,0 +1,65 @@
...
+pragma experimental ABIEncoderV2;
+// solhint-disable not-rely-on-time
+
+import "../interfaces/IVersionOracle.sol";
+import "@openzeppelin/contracts/access/Ownable.sol";
+
+contract VersionOracle is IVersionOracle, Ownable {
+
+    mapping(bytes32 => bytes32) public idToVersion;
+    mapping(bytes32 => mapping(bytes32 => bytes32)) public versionHistory;
```
forshtat, line 12: Why do we need to store history at all?

contracts/utils/VersionOracle.sol
```diff
@@ -0,0 +1,65 @@
...
+pragma solidity ^0.6.2;
+pragma experimental ABIEncoderV2;
+// solhint-disable not-rely-on-time
+
+import "../interfaces/IVersionOracle.sol";
+import "@openzeppelin/contracts/access/Ownable.sol";
+
+contract VersionOracle is IVersionOracle, Ownable {
+
+    mapping(bytes32 => bytes32) public idToVersion;
```
forshtat, line 11: what?

contracts/utils/VersionOracle.sol
```diff
@@ -0,0 +1,65 @@
...
+    mapping(bytes32 => mapping(bytes32 => bytes32)) public versionHistory;
+    mapping(bytes32 => mapping(bytes32 => VersionInfo)) public values;
```
forshtat, line 13: values what? how is it a two-dimensional mapping?

drortirosh: for each ID we need to know the latest version (that's idToVersion)
we need a linked list of all versions of a given ID - that is versionHistory
values are the actual data (and metadata) for each version

.eslintrc.js
drortirosh: ok, so I'll remove the allowAny (so you should either toString() or JSON.stringify) - but adding "eslint-ignore" before almost every debug line is making the exception-case too common...

contracts/utils/VersionOracle.sol
drortirosh: removed on-chain logic (even though it was harder to implement and test in JS than it was in solidity...)

.eslintrc.js
forshtat: ok, so will you?

PR #468, opened 2020-08-11
OG-151 expose to relayer the paymaster's maxAcceptanceBudget

drortirosh:
add `paymasterMaxAcceptanceBudget` parameter to `RelayHub.relayCall()`
This is the highest acceptanceBudget the relayer agrees to pay for.
(the relayer can read it from the `paymaster.getGasLimits()` - but doesn't have to, and can just put the highest value it agrees to pay for.


contracts/RelayHub.sol
```diff
@@ -123,6 +124,9 @@ contract RelayHub is IRelayHub {
     returns (IPaymaster.GasLimits memory gasLimits, uint256 maxPossibleGas) {
         gasLimits =
             IPaymaster(relayRequest.relayData.paymaster).getGasLimits();
+
+        require(paymasterMaxAcceptanceBudget > gasLimits.acceptanceBudget, "unexpected high acceptanceBudget");
```
shahafn, line 128: Isn't it supposed to be `>=`? No test for that?

drortirosh: you're right. I should be >=

PR #471, opened 2020-08-22
trustedForwrader member should be public

drortirosh:
like all other contract member fields.


shahafn (approved):

forshtat (commented):
Isn't it a trojan horse for 'getTrustedForwarder' method that we were at war over once long time ago?

PR #476, opened 2020-08-26
Re-introduce re-registration test

forshtat:
Also:

1. Replace BlockHeader with BlockNumber as we are not really interested
in the rest of it.

2. Add chack that _worker is not provided with an older block

### DIFF
--- a/src/relayserver/RelayServer.ts
+++ b/src/relayserver/RelayServer.ts
@@ -1,6 +1,5 @@
 import log from 'loglevel'
 import ow from 'ow'
-import { BlockHeader } from 'web3-eth'
 import { EventData } from 'web3-eth-contract'
 import { EventEmitter } from 'events'
 import { PrefixedHexString } from 'ethereumjs-tx'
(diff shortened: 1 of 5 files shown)


drortirosh:
This is a refactor with no real benefit - other than temporarily reducing the transfer size (not request count) from infura.
blockHeader contains more relevant info than just blockNumber.
e.g. we use gasPrice - so its beter to replace getGasPrice() with blockHeader.gasPrice.

forshtat:
> This is a refactor with no real benefit - other than temporarily reducing the transfer size (not request count) from infura.
> blockHeader contains more relevant info than just blockNumber.
> e.g. we use gasPrice - so its beter to replace getGasPrice() with blockHeader.gasPrice.

The purpose of this PR is to re-introduce the two re-registration tests. I did omit one that I found a little too repetitive - it is better to be tested with unit tests. 

The 'BlockHeader' object is not currently used at all. I did not change the RPC request we are making, but this is now an option.
I removed it because a) it comes from web3 b) it bears a lot of data that can be used but should not

In your example, you want to use the gas price from the last block instead of the 'getGasPrice' RPC call.
May I say this is in no way equivalent as nodes are expected to give a weighted average of a gas price of the latest X blocks as a gas price, and block gas price fluctuates a lot.

forshtat:
https://opengsn.atlassian.net/browse/OG-205

shahafn (approved):

PR #477, opened 2020-08-27
OG-202 RelayProvider doesn't handle exceptions in callbacks

drortirosh:


test/relayclient/RelayClient.test.ts
```diff
@@ -156,6 +156,44 @@ contract('RelayClient', function (accounts) {
...
+          asyncPaymasterData: async () => { throw new Error('paymasterData-error') }
+        })
+      const { transaction, relayingErrors, pingErrors } = await relayClient.relayTransaction(options)
+      assert.isUndefined(transaction)
+      assert.equal(pingErrors.size, 0)
+      assert.equal(relayingErrors.size, 1)
+      assert.match(relayingErrors.values().next().value.message, /paymasterData-error/)
+    })
+
+    it.skip('should return errors in callback (scoreCalculator) ', async function () {
```
forshtat, line 184: is this test added skipped on purpose?

test/Flows.test.ts
```diff
@@ -155,29 +154,33 @@ options.forEach(params => {
...
-          const relayProvider =
+        const setRecipientProvider = function (asyncApprovalData: AsyncDataCallback): void {
+          relayProvider =
             // @ts-ignore
             new RelayProvider(web3.currentProvider,
-              relayClientConfig, { asyncApprovalData: async () => await Promise.resolve(approvalData) })
+              relayClientConfig, { asyncApprovalData })
           TestRecipient.web3.setProvider(relayProvider)
-        })
+        }
```
forshtat, line 173: This is a weird construct tbh

forshtat (approved):

test/Flows.test.ts
drortirosh: I wanted a local function to set the provider with "custom" param.
it didn't let me define a function in this scope, only variables - and it does have side-effect of modifying "describe"-context variable, so it can't be moved out

test/relayclient/RelayClient.test.ts
drortirosh: yes: I added it for completeness, but it fails, since unused function doesn't throw....
we forgot to propagate the scoring function to the RelaysManager
(its a separate issue, and I think we also need to modify the scoring function: it should have a "next" param, so you can "chain" the current one with extra logic, otherwise you need to re-implment it entirely from scratch (e.g. handling failures)

PR #479, opened 2020-08-27
OG-192 Contract Interactor should cache contract instances

forshtat:


drortirosh:
getCode was never the major issue. why do you stress on working on it?
also: caching Recipient is useless: good if you have ONE recipient. the cache is invalidated the 2nd recipient is accessing the server. if you want, you can use cache-by-address with LRU  (e.g. map-cache) - but again, is looking under the lantern: this is not the major issue.

forshtat:
> getCode was never the major issue. why do you stress on working on it?
> also: caching Recipient is useless: good if you have ONE recipient. the cache is invalidated the 2nd recipient is accessing the server. if you want, you can use cache-by-address with LRU (e.g. map-cache) - but again, is looking under the lantern: this is not the major issue.

1. This is a very low hanging win. We've had this issue for quite some time and it's trivial to fix.
2. Every truffle-contract creation takes 3 RPC calls -net_version, getBlock and getCode. This is wasteful.
3. I believe in vast majority of cases there will be a single recipient (for clients), and a ton of recipients for servers. So I choose to cache only the last one for now.

drortirosh:
> 3. I believe in vast majority of cases there will be a single recipient (for clients), and a ton of recipients for servers. So I choose to cache only the last one for now.

we don't really care about client network access at this point.
for server, to cache the last recipient is great as long as there is one dapp.
once there's a 2nd dapp, the cache becomes useless.

Seems like a better approach would be to have a provider that caches the results of getCode, getVersion, and thus the code stays the same...

a bit larger change would be to completely remove the TruffleContract, as they bring very little good (a bit simpler syntax of call method) - at a very high cost (both in storage and network access)

PR #480, opened 2020-08-29
OG-196: Decrease rpc calls for worker

forshtat:


src/relayserver/RegistrationManager.ts
```diff
@@ -124,19 +135,38 @@ export class RegistrationManager {
...
+    if (lastRegisteredTxSinceLastScan != null) {
+      this.lastMinedRegisterTransaction = lastRegisteredTxSinceLastScan
+    }
+    if (this.lastMinedRegisterTransaction == null) {
+      this.lastMinedRegisterTransaction = await this._findOlderRegistrationEvent()
+    }
+    return isRegistrationValid(this.lastMinedRegisterTransaction, this.config, this.managerAddress)
+  }
+
+  async _findOlderRegistrationEvent (): Promise<EventData | undefined> {
```
drortirosh, line 159: is it older ? you're using reduceToLatest, which is the youngest event..

src/relayserver/RelayServer.ts
```diff
@@ -440,50 +439,53 @@ export class RelayServer extends EventEmitter {
...
-    }
+    return await this.contractInteractor.getPastEventsForHub(topics, options)
   }
 
-  async _handleTransactionRejectedByPaymasterEvent (dlog: EventData, blockNumber: number): Promise<void> {
-    log.debug('handle TransactionRejectedByPaymaster event', dlog)
+  async _handleTransactionRejectedByPaymasterEvent (blockNumber: number): Promise<void> {
     this.alerted = true
     this.alertedBlock = blockNumber
     console.error(`Relay entered alerted state. Block number: ${blockNumber}`)
```
drortirosh, line 470: you removed the reason for being alerted (=the paymaster/tx) even in debug.

src/relayserver/RelayServer.ts
```diff
@@ -440,50 +439,53 @@ export class RelayServer extends EventEmitter {
...
-        (e: EventData) =>
-          e.blockNumber)
-      .reduce(
-        (b1: any, b2: any) => Math.max(b1, b2), 0)
+  async _getLatestTxBlockNumber (eventsSinceLastScan: EventData[]): Promise<number> {
+    const latestTransactionSinceLastScan = reduceToLatestTx(eventsSinceLastScan)
+    if (latestTransactionSinceLastScan != null) {
+      this.lastMinedActiveTransaction = latestTransactionSinceLastScan
+    }
+    if (this.lastMinedActiveTransaction == null) {
```
drortirosh, line 478: instead of checking each time if it is initialized, this is something that should be initialized once, during _init

src/relayserver/RelayServer.ts
```diff
@@ -436,50 +439,53 @@ export class RelayServer extends EventEmitter {
     return toBN(await this.contractInteractor.getBalance(this.workerAddress))
   }
 
-  async getShouldRegisterAgain (): Promise<boolean> {
-    const currentBlock = await this.contractInteractor.getBlockNumber()
-    const latestTxBlockNumber = await this._getLatestTxBlockNumber()
+  async getShouldRegisterAgain (currentBlock: number, hubEventsSinceLastScan: EventData[]): Promise<boolean> {
```
shahafn, line 442: Please rename to just `shouldRegisterAgain()`

src/relayserver/RelayServer.ts
```diff
@@ -436,50 +439,53 @@ export class RelayServer extends EventEmitter {
...
+    if (latestTransactionSinceLastScan != null) {
+      this.lastMinedActiveTransaction = latestTransactionSinceLastScan
+    }
+    if (this.lastMinedActiveTransaction == null) {
+      this.lastMinedActiveTransaction = await this._findOlderActiveEvent()
+    }
+    return this.lastMinedActiveTransaction?.blockNumber ?? -1
+  }
+
+  async _findOlderActiveEvent (): Promise<EventData | undefined> {
```
shahafn, line 484: isn't that `__findLatestActiveEvent()`?

src/relayserver/RegistrationManager.ts
forshtat: renamed to '_queryLatestXX'

src/relayserver/RelayServer.ts
forshtat: logged inside the 'switch/case'. It was passed in for no other purpose then being logged.

src/relayserver/RelayServer.ts
forshtat: Fixed

src/relayserver/RelayServer.ts
forshtat: renamed to '_queryLatestXX'

src/relayserver/RelayServer.ts
forshtat: Note that there may be none if the server just started. This would be an awkward special case if I had to initialise `RelayServer` with the `lastMinedActiveTransaction` already not-null.

shahafn (approved):

PR #481, opened 2020-08-31
OG-198 Penalizer contract fix

shahafn:


contracts/Penalizer.sol
```diff
@@ -26,6 +26,10 @@ contract Penalizer is IPenalizer{
 
     }
 
+    function relayManagerOnly(IRelayHub hub) private view {
```
forshtat, line 29: Call it 'require' so it is obvious this method reverts if not manager. 'xxOnly' is more of a modifier name.

test/RelayHubPenalizations.test.ts
```diff
@@ -158,6 +165,30 @@ contract('RelayHub Penalizations', function ([_, relayOwner, relayWorker, otherR
       return receipt
     }
 
+    describe('relay manager only can penalize', function () {
```
forshtat, line 168: describe - method/behaviour
it - should do bla bla
please

contracts/Penalizer.sol
shahafn: Converted to modifier then

forshtat (approved):

PR #484, opened 2020-09-03
Use block count instead of timestamp as pending state timeout units

forshtat:
Date/time is an extremely unreliable measurement in terms of blockchain
and is hard to test for.

This PR makes TransactionManager remember the block height at the point
of the transaction creation.

Also:
1. Create StoredTransaction in place of StoredTx and StoredParams
2. Add 'pendingTransactionTimeoutBlocks', 'confirmationsNeeded',
'retryGasPriceFactor' and 'maxGasPrice' to the Server Config struct
3. Separate database cleanubp and gas price boost logic & test
4. 'resend multiple unconfirmed transactions' skipped and must be redone

### DIFF
--- a/src/relayserver/RegistrationManager.ts
+++ b/src/relayserver/RegistrationManager.ts
@@ -13,7 +13,7 @@ import ContractInteractor, {
   StakeAdded,
   StakeUnlocked
 } from '../relayclient/ContractInteractor'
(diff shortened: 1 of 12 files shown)


forshtat:
https://opengsn.atlassian.net/browse/OG-221

PR #486, opened 2020-09-05
OG-210 Forwarder add registerDomainSeparator

drortirosh:
Without it, requests from one forwarder can be replayed through another forwarder (on the same network, or on another network)
By registering the domainSeparator, the forwarder can verify it appears in it as the verifyingContract, and that the chainId is valid.

### DIFF
--- a/contracts/forwarder/Forwarder.sol
+++ b/contracts/forwarder/Forwarder.sol
@@ -10,8 +10,7 @@ contract Forwarder is IForwarder {
 
     string public constant GENERIC_PARAMS = "address from,address to,uint256 value,uint256 gas,uint256 nonce,bytes data";
 
-    string public constant EIP712_DOMAIN_PREFIX = "EIP712Domain(";
-    string public constant EIP712_DOMAIN_SUFFIX = ",uint256 chainId,address verifyingContract)";
+    string public constant EIP712_DOMAIN_TYPE = "EIP712Domain(string name,string version,uint256 chainId,address verifyingContract)";
 
     mapping(bytes32 => bool) public typeHashes;
     mapping(bytes32 => bool) public domains;
@@ -89,22 +88,22 @@ contract Forwarder is IForwarder {
         registerRequestTypeInternal(requestType);
     }
 
-    function registerDomainSeparator(string calldata typePrefix, bytes memory dataPrefix) external override {
-        bytes memory domainType = abi.encodePacked(EIP712_DOMAIN_PREFIX, typePrefix, EIP712_DOMAIN_SUFFIX );
+    function registerDomainSeparator(string calldata name, string calldata version) external override {
         uint256 chainId;
         /* solhint-disable-next-line no-inline-assembly */
         assembly { chainId := chainid() }
 
-        bytes memory domainValue = abi.encodePacked(
-            keccak256(domainType),
-            dataPrefix,
+        bytes memory domainValue = abi.encode(
+            keccak256(bytes(EIP712_DOMAIN_TYPE)),
+            keccak256(bytes(name)),
+            keccak256(bytes(version)),
             chainId,
-            uint256(uint160(address(this))));
+            address(this));
 
         bytes32 domainHash = keccak256(domainValue);
 
         domains[domainHash] = true;
-        emit DomainRegistered(domainHash, string(domainType), domainValue);
+        emit DomainRegistered(domainHash, domainValue);
     }
 
     function registerRequestTypeInternal(string memory requestType) internal {
@@ -114,7 +113,7 @@ contract Forwarder is IForwarder {
         emit RequestTypeRegistered(requestTypehash, requestType);
     }
 
-    event DomainRegistered(bytes32 indexed domainSeparator, string domainType, bytes domainValue);
+    event DomainRegistered(bytes32 indexed domainSeparator, bytes domainValue);
 
     event RequestTypeRegistered(bytes32 indexed typeHash, string typeStr);
 
--- a/contracts/forwarder/Forwarder.sol
+++ b/contracts/forwarder/Forwarder.sol
@@ -128,8 +128,8 @@ contract Forwarder is IForwarder {
     internal
     view
     {
-        require(domains[domainSeparator], "invalid domain separator");
-        require(typeHashes[requestTypeHash], "invalid request typehash");
+        require(domains[domainSeparator], "unregistered domain separator");
+        require(typeHashes[requestTypeHash], "unregistered request typehash");
         bytes32 digest = keccak256(abi.encodePacked(
                 "\x19\x01", domainSeparator,
                 keccak256(_getEncoded(req, requestTypeHash, suffixData))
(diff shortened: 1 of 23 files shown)


contracts/forwarder/Forwarder.sol
```diff
@@ -85,13 +89,32 @@ contract Forwarder is IForwarder {
...
+        /* solhint-disable-next-line no-inline-assembly */
+        assembly { chainId := chainid() }
+
+        bytes memory domainValue = abi.encodePacked(
+            keccak256(domainType),
+            dataPrefix,
+            chainId,
+            uint256(uint160(address(this))));
+
+        bytes32 domainHash = keccak256(domainValue);
```
forshtat, line 104: `domainSeparatorHash` was defined as EIP-712 hash of `EIP712Domain` structure. 
Here `domainHash` means a hash of type (as string) and some more data thrown in. 
I am not sure this is needed.

forshtat (changes requested):

src/common/EIP712/TypedRequestData.ts
```diff
@@ -43,10 +46,24 @@ interface Types extends EIP712Types {
...
 }
 
+// use these values in registerDomainSeparator
+export const GsnDomainSeparatorType = {
+  prefix: 'string name,string version',
+  name: 'GSN Relayed Transaction',
+  version: '2'
+}
+
+export function getRegisterDomainSeparatorData (): {prefix: string, dataPrefix: string } {
```
forshtat, line 56: this is static data, why does it need a getter function?

src/common/EIP712/TypedRequestData.ts
```diff
@@ -43,10 +46,24 @@ interface Types extends EIP712Types {
...
+  return {
+    prefix: GsnDomainSeparatorType.prefix,
+    dataPrefix: keccak256(GsnDomainSeparatorType.name) + keccak256(GsnDomainSeparatorType.version).slice(2)
```
forshtat, line 59: 1) keccak returns a string with 0x, so here will be 0xaaa0xbbbb - how does it even work? decoeded as pure string?
2) why make version, name pre-hashed ?

src/common/EIP712/TypedRequestData.ts
```diff
@@ -43,10 +46,24 @@ interface Types extends EIP712Types {
...
+  }
+}
+
 export function getDomainSeparator (verifier: Address, chainId: number): EIP712Domain {
   return {
-    name: 'GSN Relayed Transaction',
-    version: '2',
+    name: GsnDomainSeparatorType.name,
+    version: GsnDomainSeparatorType.version,
```
forshtat, line 66: why?

src/common/EIP712/ForwarderUtil.ts
```diff
@@ -0,0 +1,24 @@
...
+  }
+
+  await forwarder.methods.registerRequestType(
+    GsnRequestType.typeName,
+    GsnRequestType.typeSuffix
+  ).send(sendOptions)
+
+  const { prefix, dataPrefix } = getRegisterDomainSeparatorData()
+  await forwarder.methods.registerDomainSeparator(prefix, dataPrefix).send(sendOptions)
+}
```
forshtat, line 24: **NO**

forshtat (changes requested):

contracts/forwarder/Forwarder.sol
drortirosh: see after the change.

contracts/forwarder/Forwarder.sol
```diff
@@ -85,13 +88,32 @@ contract Forwarder is IForwarder {
...
+        assembly { chainId := chainid() }
+
+        bytes memory domainValue = abi.encode(
+            keccak256(bytes(EIP712_DOMAIN_TYPE)),
+            keccak256(bytes(name)),
+            keccak256(bytes(version)),
+            chainId,
+            address(this));
+
+        bytes32 domainHash = keccak256(domainValue);
```
forshtat, line 103: There is GsnEip712Library::hashDomain, I suggest we reuse code to avoid duplication.

src/common/EIP712/ForwarderUtil.ts
```diff
@@ -0,0 +1,23 @@
+import { GsnDomainSeparatorType, GsnRequestType } from './TypedRequestData'
+import { IForwarderInstance } from '../../../types/truffle-contracts'
+import { Contract } from 'web3-eth-contract'
+
+// register a forwarder for use with GSN: the request-type and domain separator we're using.
+export async function registerForwarderForGsn (forwarderTruffleOrWeb3: IForwarderInstance|Contract, sendOptions: any = undefined): Promise<void> {
```
forshtat, line 6: Please pick one or the other.  Or create an override for both. Please do not create a single method for **two** different kinds of contract object instance. This is what 'no'.

src/common/EIP712/TypedRequestData.ts
```diff
@@ -5,6 +5,8 @@ import { EIP712Domain, EIP712TypedData, EIP712TypeProperty, EIP712Types, TypedDa
 import { bufferToHex } from 'ethereumjs-util'
 import { PrefixedHexString } from 'ethereumjs-tx'
 
+require('source-map-support').install({ errorFormatterForce: true })
```
forshtat, line 8: Does it work?

test/Forwarder.test.ts
```diff
@@ -81,6 +89,34 @@ contract('Forwarder', ([from]) => {
     })
   })
 
+  describe('#registerDomainSeparator', () => {
+    it('registered domain should match local definition', async () => {
```
forshtat, line 93: I find "it should " a better test name pattern as it describes the desired behaviour, not the outcome.
Honestly, this sentence just re-states what the asserts check, not what the method is doing.

drortirosh: ok, suggest something else.

src/common/EIP712/TypedRequestData.ts
drortirosh: don't remember... added it in several places, and at least in some it helped. we should have an "import" done everywhere (so its "import", and not "require"), and if we find a better fix, we'll put it there, in one place.

contracts/forwarder/Forwarder.sol
drortirosh: the forwarder can't use "Gsn" methods - its a standalone contract.

src/common/EIP712/ForwarderUtil.ts
drortirosh: why? yes, idealy, I'd have an overload with different types - but there are no overloads in javascript/typescript.

we moved to TruffleContract, but deployment still uses web3 contract, hence I need two flavours.
I think both should be removed in favor of "ethers", which is has API simpler to use than both, and more efficient resource-wise.

so I can either:
- add `registerForwarderForGsnWithTruffle` method for tests
- use `registerForwarderForGsn(forwarderContract.contract)` in all tests
- create a wrapper `registerForwarderForGsn` function, just for tests

do you think any of these is better?

test/Forwarder.test.ts
drortirosh: Its a sanity check, that on-chain calculation matches host (local) calculation.

drortirosh: at compares actual values that are hashed into the domain-separator, so its not has "hash is different" but actual data that gets into this hash

forshtat (approved):

PR #490, opened 2020-09-10
OG-200 client events

drortirosh:
added to client (and provider)
  `registerEventListener(handler: (event:GsnEvent)=>void)`


src/relayclient/RelayClient.ts
```diff
@@ -108,16 +139,18 @@ export class RelayClient {
   }
 
   async _init (): Promise<void> {
+    if (this.initialized) { return }
+    this.emit(new GsnInitEvent())
```
forshtat, line 143: at least do emit after the asynchronous operation

src/relayclient/RelayClient.ts
```diff
@@ -108,16 +139,18 @@ export class RelayClient {
...
 
   async relayTransaction (gsnTransactionDetails: GsnTransactionDetails): Promise<RelayingResult> {
-    if (!this.initialized) {
-      await this._init()
-    }
+    await this._init()
     // TODO: should have a better strategy to decide how often to refresh known relays
+    this.emit(new GsnRefreshRelaysEvent())
     await this.knownRelaysManager.refresh()
+    this.emit(new GsnRefreshedRelaysEvent((await this.knownRelaysManager.getRelaysSortedForTransaction({} as any)).length))
```
forshtat, line 153: WTF?

src/relayclient/GsnEvents.ts
```diff
@@ -0,0 +1,45 @@
...
+  total = TOTAL_EVENTS
+  constructor (readonly event: string, readonly step: number) {}
+}
+
+// initialize client (takes time only on first request)
+export class GsnInitEvent extends GsnEvent {
+  constructor () { super('init', 1) }
+}
+
+export class GsnRefreshRelaysEvent extends GsnEvent {
```
forshtat, line 17: 'refresh' and 'refreshed' are indistinguishable. "FinishedRefreshing" or something....

src/relayclient/RelayClient.ts
forshtat: I understand the API did not suit your use-case so you abused the API with a cast of empty object to any.
I assume this is a good place to make a change in the API.

forshtat (commented):
As I've said repeatedly, I don't think there is a place for the EventEmitter pattern in this world.

src/relayclient/GsnEvents.ts
drortirosh: RefreshDone?

src/relayclient/RelayClient.ts
drortirosh: I do the events before the action (with the exception of "refresh", where the "done" event carries some info) . the idea that the event title means we're in this state.
anyway, I think this event will go away, since the "init" should be ran always after construction, not before first relayTransaction

src/relayclient/RelayClient.ts
drortirosh: tried to limit the pr to events only. but yes, API change is needed here, to expose the list of relays.

forshtat: Note that `knownRelaysManager.knownRelays` seems to be accessible. Is it not what you are looking for? I am surprised this does not crash. It will crash one day.

src/relayclient/GsnEvents.ts
forshtat: DoneRefreshing

forshtat (approved):

PR #491, opened 2020-09-15
OG-232 Relayer withdraws too early on HubUnauthorized.

drortirosh:
1. relayer should wait the unstakeDelay (as specified in the event before withdraw)
2. only withdraw worker balances on HubUnauthorized, not manager's stake or deposit (as it is shared with other relayer instances)
3. delay withdrawals also for StakeUnlocked event.

### DIFF
--- a/src/relayserver/RegistrationManager.ts
+++ b/src/relayserver/RegistrationManager.ts
@@ -211,14 +211,14 @@ export class RegistrationManager {
 
   /**
    * @param withdrawManager - whether to send the relay manager's balance to the owner.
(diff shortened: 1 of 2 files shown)


forshtat:
I do not understand what you did here.

The problems were:
1. There may be multiple hubs. Event 'HubUnauthorized' says what hub was unauthorized. This value is not checked by code.
2. The 'HubUnauthorized' does not happen immediately. This means workers who send there balances immediately make themselves penalizable.

Please explain what was fixed in this PR, and how does it relate to the problems highlighted here.

PR #492, opened 2020-09-15
OG-187 withdraw too early

drortirosh:
relayer should withdraw funds after only after unstakeDelay
* HubUnauthorized - should wait until the removalBlock event param
* StakeUnlocked - should wait until the withdrawBlock event param

withdrawing any funds from worker before that makes the relayer penalizeable...

### DIFF
--- a/src/relayserver/RegistrationManager.ts
+++ b/src/relayserver/RegistrationManager.ts
@@ -193,9 +193,10 @@ export class RegistrationManager {
   }
 
   async _handleHubUnauthorizedEvent (dlog: EventData, currentBlock: number): Promise<TransactionReceipt[]> {
-    if (dlog.returnValues.relayHub.toLowerCase() === this.hubAddress.toLowerCase()) {
-      this.isHubAuthorized = false
+    if (dlog.returnValues.relayHub.toLowerCase() !== this.hubAddress.toLowerCase()) {
+      return []
     }
+    this.isHubAuthorized = false
     return await this.withdrawAllFunds(false, currentBlock)
   }
 
@@ -211,7 +212,7 @@ export class RegistrationManager {
 
   /**
    * @param withdrawManager - whether to send the relay manager's balance to the owner.
-   *        Note that more then one relay process could be using the same manager account.
+   *        Note that more than one relay process could be using the same manager account.
    * @param currentBlock
    */
   async withdrawAllFunds (withdrawManager: boolean, currentBlock: number): Promise<TransactionReceipt[]> {
--- a/src/relayserver/RegistrationManager.ts
+++ b/src/relayserver/RegistrationManager.ts
@@ -11,7 +11,7 @@ import ContractInteractor, {
   RelayServerRegistered,
   RelayWorkersAdded,
   StakeAdded,
-  StakeUnlocked
+  StakeUnlocked, StakeWithdrawn
 } from '../relayclient/ContractInteractor'
 import { SendTransactionDetails, TransactionManager } from './TransactionManager'
 import { defaultEnvironment } from '../common/Environments'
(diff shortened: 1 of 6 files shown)


src/relayserver/RegistrationManager.ts
```diff
@@ -126,15 +127,29 @@ export class RegistrationManager {
...
     }
+
+    // handle HubUnauthorized, StakeUnlocked only after their due time
+
+    for (const eventData of this._extractDuePendingEvents(currentBlock)) {
+      switch (eventData.event) {
+        case 'HubUnauthorized':
+          receipts = receipts.concat(await this._handleHubUnauthorizedEvent(eventData, currentBlock))
+          break
+        case 'StakeUnlocked':
```
forshtat, line 147: Please note that there is a 'StakeWithdrawn' event.
It seems to be way more reasonable to react to the stake being withdrawn.

Also, **pay attention** to the fact that 'penalizeRelayManager' does **not check**  if the stake was unlocked or not. It can be added but seems redundant as it is in the best interests of the relay operator to withdraw ASAP, especially if it might have committed a penalizeable action.

src/relayserver/RegistrationManager.ts
```diff
@@ -145,6 +160,12 @@ export class RegistrationManager {
     }
   }
 
+  _extractDuePendingEvents (currentBlock: number): EventData[] {
+    const ret = this.delayedEvents.filter(event => event.block <= currentBlock).map(e => e.eventData)
+    this.delayedEvents = [...this.delayedEvents.filter(event => event.block > currentBlock)]
```
forshtat, line 165: Why do you need to use `a = [...x]` instead of `a = x` here? 

Also, this method is getter with side effects. I don't like it. "Extract"..?

test/relayserver/RegistrationManager.test.ts
```diff
@@ -312,7 +318,20 @@ contract('RegistrationManager', function (accounts) {
         await revert(id)
       })
 
-      it('send only manager hub balance and workers\' balances to owner (not manager eth balance)', async function () {
+      it('should not send balance immediately after unauthorize (before unstake delay)', async function () {
```
forshtat, line 321: Negative tests barely test anything useful. Here you check the old behaviour is not there, but this is not necessary IMO.

test/relayserver/RegistrationManager.test.ts
```diff
@@ -323,40 +342,67 @@ contract('RegistrationManager', function (accounts) {
...
+        { block: 1, eventData: 'event1' },
+        { block: 2, eventData: 'event2' },
+        { block: 3, eventData: 'event3' }
+      ]
+      extracted = rm._extractDuePendingEvents(2) as any
+    })
+    it('should extract events which are due (lower or equal block number)', function () {
+      assert.deepEqual(extracted, ['event1', 'event2'])
+    })
+    it('should should future events in the delayedEvents list', function () {
```
forshtat, line 394: should should

forshtat (changes requested):

src/relayserver/RegistrationManager.ts
drortirosh: its an atomic operation: extract some, leave the rest. I deliberately don't want to return some, and leave a chance they will be returned again - hence its "extract" and not a "get"

test/relayserver/RegistrationManager.test.ts
drortirosh: you're right. its only here because there's a positive test just next to it (I could do multiple asserts in the same test - would it be better? it will make a huge, unreadable single test method.

test/relayserver/RegistrationManager.test.ts
drortirosh: I should should proof proof read read my my texts texts...

src/relayserver/RegistrationManager.ts
forshtat: You did not answer the first question.

Also, I don't agree with the answer to the second one.

drortirosh: I don't trust "filter" to return a "stable" array. since I keep the result in memory, I want to make sure it is indeed an array, hence the "[...]" wrapper.

I don't want to have 2 methods, one is "getDueItems" and the other "removeDueItems", as its an atomic operation.

forshtat (approved):

PR #493, opened 2020-09-15
OG-219 Make RelayServer not wait for the transaction to be mined

forshtat:
As there is no way to broadcast a transaction with Web3, implement the
low-level RPC call directly in the ContractInteractor.

Also:
1. Introduce Web3Provider type definition
2. Extract common 'WrapperProviderBase' abstract class
3. Make 'pollNonce' always return locally known nonce

### DIFF
--- a/src/relayclient/ContractInteractor.ts
+++ b/src/relayclient/ContractInteractor.ts
@@ -1,5 +1,13 @@
 import Web3 from 'web3'
-import { BlockNumber, provider, Transaction, TransactionReceipt } from 'web3-core'
+import {
+  BlockNumber,
+  HttpProvider,
+  IpcProvider,
+  provider,
+  Transaction,
+  TransactionReceipt,
+  WebsocketProvider
+} from 'web3-core'
 import { EventData, PastEventOptions } from 'web3-eth-contract'
 import { PrefixedHexString, TransactionOptions } from 'ethereumjs-tx'
 
@@ -53,6 +61,11 @@ export const StakeAdded: EventName = 'StakeAdded'
 export const StakeUnlocked: EventName = 'StakeUnlocked'
 export const StakePenalized: EventName = 'StakePenalized'
 
+export type Web3Provider =
+  | HttpProvider
+  | IpcProvider
+  | WebsocketProvider
+
 export default class ContractInteractor {
   private readonly VERSION = '2.0.0-beta.2'
 
@@ -70,8 +83,8 @@ export default class ContractInteractor {
   private relayRecipientInstance?: BaseRelayRecipientInstance
   private knowForwarderAddressInstance?: IKnowForwarderAddressInstance
 
-  private readonly web3: Web3
-  private readonly provider: provider
+  readonly web3: Web3
+  private readonly provider: Web3Provider
   private readonly config: GSNConfig
   private readonly versionManager: VersionsManager
 
@@ -80,7 +93,7 @@ export default class ContractInteractor {
   private networkId?: number
   private networkType?: string
 
-  constructor (provider: provider, config: GSNConfig) {
+  constructor (provider: Web3Provider, config: GSNConfig) {
     this.versionManager = new VersionsManager(this.VERSION)
     this.web3 = new Web3(provider)
     this.config = config
@@ -124,8 +137,6 @@ export default class ContractInteractor {
   }
 
   getProvider (): provider { return this.provider }
-
-  getWeb3 (): Web3 { return this.web3 }
 
   async init (): Promise<void> {
     if (this.rawTxOptions != null) {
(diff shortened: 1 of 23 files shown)


src/relayclient/ContractInteractor.ts
```diff
@@ -421,6 +432,36 @@ export default class ContractInteractor {
...
+        method: 'eth_sendRawTransaction',
+        params: [
+          signedTransaction
+        ],
+        id: Date.now()
+      }, (e: Error | null, r: any) => {
+        if (e != null) {
+          reject(e)
+        } else if (r.error != null) {
+          reject(r.error)
```
drortirosh, line 458: please verify whether we need to `reject(error)` or `reject( new Error(error))`

src/relayserver/TransactionManager.ts
```diff
@@ -92,14 +94,13 @@ export class TransactionManager {
     } finally {
       releaseMutex()
     }
-    const receipt = await this.contractInteractor.sendSignedTransaction(signedTx)
-    log.info('\ntxhash is', receipt.transactionHash)
-    if (receipt.transactionHash.toLowerCase() !== storedTx.txId.toLowerCase()) {
-      throw new Error(`txhash mismatch: from receipt: ${receipt.transactionHash} from txstore:${storedTx.txId}`)
+    const transactionHash = await this.contractInteractor.broadcastTransaction(signedTx)
```
drortirosh, line 97: why id you rename it? it IS sendSignedTransaction - the way it shoudl be ..

test/relayclient/ContractInteractor.test.ts
```diff
@@ -28,4 +35,28 @@ contract('ContractInteractor', function () {
...
+      provider = new ProfilingProvider(web3.currentProvider as HttpProvider)
+      contractInteractor = new ContractInteractor(provider, configureGSN({}))
+      const nonce = await web3.eth.getTransactionCount('0xb473D6BE09D0d6a23e1832046dBE258cF6E8635B')
+      const transaction = new Transaction({ to: constants.ZERO_ADDRESS, gasLimit: '0x5208', nonce })
+      transaction.sign(Buffer.from('46e6ef4a356fa3fa3929bf4b59e6b3eb9d0521ea660fd2879c67bd501002ac2b', 'hex'))
+      sampleTransactionData = '0x' + transaction.serialize().toString('hex')
+      sampleTransactionHash = '0x' + transaction.hash(true).toString('hex')
+    })
+
+    it('should sent the transaction to the blockchain directly', async function () {
```
drortirosh, line 55: you should test what the title says you test. I don't see you check the transaction was sent to the blockchain at all: e.g. getTransaction(sampleTransactionHash)
Currently, its a test for your provider.

yarn.lock
```diff
@@ -663,37 +663,38 @@
     utf8 "^3.0.0"
     web3-utils "1.2.1"
 
-"@truffle/contract-schema@^3.2.0":
-  version "3.2.0"
-  resolved "https://registry.yarnpkg.com/@truffle/contract-schema/-/contract-schema-3.2.0.tgz#4ddd43cf3eda11ec82bb2661725a19c2f4326e96"
-  integrity sha512-yeb4UoK9cbrT5/Nuz0I0p2XKbf0K1wEmyyBQmo3Q4JOrLidxf59LtDupo9Uq74RtlTAxZC0cy9DnsfWeWVma4A==
+"@truffle/contract-schema@^3.1.0":
```
drortirosh, line 666: I see you take many versions back. My suggestion: remove the yarn.lock, re-install to let it pick the latest.

src/relayclient/ContractInteractor.ts
forshtat: Type declaration says `reason?: any` and I don't see a reason to wrap it again. It will be an error of rejected promise anyways. What do you think?

src/relayserver/TransactionManager.ts
forshtat: To differentiate from web3 sendSignedTransaction that locks thread until mined. "Broadcast" implies not caring about it being delivered IMO.

test/relayclient/ContractInteractor.test.ts
forshtat: You see, web3.sendSignedTransaction would still pass the test, because the transaction does reach the network. 
This test is white box in terms of - I know tx hash is not calculated by the Contract Interactor, so I know the correct transaction reached the blockchain and if the tx hash matches.

I do test that 'broadcastTransaction' does **not** do anything else, and as I've said, negative tests are marginally useful (aka useless 😀)

yarn.lock
forshtat: Reverted the lock file

src/relayclient/ContractInteractor.ts
drortirosh: types sucks I don't know if they declare the type correctly. What I do know is that I want to get the error just like we used to without . I only suggested to verify..

drortirosh (approved):

PR #494, opened 2020-09-16
OG-236 Local pending state

forshtat:


src/relayserver/RegistrationManager.ts
```diff
@@ -166,7 +162,7 @@ export class RegistrationManager {
     const registerEvents = relayRegisteredEvents.filter(
       (eventData: EventData) =>
         isRegistrationValid(eventData, this.config, this.managerAddress))
-    return reduceToLatestTx(registerEvents)
+    return getLatestEventData(registerEvents)
```
drortirosh, line 165: I don't think this does what you think it does: if my config matches a previous RelayRegistered  (e.g. I changed it, caused the relay to re-register and now changed it back) then the above filter on isRegistrationValid will succeed - for some previous configuration. you must first do getLatest, and only then check if it is valid.

src/relayserver/RelayServer.ts
```diff
@@ -432,20 +444,26 @@ export class RelayServer extends EventEmitter {
...
   }
 
   async getWorkerBalance (workerIndex: number): Promise<BN> {
-    return toBN(await this.contractInteractor.getBalance(this.workerAddress))
+    return toBN(await this.contractInteractor.getBalance(this.workerAddress, 'pending'))
   }
 
   async _shouldRegisterAgain (currentBlock: number, hubEventsSinceLastScan: EventData[]): Promise<boolean> {
+    const isPendingActivityTransaction =
+      (await this.txStoreManager.isActionPending(ServerAction.RELAY_CALL)) ||
```
drortirosh, line 456: relaycall should not interfere with other operation - its for a worker, not mgr.

src/common/Utils.ts
```diff
@@ -163,18 +165,26 @@ export function randomInRange (min: number, max: number): number {
   return Math.floor(Math.random() * (max - min) + min)
 }
 
-export function reduceToLatestTx (events: EventData[]): EventData | undefined {
+export function getLatestEventData (events: EventData[]): EventData | undefined {
```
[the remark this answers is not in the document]
drortirosh: actually, the events are always sorted. you can just `slice(-1)`

test/TxStoreManager.test.ts
```diff
@@ -63,6 +65,7 @@ contract('TxStoreManager', function (accounts) {
         data: '',
         nonce: 113,
         txId: '12345678',
+        serverAction: 0,
```
drortirosh, line 68: even it test, avoid numeric - its enum, so use it.

src/relayserver/RegistrationManager.ts
```diff
@@ -261,6 +257,7 @@ export class RegistrationManager {
     const addRelayWorkerMethod = await this.contractInteractor.getAddRelayWorkersMethod([this.workerAddress])
     const details: SendTransactionDetails = {
       signer: this.managerAddress,
+      serverAction: ServerAction.ADD_WORKER,
```
drortirosh, line 260: noticed you added action type, but I didn't see anywhere it is used.

src/common/Utils.ts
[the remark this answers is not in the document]
forshtat: I do not know of a way to do it. At least 'sort' is better then 'reduce' IMO 😀

src/relayserver/RegistrationManager.ts
```diff
@@ -78,13 +70,15 @@ export class RegistrationManager {
...
   config: ServerConfigParams
+  txStoreManager: TxStoreManager
 
   lastMinedRegisterTransaction?: EventData
   lastWorkerAddedTransaction?: EventData
 
   constructor (
     contractInteractor: ContractInteractor,
     transactionManager: TransactionManager,
+    txStoreManager: TxStoreManager,
```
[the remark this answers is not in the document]
forshtat: let's do it in a separate PR, it will create some mess in this one

src/relayserver/RegistrationManager.ts
forshtat: You are absolutely right... In fact, I do not need to filter at all. Fixing,

src/relayserver/RelayServer.ts
forshtat: We had previously defined an option for 'relay call' event to be a relay discovery event.
Until we make a formal decision to change that (with Jira issue) I think we should keep it as was originally designed.

src/relayserver/RegistrationManager.ts
forshtat: Fixed.

drortirosh (approved):

PR #495, opened 2020-09-21
OG-237 acceptanceBalance protocol

drortirosh:
- client should pass paymaster to relayer when calling getaddr
- getaddr's maxAcceptanceBalance should be this paymaster (or default,
if given paymaster is not trusted)
- for this the relayer caches the gasLimits of each trusted relayer..

### DIFF
--- a/src/common/PingResponse.ts
+++ b/src/common/PingResponse.ts
@@ -1,7 +1,6 @@
 import { Address, IntString } from '../relayclient/types/Aliases'
 
 export default interface PingResponse {
-  // TODO: this should be 'worker'
   relayWorkerAddress: Address
   relayManagerAddress: Address
   relayHubAddress: Address
--- a/src/common/PingResponse.ts
+++ b/src/common/PingResponse.ts
@@ -6,6 +6,8 @@ export default interface PingResponse {
   relayHubAddress: Address
   minGasPrice: IntString
   maxAcceptanceBudget: IntString
+  networkId?: IntString
+  chainId?: IntString
   ready: boolean
   version: string
 }
--- a/src/common/PingResponse.ts
+++ b/src/common/PingResponse.ts
@@ -2,11 +2,11 @@ import { Address, IntString } from '../relayclient/types/Aliases'
 
 export default interface PingResponse {
   // TODO: this should be 'worker'
-  RelayServerAddress: Address
-  RelayManagerAddress: Address
-  RelayHubAddress: Address
-  MinGasPrice: IntString
-  MaxAcceptanceBudget: IntString
-  Ready: boolean
-  Version: string
+  relayWorkerAddress: Address
+  relayManagerAddress: Address
+  relayHubAddress: Address
+  minGasPrice: IntString
+  maxAcceptanceBudget: IntString
+  ready: boolean
+  version: string
 }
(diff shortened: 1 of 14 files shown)


src/common/PingResponse.ts
```diff
@@ -2,11 +2,11 @@ import { Address, IntString } from '../relayclient/types/Aliases'
 
 export default interface PingResponse {
   // TODO: this should be 'worker'
```
forshtat, line 4: this TODO is done

src/relayserver/RelayServer.ts
```diff
@@ -299,12 +307,47 @@ export class RelayServer extends EventEmitter {
     process.exit(1)
   }
 
+  trustedPaymastersGasLimits: {[key: string]: PaymasterGasLimits} = {}
```
forshtat, line 310: Please do not define class member fields on the line 310. I know we are not in Pascal times but having methods defined in the beginning of the file is very helpful IMO.

Also, I prefer to have `Map` to typed `object`. Indexing operator is cool but somehow it does not feel right...

src/relayserver/RelayServer.ts
```diff
@@ -299,12 +307,47 @@ export class RelayServer extends EventEmitter {
...
+   * - given that, we agree to give the requested acceptanceBudget (since breaking one of the above two "invariants"
+   *    is the only cases where the relayer will have to pay for this budget)
+   *
+   * @param paymasters list of trusted paymaster addresses
+   */
+  async _initTrustedPaymasters (paymasters: string[] = []): Promise<void> {
+    this.trustedPaymastersGasLimits = {}
+    for (const paymasterAddress of paymasters) {
+      const paymaster = await this.contractInteractor._createPaymaster(paymasterAddress)
+      const gasLimits = await paymaster.getGasLimits().catch(e => {
```
forshtat, line 327: You can do `.catch((e: Error) => {` and then eslint should be happy

src/relayserver/RelayServer.ts
```diff
@@ -299,12 +307,47 @@ export class RelayServer extends EventEmitter {
...
+        // eslint-disable-next-line @typescript-eslint/restrict-template-expressions
+        throw new Error(`not a valid paymaster address in trustedPaymasters list: ${paymasterAddress}: ${e.message}`)
+      })
+      this.trustedPaymastersGasLimits[paymasterAddress.toLowerCase()] = gasLimits
+    }
+  }
+
+  _getPaymasterMaxAcceptanceBudget (paymaster?: string): IntString {
+    const limits = this.trustedPaymastersGasLimits[paymaster?.toLocaleLowerCase() ?? '']
```
forshtat, line 336: You don't need to check `this.trustedPaymastersGasLimits['']` IMO. Just return `maxAcceptanceBudget`

truffle.js
```diff
@@ -3,11 +3,13 @@ require('ts-node/register/transpile-only')
 var HDWalletProvider = require('truffle-hdwallet-provider')
 var mnemonic = 'digital unknown jealous mother legal hedgehog save glory december universe spread figure custom found six'
 
-const secretMnemonicFile = './secret_mnemonic'
+// NOGIT
```
forshtat, line 6: I don't think it worked...

drortirosh: the marker was !NOGIT... I relaxed it... removing this file

src/relayserver/RelayServer.ts
drortirosh: I do, since I want to return the default either if no paymaster is given, or if a wrong one is given...

forshtat (approved):

PR #496, opened 2020-09-22
OG-175: Add _msgData back to recipient

drortirosh:


yoavw (commented):
LGTM, but I see there are no changes to the forwarder itself, just the recipient.  Is that really the case, or are there some missing forwarder changes?

drortirosh:
> LGTM, but I see there are no changes to the forwarder itself, just the recipient. Is that really the case, or are there some missing forwarder changes?

yes, its a change to Recipient only (something any recipient can do by itself, but we do want to have a "standardized" one.

forshtat (approved):

PR #497, opened 2020-09-23
fixed LeastAuthority issues

drortirosh:
OG-242 - Issue A: RelayClient unused "filter"
OG-243 - Issue B: RelayClient timeout http requests
OG-248 - Issue G: RelayHub use gas limit when calling getGasLimits()

### DIFF
--- a/contracts/RelayHub.sol
+++ b/contracts/RelayHub.sol
@@ -122,8 +122,11 @@ contract RelayHub is IRelayHub {
     private
     view
     returns (IPaymaster.GasLimits memory gasLimits, uint256 maxPossibleGas) {
-        gasLimits =
-            IPaymaster(relayRequest.relayData.paymaster).getGasLimits();
+
+        (bool success, bytes memory ret) = relayRequest.relayData.paymaster.staticcall{gas:50000}(abi.encodeWithSelector(IPaymaster.getGasLimits.selector));
+        require(success, "paymaster.getGasLimits reverted (or OOG)");
+
+        gasLimits = abi.decode(ret, (IPaymaster.GasLimits));
 
         require(paymasterMaxAcceptanceBudget >= gasLimits.acceptanceBudget, "unexpected high acceptanceBudget");
 
--- a/contracts/RelayHub.sol
+++ b/contracts/RelayHub.sol
@@ -21,7 +21,7 @@ import "./interfaces/IStakeManager.sol";
 contract RelayHub is IRelayHub {
     using SafeMath for uint256;
 
-    string public override versionHub = "2.0.0-beta.1+opengsn.hub.irelayhub";
+    string public override versionHub = "2.0.0-beta.3+opengsn.hub.irelayhub";
 
     uint256 public override minimumStake;
     uint256 public override minimumUnstakeDelay;
--- a/contracts/RelayHub.sol
+++ b/contracts/RelayHub.sol
@@ -122,11 +122,8 @@ contract RelayHub is IRelayHub {
     private
     view
     returns (IPaymaster.GasLimits memory gasLimits, uint256 maxPossibleGas) {
-
-        (bool success, bytes memory ret) = relayRequest.relayData.paymaster.staticcall{gas:50000}(abi.encodeWithSelector(IPaymaster.getGasLimits.selector));
-        require(success, "paymaster.getGasLimits reverted (or OOG)");
-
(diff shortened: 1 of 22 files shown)


contracts/RelayHub.sol
```diff
@@ -122,8 +122,11 @@ contract RelayHub is IRelayHub {
     private
     view
     returns (IPaymaster.GasLimits memory gasLimits, uint256 maxPossibleGas) {
-        gasLimits =
-            IPaymaster(relayRequest.relayData.paymaster).getGasLimits();
+
+        (bool success, bytes memory ret) = relayRequest.relayData.paymaster.staticcall{gas:50000}(abi.encodeWithSelector(IPaymaster.getGasLimits.selector));
```
forshtat, line 126: If I understand solidity correctly, it will compile the line with a call to view function as a staticcall, won't it?
If so, is there a benefit in going to `staticcall{gas:50000}` over `.getGasLimits{gas:50000}()` ? Just asking.

src/relayclient/HttpWrapper.ts
```diff
@@ -1,13 +1,15 @@
...
 const LOGMAXLEN = 120
+const DEFAULT_TIMEOUT = 15000
 
 export default class HttpWrapper {
   private readonly provider: AxiosInstance
   private readonly logreq: boolean
 
   constructor (opts: AxiosRequestConfig = {}, logreq: boolean = false) {
     this.provider = axios.create(Object.assign({
+      timeout: DEFAULT_TIMEOUT,
```
forshtat, line 12: Is there any chance at some point we will want to make this parameter configurable?

src/relayclient/HttpWrapper.ts
```diff
@@ -26,12 +28,16 @@ export default class HttpWrapper {
...
-  async sendPromise (url: string, jsonRequestData: any): Promise<any> {
+  async sendPromise (url: string, jsonRequestData?: any): Promise<any> {
     if (this.logreq) {
       console.log('sending request:', url, JSON.stringify(jsonRequestData ?? {}).slice(0, LOGMAXLEN))
     }
 
-    const response = await this.provider.post(url, jsonRequestData)
+    const response = await this.provider.request({
+      url,
+      method: jsonRequestData != null ? 'POST' : 'GET',
```
forshtat, line 38: oh yeah, this went well...

src/relayclient/RelayClient.ts
```diff
@@ -360,3 +360,24 @@ export class RelayClient {
     return forwarderAddress
   }
 }
+
+export function _dumpRelayingResult (relayingResult: RelayingResult): string {
```
forshtat, line 364: `RelayClient.ts` is a weird placement for this function. Why?

src/relayclient/RelayProvider.ts
```diff
@@ -130,7 +130,7 @@ export class RelayProvider implements HttpProvider {
           const jsonRpcSendResult = this._convertTransactionToRpcSendResponse(relayingResult.transaction, payload)
           callback(null, jsonRpcSendResult)
         } else {
-          const message = `Failed to relay call. Results:\n${this._dumpRelayingResult(relayingResult)}`
+          const message = `Failed to relay call. Results:\n${_dumpRelayingResult(relayingResult)}`
```
forshtat, line 133: this is the only place that ever calls to '_dumpRelayingResult', innit?

contracts/RelayHub.sol
drortirosh: nope... I forgot (or completely didn't know) you can put gas limit on any method call...

src/relayclient/HttpWrapper.ts
drortirosh: yes, of course its configurable. in our test we override it with timeout:1

src/relayclient/RelayClient.ts
drortirosh: RelayProvider is weirder place for a method parse a return value of RelayClient.

I really wished this formatter to be a method off the RelayingResult itself - but its an interface, not a real class, so I ended up creating a method.
its still used only internally, so its "_"-prefixed

other option is make RelayingResult a class, and put this into its "toString"

src/relayclient/RelayProvider.ts
drortirosh: yes, this and a test.
we could also create it with the formatted string as a member. is it better?

forshtat (approved):

PR #499, opened 2020-09-24
OG-254 "Light" TruffleContract

drortirosh:
remove TruffleContract dependency
Use instead "light" version with has the same syntax ("at" function,
options as last arg, "intelligent" call/send method, getPastEvents with
"args") - as much as our project requires..

### DIFF
--- a/src/relayclient/LightTruffleContract.ts
+++ b/src/relayclient/LightTruffleContract.ts
@@ -1,0 +1,90 @@
+import Web3 from 'web3'
+import { AbiItem, AbiOutput, toBN } from 'web3-utils'
+import { Web3Provider } from './ContractInteractor'
+
+function retypeItem (abiOutput: AbiOutput, ret: any): any {
+  if (abiOutput.type.includes('int')) {
+    return toBN(ret)
+  } else {
+    return ret
+  }
+}
+
+// restore TF type: uint are returned as string in web3, and as BN in TF.
+function retype (outputs?: AbiOutput[], ret?: any): any {
+  if (outputs?.length === 1) {
+    return retypeItem(outputs[0], ret)
+  } else {
+    // seems like structure return values in truffle are left as strings,
+    return ret
+  }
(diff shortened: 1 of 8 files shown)


src/relayclient/LightTruffleContract.ts
```diff
@@ -0,0 +1,90 @@
...
+}
+
+export class Contract<T> {
+  web3!: Web3
+
+  constructor (readonly contractName: string, readonly abi: AbiItem[]) {
+  }
+
+  createContract (address: string): any {
+    return new this.web3.eth.Contract(this.abi, address)
```
forshtat, line 30: let's keep the 'code size' check logic in create contract? I came to rely on it instead of checking manually anywhere

src/relayclient/LightTruffleContract.ts
```diff
@@ -0,0 +1,90 @@
...
+          args: e.returnValues // TODO: web3 uses strings, Truffle uses BN for numbers
+        }))
+      }
+    } as any
+
+    this.abi.forEach(m => {
+      const methodName: string = m.name ?? ''
+      const nArgs = m.inputs?.length ?? 0
+      const isViewFunction = m.stateMutability === 'view' || m.stateMutability === 'pure'
+      obj[methodName] = async function () {
```
forshtat, line 55: is this copied from truffle or manual rewrite?

src/relayclient/LightTruffleContract.ts
drortirosh: we always query contracts for their version. why duplicate (and pull entire bytecode) each time?

src/relayclient/LightTruffleContract.ts
drortirosh: its a complete rewrite from scratch,.

forshtat (approved):

forshtat (approved):

PR #504, opened 2020-09-27
OG-255 Use version range in package.json

forshtat:


package.json
```diff
@@ -50,49 +50,48 @@
   },
   "dependencies": {
     "@openeth/truffle-typings": "0.0.6",
-    "@openzeppelin/contracts": "3.0.2",
-    "@truffle/contract": "4.2.4",
+    "@openzeppelin/contracts": "^3.2.0",
+    "@truffle/contract": "^4.2.23",
     "@truffle/hdwallet-provider": "1.0.34",
```
drortirosh, line 55: why is hdwallet provider kept the same? unable to use later version?

drortirosh (approved):

PR #506, opened 2020-09-29
fix relayer registration flow.

drortirosh:
- removed error messages during registration (which are not errors)
- it takes ~3 blocks to register relayer. increased "gsn
relayer-register" timeout to 60 seconds.

- moved some "info" to "debug"
- "null block in receipt" appears too often, and doesn't reflect an
error. now doesn't throw an exception - just log a line.
- polling interval reduced to 3sec
- ready state times-out after 3 `checkInterval` intervals, so any kind
of error will make the server "not ready" at most with 30 sec delay
(client has 10-sec timeout, so they will switch to other relayer faster)

### DIFF
--- a/src/cli/CommandsLogic.ts
+++ b/src/cli/CommandsLogic.ts
@@ -104,11 +104,11 @@ export default class CommandsLogic {
     return response.ready
   }
 
-  async waitForRelay (relayUrl: string): Promise<void> {
-    const timeout = 30
+  async waitForRelay (relayUrl: string, timeout = 60): Promise<void> {
     console.error(`Will wait up to ${timeout}s for the relay to be ready`)
 
-    for (let i = 0; i < timeout; ++i) {
+    const endTime = Date.now() + timeout * 1000
+    while (Date.now() < endTime) {
       let isReady = false
       try {
         isReady = await this.isRelayReady(relayUrl)
@@ -118,7 +118,7 @@ export default class CommandsLogic {
       if (isReady) {
         return
       }
-      await sleep(1000)
+      await sleep(3000)
     }
     throw Error(`Relay not ready after ${timeout}s`)
   }
--- a/src/relayserver/RelayServer.ts
+++ b/src/relayserver/RelayServer.ts
@@ -37,15 +37,11 @@ import Timeout = NodeJS.Timeout
 const VERSION = '2.0.0-beta.3'
 const GAS_RESERVE = 100000
 
-// must receive some response for request from node
-const READY_TIMEOUT = 30000
-
 export class RelayServer extends EventEmitter {
   lastScannedBlock = 0
   lastRefreshBlock = 0
   ready = false
   lastWorkerFinished = Date.now()
-  lastError = null
   readonly managerAddress: PrefixedHexString
   readonly workerAddress: PrefixedHexString
   gasPrice: number = 0
@@ -315,19 +311,13 @@ returnValue        | ${viewRelayCallRet.returnValue}
         if (transactions.length !== 0) {
           log.debug(`Done handling block #${blockNumber}. Created ${transactions.length} transactions.`)
         }
-        this.lastError = null
-        this.lastWorkerFinished = Date.now()
-        this._workerSemaphoreOn = false
       })
       .catch((e) => {
-        // don't spam log with repeated error: report only the first time, until any change (either in error string,
-        // or successful _worker with no error)
-        if (e !== this.lastError) {
-          this.emit('error', e)
-          log.error('error in worker:', e)
-          this.lastError = e
-        }
+        this.emit('error', e)
+        log.error('error in worker:', e)
         this.setReadyState(false)
+      })
+      .finally(() => {
         this.lastWorkerFinished = Date.now()
         this._workerSemaphoreOn = false
       })
@@ -633,7 +623,7 @@ latestBlock timestamp   | ${latestBlock.timestamp}
   isReady (): boolean {
     if (!this.ready) { return false }
 
-    if ((Date.now() - this.lastWorkerFinished) > READY_TIMEOUT) {
+    if ((Date.now() - this.lastWorkerFinished) > this.config.readyTimeout) {
       log.warn(chalk.bgRedBright('Relay state: Timed-out'))
       this.ready = false
     }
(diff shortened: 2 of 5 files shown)


src/cli/CommandsLogic.ts
```diff
@@ -104,11 +104,11 @@ export default class CommandsLogic {
     return response.ready
   }
 
-  async waitForRelay (relayUrl: string): Promise<void> {
-    const timeout = 30
+  async waitForRelay (relayUrl: string, timeout = 60): Promise<void> {
```
forshtat, line 107: I don't see where you are using the ability to specify a custom timeout. What was wring with the old code?

src/relayserver/RelayServer.ts
```diff
@@ -37,10 +37,15 @@ import Timeout = NodeJS.Timeout
 const VERSION = '2.0.0-beta.3'
 const GAS_RESERVE = 100000
 
+// must receive some response for request from node
+const READY_TIMEOUT = 30000
```
forshtat, line 41: I think this should go to the configuration

src/relayserver/RelayServer.ts
```diff
@@ -37,10 +37,15 @@ import Timeout = NodeJS.Timeout
...
+
 export class RelayServer extends EventEmitter {
   lastScannedBlock = 0
   lastRefreshBlock = 0
   ready = false
+  lastWorkerFinished = -1
+  lastError = null
```
forshtat, line 48: I am not sure this is a right approach to error logging, to be honest.

src/relayserver/RelayServer.ts
```diff
@@ -309,12 +315,20 @@ returnValue        | ${viewRelayCallRet.returnValue}
...
+        this.lastError = null
+        this.lastWorkerFinished = Date.now()
         this._workerSemaphoreOn = false
       })
       .catch((e) => {
-        this.emit('error', e)
-        log.error('error in worker:', e)
+        // don't spam log with repeated error: report only the first time, until any change (either in error string,
+        // or successful _worker with no error)
+        if (e !== this.lastError) {
```
forshtat, line 325: You are comparing Error instances, will never be false AFAIK

src/relayserver/RelayServer.ts
```diff
@@ -309,12 +315,20 @@ returnValue        | ${viewRelayCallRet.returnValue}
...
+          this.emit('error', e)
+          log.error('error in worker:', e)
+          this.lastError = e
+        }
         this.setReadyState(false)
+        this.lastWorkerFinished = Date.now()
```
forshtat, line 331: Any reason for 'catch' to be considered 'worker finished'? I know `ready` is false anyway, so it does not matter for timed-out readiness.

src/relayserver/RelayServer.ts
```diff
@@ -555,7 +569,8 @@ latestBlock timestamp   | ${latestBlock.timestamp}
       fromBlock: this.lastScannedBlock + 1,
       toBlock: 'latest'
     }
-    return await this.contractInteractor.getPastEventsForHub(topics, options)
+    const events = await this.contractInteractor.getPastEventsForHub(topics, options)
+    return events
```
forshtat, line 573: Idea complains about it as 'redundant', do you really need it?

src/relayserver/TransactionManager.ts
```diff
@@ -226,11 +226,12 @@ data         | 0x${transaction.data.toString('hex')}
       if (shouldRecheck) {
         const receipt = await this.contractInteractor.getTransaction(transaction.txId)
         if (receipt == null) {
-          log.warn(`failed to fetch receipt for tx ${transaction.txId}`)
+          log.warn(`warning: failed to fetch receipt for tx ${transaction.txId}`)
```
forshtat, line 229: That's actually fine, the transaction is not yet mined so the node does not give you a receipt.

src/relayserver/TransactionManager.ts
```diff
@@ -226,11 +226,12 @@ data         | 0x${transaction.data.toString('hex')}
...
           continue
         }
         if (receipt.blockNumber == null) {
-          throw new Error(`invalid block number in receipt ${JSON.stringify(receipt)}`)
+          log.warn(`warning: null block number in receipt for ${transaction.txId}`)
```
forshtat, line 233: And this happens a lot and I have no idea why. Not mined yet but here is your receipt? Is this how this RPC call supposed to work? Not what the official wiki says.

```
blockNumber: QUANTITY - block number where this transaction was in.
...
...
contractAddress : DATA, 20 Bytes - The contract address created, if the transaction was a contract creation, otherwise null.
```

src/relayserver/RelayServer.ts
forshtat: Like, "this error happened once and I don't see it happening ever again" is a very different thing from "ok, this error has happened 356834 times over the last minute, something is clearly not good".

src/relayserver/RelayServer.ts
drortirosh: yes, its here for completeness only. we need a kind of "finally",

src/relayserver/RelayServer.ts
```diff
@@ -309,12 +315,20 @@ returnValue        | ${viewRelayCallRet.returnValue}
...
+        this.lastWorkerFinished = Date.now()
         this._workerSemaphoreOn = false
       })
       .catch((e) => {
-        this.emit('error', e)
-        log.error('error in worker:', e)
+        // don't spam log with repeated error: report only the first time, until any change (either in error string,
+        // or successful _worker with no error)
+        if (e !== this.lastError) {
+          this.emit('error', e)
```
drortirosh, line 326: ok, removing. need to be written better (and yes, this doesn't work..)

src/relayserver/RelayServer.ts
drortirosh: ok

src/relayserver/RelayServer.ts
drortirosh: not sure - doesn't complain for me..

src/relayserver/TransactionManager.ts
drortirosh: I didn't remove it, since I don't understand when it happens. just removed the stack trace, which is complete nonsense.

forshtat (approved):

PR #512, opened 2020-10-05
OG-275 fix relayer state timeout

drortirosh:
- relayer timeout now managed by setTimeout callback
- added deferReadiness: control when setReady takes effect immediately or after delay (see below): deferReadiness becomes "true" after first crash of worker.
- added timeToReady: when setting relayer to ready (in deferReadiness), it doesn't take effect immediately but only after readyTimeout.

- removed the "ready" flag. instead, there is "futureReadyTime":
  - < Date.now() - we're already ready
  - Nan - we're in "not-ready" state (after exception)
  - &gt; Date.now() - we're waiting for this future time to become ready

- The "deferReadiness" controls whether setReadyState(true) will set
    futureTime to "now" or to some time into the future

### DIFF
--- a/src/relayserver/RelayServer.ts
+++ b/src/relayserver/RelayServer.ts
@@ -623,7 +623,8 @@ latestBlock timestamp   | ${latestBlock.timestamp}
   isReady (): boolean {
     if (!this.ready) { return false }
 
-    if ((Date.now() - this.lastWorkerFinished) > this.config.readyTimeout) {
+    const timedOut = (Date.now() - this.lastWorkerFinished) > this.config.readyTimeout
+    if (!this.config.devMode && timedOut) {
       log.warn(chalk.bgRedBright('Relay state: Timed-out'))
       this.ready = false
     }
--- a/src/relayserver/RelayServer.ts
+++ b/src/relayserver/RelayServer.ts
@@ -282,37 +282,38 @@ returnValue        | ${viewRelayCallRet.returnValue}
     return signedTx
   }
 
+  async intervalHandler (): Promise<void> {
+    const now = Date.now()
+    let workerTimeout: Timeout
+    if (!this.config.devMode) {
+      workerTimeout = setTimeout(() => {
+        log.warn(chalk.bgRedBright('Relay state: Timed-out after %d'), Date.now() - now)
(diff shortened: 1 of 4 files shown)


src/relayserver/RelayServer.ts
```diff
@@ -79,6 +87,7 @@ export class RelayServer extends EventEmitter {
     this.workerBalanceRequired = new AmountRequired('Worker Balance', toBN(this.config.workerMinBalance))
     this.printServerAddresses()
     log.setLevel(this.config.logLevel)
+    log.setLevel(0)
```
forshtat, line 90: !

src/relayserver/RelayServer.ts
```diff
@@ -40,8 +40,16 @@ const GAS_RESERVE = 100000
...
+
+  // setReadyState set the time when the relayer becomes ready. negative= not ready
+  // Nan = not ready.
+  // Zero is already ready Since date.now is bigger)
+  // now+timeout - will be ready after timeout.
+  futureReadyTime = NaN
+
+  // if deferReadiness is true, then setReadyState(true) takes effect only after relayTimeout.
+  // it is set only after worker exception (=probably some provider error)
+  deferReadiness = false
```
forshtat, line 52: Ok, I am not approving this. This is a patchwork and it does not solve the issue.

Ready flag should stay there - it indicates balances and registration is going well. It should not be abused to also mean 'Infura connection is good' - lets create a separate mechanism for that.

Also, I don't see a need for 'deferReadiness' flag.  It never becomes 'false' again, so it is basically one-time thing.

The 'futureReadyTime' is also problematic IMO.

How about that - we have a 'successful handler runs' variable, init it with, IDK, `requiredSuccessfulHandlerRuns == 10`.

On a single thrown exception it is reset to 0, on a successful run it is incremented. Also if the worker does not complete before `workerTimeout` is executed.

Only consider relay to be 'ready' if it is 
a) old code `ready === true`
b) `this.successfulHandlerRuns > this.config.requiredSuccessfulHandlerRuns`

forshtat (changes requested):

forshtat (approved):

PR #514, opened 2020-10-07
OG-293 Allow skipping 'isTrustedForwarder' check

forshtat:
It is possible tot make relayed calls to contracts that are not
recipients if they do not accesss 'msg.sender'.
This flag enables such use-case.

### DIFF
--- a/src/relayclient/GSNConfigurator.ts
+++ b/src/relayclient/GSNConfigurator.ts
@@ -28,7 +28,7 @@ const DEFAULT_RELAY_TIMEOUT_GRACE_SEC = 1800
 const DEFAULT_LOOKUP_WINDOW_BLOCKS = 60000
 
 const defaultGsnConfig: GSNConfig = {
-  tmpCheckRecipientForwarder: true,
+  skipRecipientForwarderValidation: false,
   preferredRelays: [],
   relayLookupWindowBlocks: DEFAULT_LOOKUP_WINDOW_BLOCKS,
   gasPriceFactorPercent: GAS_PRICE_PERCENT,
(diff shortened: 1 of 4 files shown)


src/relayclient/GSNConfigurator.ts
```diff
@@ -19,6 +19,7 @@ const DEFAULT_RELAY_TIMEOUT_GRACE_SEC = 1800
 const DEFAULT_LOOKUP_WINDOW_BLOCKS = 60000
 
 const defaultGsnConfig: GSNConfig = {
+  tmpCheckRecipientForwarder: true,
```
shahafn, line 22: Why is it 'tmp' then?

src/relayclient/RelayClient.ts
```diff
@@ -336,7 +336,7 @@ export class RelayClient {
       if (!isRecipientDeployed) {
         console.warn(`No IRelayRecipient code at ${gsnTransactionDetails.to}, proceeding without validating 'isTrustedForwarder'!
         Unless you are using some counterfactual contract deployment technique the transaction will fail!`)
-      } else {
+      } else if (this.config.tmpCheckRecipientForwarder) {
```
drortirosh, line 339: 1. do we have test for the old behaviour? - if not, we should
2. check the new behaviour when disabling the check
3. if behaviour is the same (e.g. just change of revert reason) - then no need for the check at all... (hint: I think that's the case: the paymaster check it, and reports its result to the client)

drortirosh: Actually, it should be simpler: if there is a forwarder (`if (forwarderAddress !== constants.ZERO_ADDRESS)`  in line 334 above) - then don't test anything.
All those tests are done by the paymaster, and reported as errors by it.  all the tests in the client are completely redundant.

forshtat: Well, the problem with this approach is on-chain the test will be done after signing, and will be treated as a 'relay' failure (so the user will be prompted to sign again, and again).
I think it is well worth it to make this extra RPC call by default, and expose the config to skip the check.

drortirosh (approved):

PR #515, opened 2020-10-11
OG-268: revert reason of relayCall

drortirosh:
Client should see revert reasons (when called as view functions), as
these often contains useful diagnostics (e.g. paymaster not funded)

Unfortunately, there is no standard of returning revert reasons, and
Parity revert reason is hidden by the underlying provider in generic
exception.
Geth revert reason can be parsed (as ganache...)
The new test (RevertMessage.test.ts) runs on ganache, but can also be
executed as stand-alone against rinkeby (geth) where it succeeds ,and
kova (parity) where it fails..

### DIFF
--- a/src/relayclient/ContractInteractor.ts
+++ b/src/relayclient/ContractInteractor.ts
@@ -1,10 +1,9 @@
 import Common from 'ethereumjs-common'
 import Web3 from 'web3'
-import log from 'loglevel'
 import { BlockTransactionString } from 'web3-eth'
 import { EventData, PastEventOptions } from 'web3-eth-contract'
 import { PrefixedHexString, TransactionOptions } from 'ethereumjs-tx'
-import { toBN } from 'web3-utils'
+import { toBN, toHex } from 'web3-utils'
 import {
   BlockNumber,
   HttpProvider,
@@ -15,6 +14,7 @@ import {
   WebsocketProvider
 } from 'web3-core'
 
+import abi from 'web3-eth-abi'
 import RelayRequest from '../common/EIP712/RelayRequest'
 import paymasterAbi from '../common/interfaces/IPaymaster.json'
 import relayHubAbi from '../common/interfaces/IRelayHub.json'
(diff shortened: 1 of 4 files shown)


src/relayclient/ContractInteractor.ts
```diff
@@ -259,31 +259,64 @@ export default class ContractInteractor {
...
+            reject(err)
+          } else {
+            resolve(res.result)
+          }
         })
-      log.info(res)
+      })
+
+      // @ts-ignore
+      const decoded = abi.decodeParameters(['bool', 'string'], res as string)
```
forshtat, line 314: I think you can do `const res: string = await blah()`, so you don't need to cast every time you use it

test/relayclient/RevertMessage.test.ts
```diff
@@ -0,0 +1,106 @@
...
+      // however, in Truffle, all contracts are built BEFORE the test have started, and COPIED the web3,
+      // so changing the global one is not enough.
+      // @ts-ignore
+      TestRecipient.web3.setProvider(relayProvider)
+
+      gasLess = relayProvider.newAccount().address
+    })
+
+    it('should show relayCall revert messages', async function () {
+      this.timeout(30000)
```
forshtat, line 95: Why does this simple test require a 30 seconds timeout?

test/relayclient/RevertMessage.test.ts
```diff
@@ -0,0 +1,106 @@
...
+      const gsnConfig = configureGSN({
+        logLevel: 5,
+        relayHubAddress,
+        paymasterAddress
+      })
+
+      relayProvider = new RelayProvider(underlyingProvider as any, gsnConfig)
+      // NOTE: in real application its enough to set the provider in web3.
+      // however, in Truffle, all contracts are built BEFORE the test have started, and COPIED the web3,
+      // so changing the global one is not enough.
```
forshtat, line 87: I think this comment was copied into here. I am not sure what our policy should be for multiline comments that are relevant in multiple places. What is the best practice?

forshtat (dismissed):

src/relayclient/ContractInteractor.ts
drortirosh: ok

test/relayclient/RevertMessage.test.ts
drortirosh: removed. indeed, this is a "side-test" (you can compare with RelayClient.test, as here I use startGsn which makes it MUCH smaller, even though its a little slower to start

test/relayclient/RevertMessage.test.ts
drortirosh: it isn't. but on ropsten it sometimes take more than default 2 seconds. I didn't calculate how much timeout is good enough...

forshtat (approved):

PR #516, opened 2020-10-11
OG-280 relayclient initialization outside of first transaction

drortirosh:
default construction should be: await new RelayClient(...).init()
Can still construct the old way, but it will generate a warning (since
it delays the first transaction)

NOTE: init() is now guarded with mutex, since its possibel that
relayTransaction() will be called before original init() finished.

### DIFF
--- a/src/relayclient/RelayClient.ts
+++ b/src/relayclient/RelayClient.ts
@@ -179,7 +179,7 @@ export class RelayClient {
   async relayTransaction (gsnTransactionDetails: GsnTransactionDetails): Promise<RelayingResult> {
     if (!this.initialized) {
       if (!this.initMutex.isLocked()) {
-        log.warn('better call RelayClient.init() in advance (to make first request faster)')
+        this._warn('better call RelayClient.init() in advance (to make first request faster)')
       }
       await this.init()
     }
@@ -216,6 +216,10 @@ export class RelayClient {
         pingErrors: relaySelectionManager.errors
       }
     }
+  }
+
+  _warn (msg: string): void {
+    log.warn(msg)
   }
 
   async _calculateGasPrice (): Promise<PrefixedHexString> {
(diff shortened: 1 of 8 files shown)


src/relayclient/RelayClient.ts
```diff
@@ -162,15 +164,25 @@ export class RelayClient {
...
+      this.initialized = true
+    })
+    return this
   }
 
   async relayTransaction (gsnTransactionDetails: GsnTransactionDetails): Promise<RelayingResult> {
-    await this._init()
+    if (!this.initialized) {
+      if (!this.initMutex.isLocked()) {
+        this._warn('better call RelayClient.init() in advance (to make first request faster)')
```
forshtat, line 182: How big of a deal it would be to just enforce it? I see very little value in keeping the old flow possible.

forshtat, line 182: How big of a deal it would be to just enforce it? I see very little value in keeping the old flow possible.

src/relayclient/RelayClient.ts
```diff
@@ -162,15 +164,25 @@ export class RelayClient {
...
   }
 
-  async _init (): Promise<void> {
-    if (this.initialized) { return }
-    this.emit(new GsnInitEvent())
-    await this.contractInteractor.init()
-    this.initialized = true
+  async init (): Promise<this> {
+    await this.initMutex.runExclusive(async () => {
+      if (this.initialized) {
```
forshtat, line 169: Obviously, it would be much simpler not to have a mutex and only support one way to init an object, wouldn't it?

forshtat (changes requested):

src/relayclient/RelayClient.ts
drortirosh: one way would mean you MUST wait for complete initialization before calling "send".
this way, you can init in the background, and call send even before its finished. the send will wait for init to complete and continue.

src/relayclient/RelayClient.ts
drortirosh: see my previous comment: enforcing it means the UI has to wait for complete initialization before allowing calling "send"

forshtat (approved):

forshtat (approved):

PR #517, opened 2020-10-12
Don't suppress errors when resolving config on init

amaurer:
"Not a paymaster contract" thrown error was hiding the real error which was related to the web3 instance connection which had nothing to do with the contract instance or deployment.

image


drortirosh:
this issue was already resolved (expose underlying error)

PR #518, opened 2020-10-13
OG-272 Implement Winston logging collection, configurable for Loggly …

forshtat:
…transport

PR #519, opened 2020-10-19
simplified logging URL handling

drortirosh:
- use URL in browser. override it in node (which doesn't have native "URL")
- simplified console logging:
  - reduced userid to 6 digits (no need for 10-12)
  - use "simple colorified" format

e.g.: 
```
debug: Created remote logs collecting logger for gsnuser958408 {"version":"2.0.1","service":"gsn-client","isBrowser":true,"userId":"gsnuser958408"}
```

(note that meta data is still added, but at the end, and the text is left intact. especially important since some of our logs are multiline, which looks ugly if log is jsnoified...


forshtat:
Implemented as part of the https://github.com/opengsn/gsn/pull/518/

PR #522, opened 2020-11-03
OG-253 fix preferred relays

drortirosh:
add a test to make sure preferred relays are used

### DIFF
--- a/src/relayclient/KnownRelaysManager.ts
+++ b/src/relayclient/KnownRelaysManager.ts
@@ -140,7 +140,9 @@ export default class KnownRelaysManager implements IKnownRelaysManager {
 
   async getRelaysSortedForTransaction (gsnTransactionDetails: GsnTransactionDetails): Promise<RelayInfoUrl[][]> {
     const sortedRelays: RelayInfoUrl[][] = []
-    for (let i = 0; i < this.knownRelays.length; i++) {
+    // preferred relays are kept as-is, unsorted (we don't have any info about them anyway to sort)
+    sortedRelays[0] = this.knownRelays[0] ?? []
+    for (let i = 1; i < this.knownRelays.length; i++) {
       sortedRelays[i] = await this._sortRelaysInternal(gsnTransactionDetails, this.knownRelays[i])
     }
     return sortedRelays
--- a/src/relayclient/KnownRelaysManager.ts
+++ b/src/relayclient/KnownRelaysManager.ts
@@ -52,7 +52,8 @@ export default class KnownRelaysManager implements IKnownRelaysManager {
   private latestScannedBlock: number = 0
   private relayFailures = new Map<string, RelayFailureInfo[]>()
 
-  public readonly knownRelays: RelayInfoUrl[][] = []
+  public preferredRelayers: RelayInfoUrl[] = []
+  public allRelayers: RelayInfoUrl[] = []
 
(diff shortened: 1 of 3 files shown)


src/relayclient/KnownRelaysManager.ts
```diff
@@ -140,7 +140,9 @@ export default class KnownRelaysManager implements IKnownRelaysManager {
 
   async getRelaysSortedForTransaction (gsnTransactionDetails: GsnTransactionDetails): Promise<RelayInfoUrl[][]> {
     const sortedRelays: RelayInfoUrl[][] = []
-    for (let i = 0; i < this.knownRelays.length; i++) {
+    // preferred relays are kept as-is, unsorted (we don't have any info about them anyway to sort)
```
forshtat, line 143: Why does it matter? No data to sort on -> no sorting, are you optimising CPU cycles here? If so, please don't.

test/relayclient/RelayClient.test.ts
```diff
@@ -87,6 +88,22 @@ contract('RelayClient', function (accounts) {
   let from: Address
   let data: PrefixedHexString
   let gsnEvents: GsnEvent[] = []
+  const cheapRelayerUrl = 'http://localhost:54321'
+
+  // register a very cheap relayer, so client will attempt to use it first.
+  async function registerCheapRelayer (relayHub: RelayHubInstance): Promise<void> {
```
forshtat, line 94: I learned a hard way that utility functions inside test files are harmful. They can not be exported and lead to so much duplication it eventually hurts. Especially if you only use this function once so far.

forshtat (changes requested):

test/relayclient/RelayClient.test.ts
drortirosh: what do you suggest? I prefer a helper function than put the code inline.
there's a chance that it will be reused, and in that case, it should be easy to export or move into util folder.
but since its used just once right now, there's no need to do it in advance.

src/relayclient/KnownRelaysManager.ts
drortirosh: this IS the bug that was in preferred relay. the root cause is that the sorting code chokes when passed a no-info relayinfo... 
instead, we just need to leave the preferred ones as-is, so that's what I do.

test/relayclient/RelayClient.test.ts
forshtat: There is totally a need to do it in advance. It grows into a net of functions all relying on local, global variables, state and stuff. Then once you want to export it you have to get to some advanced copy-pasting. All functions should be imported from somewhere. Then it can be easily reused in different tests.

src/relayclient/KnownRelaysManager.ts
forshtat: Ok, so maybe the sorting code should not choke on no-info, but just leave the items unsorted? 
Why do we even put relays in a two-dimensional array if we treat them as separate variables?

drortirosh: we don't. after sorting (which relies on the struct info) we no longer treat them separately. That's why keeping the first line out of sorting solves the bug (try to checkout, remove this fix and see the new test fails..)

test/relayclient/RelayClient.test.ts
drortirosh: but if you keep every one-use utility function in a central place, you end up with a pile of garbage you no longer remember if/when it is in use.
The time to refactor out is when it is used twice.

forshtat (approved):

PR #524, opened 2020-11-03
OG-264 simplify Forwarder.updateNonce

drortirosh:
Forwarder.updateNonce should also do the work of verifyNonce.
The separation (when called from Forwrader.execute()) costs ~1500 gas


drortirosh:
Actually, this issue should be deferred to be incorporate at a gas-optimization overhaul

forshtat (approved):

PR #525, opened 2020-11-05
OG-122: split lookup window

drortirosh:
added configuration item `relayLookupWindowParts` (which defaults to
"1"), to split the lookup window (defined in blocks) to multiple
requests.
added a code to increase the lookupWindowParts in case the lookup
transaction fails on "query returned more than XXX events"
(note that the increase stays in memory. if the client reloads, it will
use the configured value again)

### DIFF
--- a/src/relayclient/KnownRelaysManager.ts
+++ b/src/relayclient/KnownRelaysManager.ts
@@ -13,6 +13,7 @@ import ContractInteractor, {
   StakeUnlocked
 } from './ContractInteractor'
 import { LoggerInterface } from '../common/LoggerInterface'
+import { EventData } from 'web3-eth-contract'
 
 export const EmptyFilter: RelayFilter = (): boolean => {
   return true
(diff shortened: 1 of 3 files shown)


src/relayclient/KnownRelaysManager.ts
```diff
@@ -103,14 +104,51 @@ export default class KnownRelaysManager implements IKnownRelaysManager {
...
+
+  // return events from hub. split requested range into "window parts", to avoid
+  // fetching too many events at once.
+  async getPastEventsForHub (fromBlock: number, toBlock: number): Promise<EventData[]> {
+    let relayEventParts: any[]
+    while (true) {
+      const rangeParts = this.splitRange(fromBlock, toBlock, this.config.relayLookupWindowParts)
+      try {
+        // eslint-disable-next-line @typescript-eslint/promise-function-async
+        relayEventParts = await Promise.all(rangeParts.map(({ fromBlock, toBlock }) =>
```
forshtat, line 127: Please create an array of promises and await on it. This is a little hard to read.

src/relayclient/KnownRelaysManager.ts
```diff
@@ -103,14 +104,51 @@ export default class KnownRelaysManager implements IKnownRelaysManager {
...
+            fromBlock,
+            toBlock
+          })))
+        break
+      } catch (e) {
+        if (e.toString().match(/query returned more than/) != null &&
+          this.config.relayLookupWindowBlocks > this.config.relayLookupWindowParts
+        ) {
+          this.config.relayLookupWindowParts *= 2
+          continue
```
forshtat, line 138: It is not clear what is going on here. Also, there is no clear halting condition that does not rely on remote API response. I suggest if the value is not 1, don't try to paginate. If it is 1, maybe try also 5, 10 and halt there.  Say we need  to split in like 10 pages.
With this code it will try 2, 4, 8 before it reaches 16. I think if we can't make it in 10 pages, we bail.

drortirosh: ok, maybe double is little. I still like binaries, so put 4, 16 
halting condition is needed, so I'll stop after that.

src/relayclient/KnownRelaysManager.ts
drortirosh: ok.

src/relayclient/KnownRelaysManager.ts
drortirosh: also, if the client's default is (say) 3 and it fails on "too many results". why not attempt to raise it?

src/relayclient/KnownRelaysManager.ts
```diff
@@ -103,14 +104,55 @@ export default class KnownRelaysManager implements IKnownRelaysManager {
...
+        relayEventParts = await Promise.all(getPastEventsPromises)
+        break
+      } catch (e) {
+        if (e.toString().match(/query returned more than/) != null &&
+          this.config.relayLookupWindowBlocks > this.config.relayLookupWindowParts
+        ) {
+          if (this.config.relayLookupWindowParts >= 16) {
+            throw new Error(`Too many events after splitting by ${this.config.relayLookupWindowParts}`)
+          }
+          this.config.relayLookupWindowParts *= 4
```
forshtat, line 141: You are modifying a value in an object that is supposed to be user input and is shared by all components. I suggest KRM to have a member that holds this value if it is to be reused and config to be read-only (something we don't have in code for some reason).

src/relayclient/KnownRelaysManager.ts
```diff
@@ -103,14 +104,55 @@ export default class KnownRelaysManager implements IKnownRelaysManager {
...
+        // eslint-disable-next-line @typescript-eslint/promise-function-async
+        const getPastEventsPromises = rangeParts.map(({ fromBlock, toBlock }): Promise<any> =>
+          this.contractInteractor.getPastEventsForHub([], {
+            fromBlock,
+            toBlock
+          }))
+        relayEventParts = await Promise.all(getPastEventsPromises)
+        break
+      } catch (e) {
+        if (e.toString().match(/query returned more than/) != null &&
```
forshtat, line 135: Isn't this 100% specific to Infura? Is it ok to have Infura-specific error messages hard-coded in our code?

src/relayclient/KnownRelaysManager.ts
drortirosh: the reason to modify it in-place is that future transactions will use same value, and not have to go through the doubling procedure again.
I don't persist it, since I'm not sure its a good value for all future use..
Note sure what you meant by KRM.
it is possible to copy the value out of the config, and into a member of the relayClient

src/relayclient/KnownRelaysManager.ts
drortirosh: Yes, it is... alas, I don't have any other way to recognize it.
(it has `"code":-32005`, but even that is not unique for this error)
AFAIK, the native node limits of geth/parity are much higher
(BTW: EtherScan API , which we currently don't support since its not exactly  json-rpc, return the FIRST 10000 responses, effectively paginate the response. I like that solution much better - even though its not 100% , but we will add it after moving to `ethers`, which supports etherscan API natively.)

forshtat (dismissed):

forshtat (approved):

PR #527, opened 2020-11-17
OG-306  Client doesn't expose paymaster revert reason

drortirosh:
if relayCall return "false", then the bytes are revert result, so need
to parse it as such (and not just as a string, which can fail on
"unicode error")

(found while working on the "makerDao demo.."


forshtat (approved):

src/relayclient/ContractInteractor.ts
```diff
@@ -311,12 +311,12 @@ export default class ContractInteractor {
...
-      this.logger.info(res)
+      this.logger.debug('relayCall res=' + res)
 
       // @ts-ignore
-      const decoded = abi.decodeParameters(['bool', 'string'], res)
-      const paymasterAccepted = decoded[0]
-      const returnValue = decoded[1]
+      const decoded = abi.decodeParameters(['bool', 'bytes'], res)
+      const paymasterAccepted: boolean = decoded[0]
+      const returnValue = paymasterAccepted ? decoded[1] : this._decodeRevertFromResponse({}, { result: decoded[1] }) ?? decoded[1]
```
forshtat, line 319: This is a one ugly line, I would say...

forshtat (approved):
Doesn't pass tests, though...

drortirosh:
> Doesn't pass tests, though...

the test that fails is reputation test.
created https://opengsn.atlassian.net/browse/OG-318 to fix it.

PR #531, opened 2020-11-22
OG-314 always call resolveConfigurationGSN

drortirosh:
RelayClient (and relayProvider) call resolveConfigurationGSN from the init() method.
We also validate that init() is called before actually using the provider.


src/relayclient/GSNConfigurator.ts
```diff
@@ -89,8 +86,10 @@ export async function resolveConfigurationGSN (provider: Web3Provider, partialCo
   ] = await Promise.all([
 
     partialConfig.chainId ?? contractInteractor.getAsyncChainId(),
-    paymasterInstance.getHubAddr().catch(e => { throw new Error('Not a paymaster contract') }),
-    partialConfig.forwarderAddress ?? paymasterInstance.trustedForwarder().catch(e => { throw new Error('paymaster has no trustedForwarder()') }),
+    // @ts-ignore
```
forshtat, line 89: Why did you need a `ts-ignore` here?

src/relayclient/RelayClient.ts
```diff
@@ -80,8 +81,34 @@ export class RelayClient {
...
   ) {
-    this.config = configureGSN(configOverride)
-    const dependencies = getDependencies(this.config, provider, overrideDependencies)
+    this.partialConfig = { provider, configOverride, overrideDependencies }
+    this.logger = console
+  }
+
+  initializing?: Promise<void>
+
+  async init (): Promise<this> {
```
forshtat, line 90: No, wait, we can't have both `init` and `_init`, this will cause a ton of confusion. Why do they have to be two different methods?

src/relayclient/RelayClient.ts
```diff
@@ -80,8 +81,34 @@ export class RelayClient {
...
+    if (this.initializing != null) {
+      await this.initializing
+    }
+    if (this.initialized) {
+      return this
+    }
+    let initializingResolve!: () => void
+    this.initializing = new Promise((resolve) => {
```
forshtat, line 98: What's going on here?

src/relayclient/RelayClient.ts
```diff
@@ -164,22 +195,10 @@ export class RelayClient {
...
-      this.initialized = true
-    })
-    return this
-  }
-
   async relayTransaction (gsnTransactionDetails: GsnTransactionDetails): Promise<RelayingResult> {
     if (!this.initialized) {
       if (!this.initMutex.isLocked()) {
-        this._warn('better call RelayClient.init() in advance (to make first request faster)')
+        this._warn('suggestion: call RelayProvider.init()/RelayClient.init() in advance (to make first request faster)')
```
forshtat, line 201: I think we should remove this flow. Users should just do `new RelayClient().init().then(enableSendButton)` if needed - the elapsed time will be the same.
What do you think?

test/Flows.test.ts
```diff
@@ -92,13 +92,14 @@ options.forEach(params => {
         await rhub.depositFor(paymaster.address, { value: (1e18).toString() })
 
         relayClientConfig = {
-          logLevel: 'error',
+          logLevel: 'debug',
```
forshtat, line 95: I don't want tests to output thousands of lines of useless logs by default. I prefer having 'error' level checked-in, and setting it to 'debug' when, well, actually debugging the test.

test/relayclient/RelayClient.test.ts
```diff
@@ -156,13 +162,6 @@ contract('RelayClient', function (accounts) {
...
     })
 
-    it('should throw an exception if recipient does not return address for getTrustedForwarder', async function () {
-      const badRecipient = await TestVersions.new()
-      const noForwarderDetails = Object.assign({}, options, { forwarder: undefined, to: badRecipient.address })
-      await expect(relayClient.resolveForwarder(noForwarderDetails))
-        .to.eventually.be.rejectedWith('No forwarder address configured and no getTrustedForwarder in target contract')
-    })
-
     it('should throw an exception if passed forwarder is not trusted', async function () {
```
forshtat, line 165: Don't you still have an error thrown for a similar scenario? There is no new test for the `paymaster has no trustedForwarder` error, I think.

test/relayclient/RelayClient.test.ts
```diff
@@ -179,25 +178,25 @@ contract('RelayClient', function (accounts) {
 
   describe('#relayTransaction()', function () {
     it('should warn if called relayTransaction without calling init first', async function () {
-      relayClient = new RelayClient(underlyingProvider, gsnConfig)
-      sinon.spy(relayClient, '_warn')
+      const relayClient1 = new RelayClient(underlyingProvider, gsnConfig)
```
forshtat, line 181: Why?

test/relayclient/RelayProvider.test.ts
```diff
@@ -336,6 +329,19 @@ contract('RelayProvider', function (accounts) {
...
+      const gsnConfig: Partial<GSNConfig> = {
+        paymasterAddress: misbehavingPaymaster.address,
+        relayHubAddress: relayHub.address,
+        logLevel: 'error'
+      }
+      // @ts-ignore
+      Object.keys(TestRecipient.events).forEach(function (topic) {
+        // @ts-ignore
+        relayHub.constructor.network.events[topic] = TestRecipient.events[topic]
+      })
```
forshtat, line 341: Why?

test/relayclient/RelayClient.test.ts
forshtat: I mean, the only change I can see here is rename `relayClient` to `relayClient1`.

src/relayclient/RelayClient.ts
drortirosh: I don't want to enforce it.
I want the provider to behave ok even if you don't call init()
There are scenarios where is not that easy to enable/disable the ui, or use async call.

test/Flows.test.ts
drortirosh: Mistake.. Should have added //NOGIT locally to prevent myself from committing it

test/relayclient/RelayClient.test.ts
drortirosh: Yes.
I want to add a separate PR to validate call: is optionally (and defaults to true) to check all request params and values.
When disabled, only solidity checks are done, and the call is faster.

src/relayclient/GSNConfigurator.ts
drortirosh: I think the "as Error" resolved it (and that's why ts-error is so much better)

src/relayclient/RelayClient.ts
drortirosh: Init is api to call, and does the sync check.
_init is internal. I'll rename it to _initInternal
I used to use the runExclusive mechanism, but it trashes any stack trace of errors during init.

test/relayclient/RelayClient.test.ts
drortirosh: Right. I'll remove it (it was changed a lot of times, and finally turned to it's original structure...)

test/relayclient/RelayProvider.test.ts
drortirosh: Moved it up because we need to put the paymaster into the config. 
It's true that in test it is not required, but ONLY in tests.

test/relayclient/RelayClient.test.ts
drortirosh: Will be refu removed (the code was changed several times, and eventually ended up where it started...)

src/relayclient/RelayClient.ts
drortirosh: (renamed "initializing" to "initializingPromise")
1. if currently "initializing", then await for it to finish
2. check if fully initialized, and if so, do nothing
3. create the "initializing" promise, and save its "resolve" in a local variable (initializingResolve)
4. do internalInit
5. call initizlizingResolve to let anyone who entered this method while we're initializing to know its pending

test/relayclient/RelayClient.test.ts
drortirosh: actually, misconfigured paymaster (zero forwarder or zero relayhub) can easily be checked in resolveConfigurationGsn, but it doesn't belong to this PR.

test/relayclient/RelayProvider.test.ts
forshtat: I do not unserstand the relationship between the code in lines 337 to 341 to the sentence you've written here

src/relayclient/RelayClient.ts
```diff
@@ -80,8 +81,38 @@ export class RelayClient {
...
+
+  async init (): Promise<this> {
+    if (this.initializingPromise != null) {
+      await this.initializingPromise
+    }
+    if (this.initialized) {
+      return this
+    }
+    let initializingResolve!: () => void
+    this.initializingPromise = new Promise((resolve) => {
```
forshtat, line 98: would this work?

```
async init (): Promise<this> {
  if (this.initialized) {
      throw Error('Do not call init twice!')
    }
  this.initializingPromise = this._initInternal() // no await here
  // es-lint will be angry but this is actually functionally equivalent
  return this.initializingPromise
}

  if (!this.initialized) {
    console.log('Suggestion here)
    await this.initializingPromise
  }
```


I feel like the trick with extracting the resolve of an artificial promise is a bit of a clutch. Not sure this is functionally equivalent though.

src/relayclient/RelayClient.ts
```diff
@@ -164,22 +197,10 @@ export class RelayClient {
...
-      this.emit(new GsnInitEvent())
-      await this.contractInteractor.init()
-      this.initialized = true
-    })
-    return this
-  }
-
   async relayTransaction (gsnTransactionDetails: GsnTransactionDetails): Promise<RelayingResult> {
     if (!this.initialized) {
       if (!this.initMutex.isLocked()) {
```
forshtat, line 202: Note that `initMutex` is not being locked any more, so this `isLocked ` check is not useful anymore

src/relayclient/RelayClient.ts
```diff
@@ -370,6 +391,12 @@ export class RelayClient {
 
     return forwarderAddress
   }
+
+  verifyInitialized (): void {
```
forshtat, line 395: I don't think it is important to make this state observable at all. It is an internal detail, and R.C. should be able to cope with it if you don't enforce `await rc.init()` being done first

forshtat: To be read as 'please don't' 😀

src/relayclient/RelayClient.ts
drortirosh: looks good. cleaner than my construct.

src/relayclient/RelayClient.ts
drortirosh: Please see where I used it.
e.g. addAccount/newAccount: these are non-async (and I don't want to make them so), so I can't do init().
also send
But I also don't want them to crash on strange "no member X of undefined" or something like that.

src/relayclient/RelayClient.ts
drortirosh: yes, completelyremoved it.

src/relayclient/RelayClient.ts
forshtat: But these methods do not require initialization, do they? I object to making `verifyInitialized` part of RelayClient API. Vehemently.

forshtat (approved):

PR #532, opened 2020-11-22
OG-79: Add account should accept hex string for private key

forshtat:
Remove option to add keypair with a buffer as I can't think of a usecase
where this is more convenient

### DIFF
--- a/src/relayclient/AccountManager.ts
+++ b/src/relayclient/AccountManager.ts
@@ -37,6 +37,15 @@ export default class AccountManager {
   }
 
   addAccount (privateKey: PrefixedHexString): void {
+    // temporary backward-compatibility mode: addAccount used to accept AccountKeypair with Buffer in it
+    // @ts-ignore
+    // eslint-disable-next-line @typescript-eslint/strict-boolean-expressions
+    if (privateKey.privateKey) {
+      console.error('ERROR: addAccount accepts a private key as a prefixed hex string now!')
+      // @ts-ignore
+      // eslint-disable-next-line @typescript-eslint/restrict-template-expressions
+      privateKey = `0x${privateKey.privateKey.toString('hex')}`
+    }
     const address = toAddress(privateKey)
     const keypair: AccountKeypair = {
       privateKey,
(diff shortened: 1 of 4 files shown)


src/relayclient/AccountManager.ts
```diff
@@ -32,23 +36,24 @@ export default class AccountManager {
     this.config = config
   }
 
-  addAccount (keypair: AccountKeypair): void {
-    const wallet = ethWallet.fromPrivateKey(keypair.privateKey)
-    if (!isSameAddress(toAddress(wallet), keypair.address)) {
-      throw new Error('invalid keypair')
+  addAccount (privateKey: PrefixedHexString): void {
```
drortirosh, line 39: Why change the api? 
We already released code where addAccount accepts { address, privateKey }, and now it's changed. 
This is a breaking change (even if small) in the api.
We should start exercising respecting your api and keep backward compatibility

drortirosh: at a minimum, add "backward compatibility" for JS users (which otherwise can't tell what is wrong)
```
if ( privateKey.privateKey ) privateKey = privateKey.privateKey.toString('hex')
```

forshtat: Makes sense. Fixed. This code makes TypeScript very unhappy so I think we should remove it on the GSN v.2.2.

In general I think such a minor API change is acceptable for a minor version bump.

drortirosh: as a mindset, I want us to be very careful about API.
right now we excuses: nobody uses the stack, this feature is unused anyway, etc.
but we do need to practice thinking about backward compatibility, since it takes practice...

drortirosh (approved):

src/relayclient/AccountManager.ts
drortirosh: one addition though: let's add a comment : e.g. // backward compatible to 2.0.1
before such items, so we can search them when making next version, and remove them
(there is a `@deprecated` attribute, but its for constructor or method..

PR #533, opened 2020-11-23
OG-207 commit/reveal penalization

drortirosh:
penalizer doesn't have to be a relayer:
any sender can call `commit()` and then penalize.
(relayers can still penalize without commit)

### DIFF
--- a/contracts/Penalizer.sol
+++ b/contracts/Penalizer.sol
@@ -11,7 +11,7 @@ import "./interfaces/IPenalizer.sol";
 
 contract Penalizer is IPenalizer{
 
-    string public override versionPenalizer = "2.0.0+opengsn.penalizer.ipenalizer";
+    string public override versionPenalizer = "2.1.0+opengsn.penalizer.ipenalizer";
 
     using ECDSA for bytes32;
 
@@ -26,8 +26,30 @@ contract Penalizer is IPenalizer{
 
     }
 
-    modifier relayManagerOnly(IRelayHub hub) {
-        require(hub.isRelayManagerStaked(msg.sender), "Unknown relay manager");
+    mapping( bytes32=>bool ) public commits;
+
+    /**
+     * any sender can call "commit(keccak(encodedPenalizeFunction))", to make sure
+     * no-one can front-run it to claim this penalization
+     */
+    function commit(bytes32 commitHash) public {
+        bytes32 idx=keccak256(abi.encodePacked(commitHash,msg.sender));
+        commits[idx] = true;
+    }
+
+    function _commitRevealOrRelayManager(IRelayHub hub) internal {
+        bytes32 requestHash = keccak256(msg.data);
+        bytes32 idx=keccak256(abi.encodePacked(requestHash,msg.sender));
+
+        if ( commits[idx] ) {
+            commits[idx] = false;
+        } else {
+            require(hub.isRelayManagerStaked(msg.sender), "Unknown relay manager");
+        }
+    }
+
+    modifier commitRevealOrRelayManager(IRelayHub hub) {
+        _commitRevealOrRelayManager(hub);
         _;
     }
 
@@ -40,7 +62,7 @@ contract Penalizer is IPenalizer{
     )
     public
     override
-    relayManagerOnly(hub)
+    commitRevealOrRelayManager(hub)
     {
         // Can be called by a relay manager only.
         // If a relay attacked the system by signing multiple transactions with the same nonce
@@ -85,7 +107,7 @@ contract Penalizer is IPenalizer{
     )
     public
     override
-    relayManagerOnly(hub)
+    commitRevealOrRelayManager(hub)
     {
         Transaction memory decodedTx = decodeTransaction(unsignedTx);
         if (decodedTx.to == address(hub)) {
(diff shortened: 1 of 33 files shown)


contracts/Penalizer.sol
```diff
@@ -11,7 +11,7 @@ import "./interfaces/IPenalizer.sol";
 
 contract Penalizer is IPenalizer{
 
-    string public override versionPenalizer = "2.0.0+opengsn.penalizer.ipenalizer";
+    string public override versionPenalizer = "2.1.0+opengsn.penalizer.ipenalizer";
```
drortirosh, line 14: first time to upgrade version: added functionality, and  fully backward-compatible..

contracts/Penalizer.sol
```diff
@@ -26,8 +26,30 @@ contract Penalizer is IPenalizer{
...
-        require(hub.isRelayManagerStaked(msg.sender), "Unknown relay manager");
+    mapping( bytes32=>bool ) public commits;
+
+    /**
+     * any sender can call "commit(keccak(encodedPenalizeFunction))", to make sure
+     * no-one can front-run it to claim this penalization
+     */
+    function commit(bytes32 commitHash) public {
+        bytes32 idx=keccak256(abi.encodePacked(commitHash,msg.sender));
+        commits[idx] = true;
```
forshtat, line 37: I believe it would be beneficial if the hash was same for the same penalizeable event regardless of the sender. This way, this method could maybe emit the 'committed' event and then whoever wants to penalize can check is someone has already commited the same fraud proof, saving ether to potentially huge number of people.

Later, we could have some first to commit - first to get paid logic, but this is not necessary for now I think.

contracts/Penalizer.sol
```diff
@@ -26,8 +26,30 @@ contract Penalizer is IPenalizer{
 
     }
 
-    modifier relayManagerOnly(IRelayHub hub) {
-        require(hub.isRelayManagerStaked(msg.sender), "Unknown relay manager");
+    mapping( bytes32=>bool ) public commits;
```
forshtat, line 29: Aslo, please keep the code style at least somewhat consistent. No spaces inside braces, spaces after commas etc.

contracts/Penalizer.sol
```diff
@@ -26,8 +26,30 @@ contract Penalizer is IPenalizer{
...
+    function commit(bytes32 commitHash) public {
+        bytes32 idx=keccak256(abi.encodePacked(commitHash,msg.sender));
+        commits[idx] = true;
+    }
+
+    function _commitRevealOrRelayManager(IRelayHub hub) internal {
+        bytes32 requestHash = keccak256(msg.data);
+        bytes32 idx=keccak256(abi.encodePacked(requestHash,msg.sender));
+
+        if ( commits[idx] ) {
```
forshtat, line 44: There is no requirement to have a delay in this method. This means that the frontrunner can totally put bot commit and reveal in the same block and still outrun the actual 'committer'.

forshtat (changes requested):

contracts/Penalizer.sol
```diff
@@ -23,14 +25,36 @@ contract Penalizer is IPenalizer{
...
+    mapping(bytes32 => uint) public commits;
 
+    event Commit(bytes32 indexed commitHash, address sender);
+
+    /**
+     * any sender can call "commit(keccak(encodedPenalizeFunction))", to make sure
+     * no-one can front-run it to claim this penalization
+     */
+    function commit(bytes32 msgDataHash, uint revealSalt) external override {
+        bytes32 idx = keccak256(abi.encodePacked(msgDataHash, revealSalt, msg.sender));
```
drortirosh, line 39: fix: give "idx" as parameter to commit, and no need for revealSalt.
(that is, the hash of (msgDataHash, msgSender) is done off-chain for the commit, but on-chain for the reveal

contracts/Penalizer.sol
```diff
@@ -23,14 +25,36 @@ contract Penalizer is IPenalizer{
...
+
+    mapping(bytes32 => uint) public commits;
 
+    event Commit(bytes32 indexed commitHash, address sender);
+
+    /**
+     * any sender can call "commit(keccak(encodedPenalizeFunction))", to make sure
+     * no-one can front-run it to claim this penalization
+     */
+    function commit(bytes32 msgDataHash, uint revealSalt) external override {
```
forshtat, line 38: No, don't reveal salt :)

Salt should be revealed in 'reveal' stage.

'hash' should include secret only known to the signer.

contracts/Penalizer.sol
```diff
@@ -23,14 +25,34 @@ contract Penalizer is IPenalizer{
...
 
+    modifier commitRevealOnly() {
+        bytes32 commitHash = keccak256(abi.encodePacked(keccak256(msg.data), msg.sender));
+        uint commitBlockNumber = commits[commitHash];
+        require(commitBlockNumber != 0, "no commit");
+        require(commitBlockNumber + PENALIZE_BLOCK_DELAY < block.number, "reveal penalize too soon");
+        _;
+    }
+
     function penalizeRepeatedNonce(
```
forshtat, line 56: Suggest original path to be removed. 
Registered relays can be added just to front-run penalization, decreasing motivation to monitor chain.

forshtat (approved):
Includes a lot of my own changes - need a second review.

PR #535, opened 2020-11-29
Update contracts to solc 0.7.5

drortirosh:
Interfaces are kept "version free", so they can be used with older versions (using ">0.6.10")


.solhint.json
```diff
@@ -1,7 +1,9 @@
 {
   "extends": "solhint:recommended",
   "rules": {
-    "compiler-version": ["error","^0.6.2"],
+    "compiler-version": ["error",">=0.6.10"],
+    "func-visibility": ["warn",{"ignoreConstructors":true}],
+    "reason-string": ["warn",{"maxLength":64}],
```
drortirosh, line 6: default recommendation is <32 bytes of message, which generates lighter code.
Our RelayHub, StakeManager have some longer msgs (and also some tests)

src/common/Environments.ts
```diff
@@ -12,8 +12,8 @@ interface Environment {
 }
 
 const defaultRelayHubConfiguration: RelayHubConfiguration = {
-  gasOverhead: 35965,
-  postOverhead: 13950,
+  gasOverhead: 35915,
```
drortirosh, line 15: slightly improved compiler optimization...

drortirosh:
ISSUE DEFERRED: while none of the changes in this issue seemed dangerous, it breaks the tests in a very subtle way: tests take longer and longer, until ganache completely chokes and refuses connections.
Since we don't really need the 0.7.5 code right now, we're not going to merge it.

PR #536, opened 2020-11-30
OG-20: Extract ContractInteractor from client module to be a shared c…

forshtat:
…omponent

1. Extract GSNContracctsDeployment from the GSNConfig interface
2. Move './types/' folder to 'common' module
3. Use --chainId flag for Ganache instead of in-code workarounds
4. Use named parameters for some classess' constructors
5. Rename VersionManager:isMinorSameOrNewer to a less code-specific name
6. Create ISendProvider interface for underlying providers (WIP)
7. Make LoggerConfiguration hierarchical in GSNCOnfig
8. Merge 'resolveConfigGSN' and ' getDependencies' into the RelayClient
9. Remove flat-out of 'dependencies' in RelayClient and keep them boxed
10. Remove 'resolveForwarder' as it is no longer useful (read from PM)


src/relayclient/RelayClient.ts
```diff
@@ -440,6 +430,66 @@ export class RelayClient {
       message: `Transaction was not audited. Failed audit calls: ${failedAuditorsCount}/${auditors.length}`
     }
   }
+
+  getUnderlyingProvider (): HttpProvider {
```
drortirosh, line 434: IWebProvider (aka ISendProvider)

src/relayclient/RelayClient.ts
```diff
@@ -440,6 +430,66 @@ export class RelayClient {
...
+      methodSuffix,
+      jsonStringifyRequest
+    }
+    this.config = {
+      ...defaultGsnConfig,
+      ...resolvedConfig,
+      ...partialConfig
+    }
+
+    const accountManager = overrideDependencies?.accountManager ?? new AccountManager(provider, contractInteractor.chainId, this.config)
```
drortirosh, line 468: I think that even as internals, resolve config and resolve deps should be separate functions.

src/relayclient/RelayClient.ts
```diff
@@ -440,6 +430,66 @@ export class RelayClient {
...
+
+    const resolvedConfig: Partial<GSNConfig> = {
+      methodSuffix,
+      jsonStringifyRequest
+    }
+    this.config = {
+      ...defaultGsnConfig,
+      ...resolvedConfig,
+      ...partialConfig
+    }
```
drortirosh, line 466: better resolveConfiguration return the resolved config, and not set it.
that is, init should do `this.config = await this.resolveConfiguration(this.rawConfig)`

src/relayclient/RelayProvider.ts
```diff
@@ -25,39 +26,36 @@ export type JsonRpcCallback = (error: Error | null, result?: JsonRpcResponse) =>
...
-  constructor (origProvider: HttpProvider, gsnConfig: Partial<GSNConfig>, overrideDependencies?: Partial<GSNDependencies>, relayClient?: RelayClient) {
-    this.host = origProvider.host
-    this.connected = origProvider.connected
+  constructor (
+    relayClient: RelayClient
+  ) {
+    this.relayClient = relayClient
+    this.origProvider = this.relayClient.getUnderlyingProvider()
+    this.host = this.origProvider.host
+    this.connected = this.origProvider.connected
```
drortirosh, line 51: wow. this "connected" is buggy, since we copy the current connected status, and save it.

test/RelayHubPenalizations.test.ts
```diff
@@ -469,44 +475,33 @@ contract('RelayHub Penalizations', function ([_, relayOwner, relayWorker, otherR
...
     async function getDataAndSignatureFromHash (txHash: string, chainId: number): Promise<{ data: string, signature: string }> {
-      // @ts-ignore
-      const rpcTx = await web3.eth.getTransaction(txHash)
-      // eslint: this is stupid how many checks for 0 there are
-      // @ts-ignore
-      // eslint-disable-next-line @typescript-eslint/strict-boolean-expressions
-      if (!chainId && parseInt(rpcTx.v, 16) > 28) {
+      const rpcTx: any = await web3.eth.getTransaction(txHash)
+      const vInteger = parseInt(rpcTx.v, 16)
+      if (chainId == null && vInteger > 28) {
```
drortirosh, line 488: does this work with chainId==null? you throw for vInteger>28 AND for vInteger <=28

test/RelayHubPenalizations.test.ts
```diff
@@ -469,44 +475,33 @@ contract('RelayHub Penalizations', function ([_, relayOwner, relayWorker, otherR
...
 
     async function getDataAndSignatureFromHash (txHash: string, chainId: number): Promise<{ data: string, signature: string }> {
-      // @ts-ignore
-      const rpcTx = await web3.eth.getTransaction(txHash)
-      // eslint: this is stupid how many checks for 0 there are
-      // @ts-ignore
-      // eslint-disable-next-line @typescript-eslint/strict-boolean-expressions
-      if (!chainId && parseInt(rpcTx.v, 16) > 28) {
+      const rpcTx: any = await web3.eth.getTransaction(txHash)
+      const vInteger = parseInt(rpcTx.v, 16)
```
drortirosh, line 487: don't you need to add .replace(/0x/,'') ?

test/common/ContractInteractor.test.ts
```diff
@@ -1,27 +1,28 @@
...
 import { PrefixedHexString } from 'ethereumjs-tx'
 import Transaction from 'ethereumjs-tx/dist/transaction'
 import { constants } from '../../src/common/Constants'
 import { createClientLogger } from '../../src/relayclient/ClientWinstonLogger'
 import RelayRequest from '../../src/common/EIP712/RelayRequest'
 import { deployHub } from '../TestUtils'
+import VersionsManager from '../../src/common/VersionsManager'
+import { gsnRuntimeVersion } from '../../src/common/Version'
 
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
```
drortirosh, line 17: if you don't use expect, don't define it. just do chai.use

test/common/ContractInteractor.test.ts
```diff
@@ -1,27 +1,28 @@
...
 const { expect } = chai.use(chaiAsPromised)
 
-const TestVersions = artifacts.require('TestVersions')
 const TestPaymasterConfigurableMisbehavior = artifacts.require('TestPaymasterConfigurableMisbehavior')
 const StakeManager = artifacts.require('StakeManager')
 const Penalizer = artifacts.require('Penalizer')
 
 contract('ContractInteractor', function (accounts) {
+  /*
```
drortirosh, line 25: whyh comment out skipped tst?

test/dummies/BadContractInteractor.ts
```diff
@@ -1,18 +1,16 @@
...
 
 export default class BadContractInteractor extends ContractInteractor {
   static readonly message = 'This is not the contract you are looking for'
   static readonly wrongNonceMessage = 'the tx doesn\'t have the correct nonce'
 
   private readonly failValidateARC: boolean
 
-  constructor (provider: Web3Provider, logger: LoggerInterface, config: GSNConfig, failValidateARC: boolean) {
-    super(provider, logger, config)
+  constructor (constructorParams: ConstructorParams, failValidateARC: boolean) {
```
drortirosh, line 12: nitpic: if you define named params, then why do you add new unnamed one?
(I know, its only a test. i think you can extend type in-place to add new member. I just don't know how.)

src/relayclient/RelayProvider.ts
```diff
@@ -25,39 +26,36 @@ export type JsonRpcCallback = (error: Error | null, result?: JsonRpcResponse) =>
...
+// TODO: stop faking the HttpProvider implementation -  it won't work for any other 'origProvider' type
+export class RelayProvider implements HttpProvider, ISendProvider {
   protected readonly origProvider: HttpProvider & ISendAsync
   private readonly origProviderSend: any
   protected config!: GSNConfig
 
   readonly relayClient: RelayClient
   logger!: LoggerInterface
 
+  static newProvider (input: GSNUnresolvedConstructorInput): RelayProvider {
```
drortirosh, line 38: the minimal construction becomes:
```
RelayProvider.newProvider({ provider:window.ethereum, partialConfig:{paymasterAddress: '0x1234'}})
```

Do you think its nice?

test/relayclient/RelaySelectionManager.test.ts
```diff
@@ -57,8 +66,15 @@ contract('RelaySelectionManager', function (accounts) {
     let stubGetNextSlice: SinonStub
 
     before(async function () {
+      contractInteractor = await new ContractInteractor({
```
drortirosh, line 69: howc omes you need it now, and didn't in the past?

src/common/types/Aliases.ts
```diff
@@ -26,3 +29,18 @@ export function notNull<TValue> (value: TValue | null | undefined): value is TVa
...
+
+export type Web3Provider =
+  | HttpProvider
+  | IpcProvider
+  | WebsocketProvider
+
+/**
+ * The only thing that is guaranteed a Web3 provider or a similar object is a {@link send} method.
+ */
+export interface ISendProvider {
```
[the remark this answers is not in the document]
forshtat: Done. Actually, `Web3Provider` is already defined in a few places, so to avoid any chance of confusion I named it `Web3ProviderBaseInterface`. Hope this is better 😂

src/common/ContractInteractor.ts
```diff
@@ -151,28 +155,66 @@ export default class ContractInteractor {
...
     // chain === 'private' means we're on ganache, and ethereumjs-tx.Transaction doesn't support that chain type
     this.rawTxOptions = getRawTxOptions(this.chainId, this.networkId, chain)
+    return this
   }
 
-  async getAsyncChainId (): Promise<number> {
-    return await this.web3.eth.getChainId()
+  async _resolveDeployment (): Promise<void> {
+    if (this.deployment.paymasterAddress != null && this.deployment.relayHubAddress != null) {
+      this.logger.error('Already resolved!')
```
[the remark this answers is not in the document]
forshtat: Fixed

src/common/ContractInteractor.ts
```diff
@@ -151,28 +155,66 @@ export default class ContractInteractor {
...
+      return
+    }
+
+    if (this.deployment.paymasterAddress != null) {
+      await this._resolveDeploymentFromPaymaster(this.deployment.paymasterAddress)
+    } else if (this.deployment.relayHubAddress != null) {
+      await this._resolveDeploymentFromRelayHub(this.deployment.relayHubAddress)
+    }
```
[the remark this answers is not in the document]
forshtat: Fixed. As a logger info though, as it is currently perfectly legit to have a ContractInteractor only to query the gas price on chain ID, so this is not an edge case.

src/common/ContractInteractor.ts
```diff
@@ -181,38 +223,27 @@ export default class ContractInteractor {
...
-      }
-      if (hubPenalizerAddress == null || hubPenalizerAddress === constants.ZERO_ADDRESS) {
-        throw new Error(`Penalizer address not set in RelayHub (or threw error: ${getAddressesError?.message})`)
-      }
-      this.stakeManagerInstance = await this._createStakeManager(hubStakeManagerAddress)
-      this.penalizerInstance = await this._createPenalizer(hubPenalizerAddress)
+    if (this.paymasterInstance == null && this.deployment.paymasterAddress != null) {
+      this.paymasterInstance = await this._createPaymaster(this.deployment.paymasterAddress)
+    }
+    if (this.deployment.forwarderAddress != null) {
```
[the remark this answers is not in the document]
forshtat: Unfortunately, these `_createX` methods are currently abused by other classes, so I had to create a separate issue:
https://opengsn.atlassian.net/browse/OG-324

src/common/ContractInteractor.ts
```diff
@@ -486,9 +511,24 @@ export default class ContractInteractor {
...
-    return await hub.balanceOf(managerAddress)
+  /**
+   * Gets balance of an address on the current RelayHub.
+   * @param address - can be a Paymaster or a Relay Manger
+   */
+  async hubBalanceOf (address: Address): Promise<BN> {
+    return await this.relayHubInstance.balanceOf(address)
+  }
+
+  async setDeployment (deployment: GSNContractsDeployment): Promise<void> {
```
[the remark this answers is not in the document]
forshtat: Fixed.
It will probably work with replacing a deployment, but of course nobody should do that.

src/common/ContractInteractor.ts
```diff
@@ -550,6 +590,10 @@ export default class ContractInteractor {
       })
     })
   }
+
+  async depositToRelayHubForAddress (paymaster: Address, transactionDetails: TransactionDetails): Promise<any> {
```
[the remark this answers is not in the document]
forshtat: Fixed.

src/relayclient/RelayClient.ts
```diff
@@ -62,33 +74,21 @@ export interface RelayingResult {
...
-   */
   constructor (
-    provider: HttpProvider,
-    configOverride: Partial<GSNConfig>,
-    overrideDependencies?: Partial<GSNDependencies>
+    rawConstructorInput: GSNUnresolvedConstructorInput
   ) {
-    this.partialConfig = { provider, configOverride, overrideDependencies }
-    this.logger = console
+    this.rawConstructorInput = rawConstructorInput
```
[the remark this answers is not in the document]
forshtat: Added

src/relayclient/RelayClient.ts
```diff
@@ -174,37 +160,37 @@ export class RelayClient {
...
-    return false
+    return pendingBlock.transactions.includes(txHash)
   }
 
   async relayTransaction (gsnTransactionDetails: GsnTransactionDetails): Promise<RelayingResult> {
     if (!this.initialized) {
       if (this.initializingPromise == null) {
         this._warn('suggestion: call RelayProvider.init()/RelayClient.init() in advance (to make first request faster)')
       }
+      if (this.rawConstructorInput == null) {
```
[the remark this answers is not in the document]
forshtat: Removed the check entirely. Passing no constructor params is an obvious mistake.

src/relayclient/RelayClient.ts
forshtat: Fixed. We are having some conflicts here as we still have to cast to `HttpProvider` when passing the `RelayProvider` to the Web3 constructor.

src/relayclient/RelayClient.ts
forshtat: Separated them, no problem.

src/relayclient/RelayClient.ts
forshtat: Cool. Fixed.

src/relayclient/RelayProvider.ts
forshtat: It was like that, and unfortunately not much we can do here

test/RelayHubPenalizations.test.ts
forshtat: This is the exact translation of the old code without the ts-ignore. I don't know if it works but it is only used in this test file so I am not concerned with it at all.

test/RelayHubPenalizations.test.ts
forshtat: Somehow it does work. This is a test though.

test/common/ContractInteractor.test.ts
forshtat: Sorry, did not commit a  part of a file. This eslint-disable is completely redundant

test/common/ContractInteractor.test.ts
forshtat: Reintroduced this test now. See "should throw if wrong contract paymaster version".

forshtat: If I understand correctly the version we care about is the Paymaster version now, not the Hub version. At least there was no place where we would've requested the version form the hub.

test/dummies/BadContractInteractor.ts
forshtat: This file has to go and be replaced by sinon mock objects, I just had no time to get to it

test/relayclient/RelaySelectionManager.test.ts
forshtat: It used to rely on `getDependencies` method being called outside the `before` block, now there is no such method so I had to create the dependency manually.

test/common/ContractInteractor.test.ts
drortirosh: yes. there should be (offline?) test that the paymaster matches the hub. its more of a job for the paymaster itself (or paymaster's owner), not the client's

src/relayclient/RelayProvider.ts
drortirosh: I mean, from the user's perspective, it should be :
```
RelayProvider.newProvider({ provider:window.ethereum, paymasterAddress: '0x1234'})
```
the 2nd nesting of "partialConfig" is completely useless (I don't care about overrideDependencies, since that's something for testing only)

src/relayclient/RelayProvider.ts
drortirosh: created https://opengsn.atlassian.net/browse/OG-326

test/relayclient/KnownRelaysManager.test.ts
```diff
@@ -76,12 +76,15 @@ contract('KnownRelaysManager', function (
...
-        logLevel: 'error',
-        relayHubAddress: relayHub.address,
+        loggerConfiguration: { logLevel: 'error' },
         relayLookupWindowBlocks
       })
-      logger = createClientLogger(config.logLevel, config.loggerUrl, config.loggerApplicationId, config.loggerUserIdOverride)
-      contractInteractor = new ContractInteractor(web3.currentProvider as HttpProvider, logger, config)
+      logger = createClientLogger(config.loggerConfiguration)
+      contractInteractor = new ContractInteractor({
+        provider: web3.currentProvider as HttpProvider,
```
drortirosh, line 84: nitpic: I think ContractInteractor should be fixed, so that it accept web3.currentProvider without casting (after all, that's one of its main usages)

test/relayclient/KnownRelaysManager.test.ts
```diff
@@ -184,16 +190,20 @@ contract('KnownRelaysManager 2', function (accounts) {
       penalizer = await Penalizer.new()
       relayHub = await deployHub(stakeManager.address, penalizer.address)
       config = configureGSN({
```
drortirosh, line 192: lost you here. you have both delpoyment and config, and it seems you don't use the config.

test/relayclient/RelaySelectionManager.test.ts
```diff
@@ -57,8 +66,15 @@ contract('RelaySelectionManager', function (accounts) {
...
+        provider: web3.currentProvider as HttpProvider,
+        logger
+      }).init()
+      httpClient = new HttpClient(new HttpWrapper(), this.logger)
```
drortirosh, line 73: different logger? why "this" ?

test/relayclient/RelaySelectionManager.test.ts
drortirosh: then we're probably missing "createDefaultDependencies()" method.

src/relayclient/RelayClient.ts
```diff
@@ -62,33 +73,25 @@ export interface RelayingResult {
...
   ) {
-    this.partialConfig = { provider, configOverride, overrideDependencies }
-    this.logger = console
+    // TODO: backwards-compatibility 102 - remove on next version bump
+    if (arguments[0] == null || arguments[0].send != null || arguments[2] != null) {
+      throw new Error('Sorry, but the constructor parameters of the RelayClient class have changed. See "GSNUnresolvedConstructorInput" interface for details.')
+    }
+    this.rawConstructorInput = rawConstructorInput
+    this.logger = rawConstructorInput.overrideDependencies?.logger ??
+      createClientLogger(rawConstructorInput.partialConfig?.loggerConfiguration ?? defaultLoggerConfiguration)
```
drortirosh, line 92: you override configuration in "all or nothing" manner: unable to change a single field

src/relayclient/RelayProvider.ts
forshtat: Well, I strongly suggest creating/reading the configuration into a variable(s) before passing it to the method. 
I write with this mentality, and yes, different code styles will vary in how well it looks.

```
const provider = window.ethereum
const partialConfig: Partial<GSNConfig> = { payamsterAddress: '0x123123' }
const relayProvider: RelayProvider = RelayProvider.newProvider({ provider, partialConfig })
```

I feel this code is way more explicit, readable and easier to work with.

test/relayclient/RelaySelectionManager.test.ts
forshtat: It is not that widely needed so I think we can get away with a few more manual constructor calls...

src/cli/CommandsLogic.ts
```diff
@@ -294,26 +299,27 @@ export default class CommandsLogic {
...
-    return {
+    this.deployment = {
+      paymasterVersion,
       relayHubAddress: rInstance.options.address,
       stakeManagerAddress: sInstance.options.address,
       penalizerAddress: pInstance.options.address,
       forwarderAddress: fInstance.options.address,
       versionRegistryAddress: regInstance.options.address,
-      naivePaymasterAddress: paymasterAddress
+      paymasterAddress: pmInstance?.options.address ?? constants.ZERO_ADDRESS
```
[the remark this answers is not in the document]
forshtat: Ok, so there was an interface called 'DeploymentResult', and it could only be generated by the `gsn-deploy` command so it made sense for it to be called 'naivePaymaster'. Now I have turned this interface to a more widely used 'GSNContractsDeployment' so it stopped making sense to have two basically identical interfaces with one having a special name for a paymaster it deploys...

src/common/types/Aliases.ts
```diff
@@ -26,3 +29,19 @@ export function notNull<TValue> (value: TValue | null | undefined): value is TVa
...
+export type Web3Provider =
+  | HttpProvider
+  | IpcProvider
+  | WebsocketProvider
+
+/**
+ * The only thing that is guaranteed a Web3 provider or a similar object is a {@link send} method.
+ */
+export interface Web3ProviderBaseInterface {
+  isMetaMask?: boolean
```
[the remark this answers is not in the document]
forshtat: I mean 'base' for us inside GSN - what we do expect a provider has. Nullable fields are optional, so IMO it is ok to have provider-specific fields as optional. What do you think I should do? Create a family of interfaces for each provider and check which type the given `origProvider` is?

src/relayclient/ClientWinstonLogger.ts
```diff
@@ -55,18 +56,19 @@ export function createClientLogger (level: NpmLogLevel, loggerUrl: string, appli
...
+  if (loggerConfiguration.userId != null) {
+    userId = loggerConfiguration.userId
   } else {
     userId = getOrCreateUserId()
   }
 
   const localhostRegExp: RegExp = /http:\/\/(localhost)|\b\d{1,3}\.\d{1,3}\.\d{1,3}\.\d{1,3}\b/
-  if (applicationId.length === 0 && typeof window !== 'undefined' && window.location != null && window.location.href != null && window.location.href.match(localhostRegExp) == null) {
+  let applicationId = loggerConfiguration.applicationId
+  if (loggerConfiguration.applicationId == null && typeof window !== 'undefined' && window.location != null && window.location.href != null && window.location.href.match(localhostRegExp) == null) {
```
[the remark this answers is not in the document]
forshtat: fixed

test/relayclient/KnownRelaysManager.test.ts
forshtat: web3.currentProvider is defined as this:
export type provider =
    | HttpProvider
    | IpcProvider
    | WebsocketProvider
    | string
    | null;

I have no idea where did it make sense for Web3 to allow both 'string' and 'null' as a valid provider type, but I don't want to allow these in the C.I. constructor.

test/relayclient/KnownRelaysManager.test.ts
forshtat: Config only has a 'prefferedRelays' set. I did not get too deep into the code as this is simply a refactor.


See here: 
KnownRelaysManager.test.ts:208:7
knownRelaysManager = new KnownRelaysManager(contractInteractor, logger, **config**)

test/relayclient/RelaySelectionManager.test.ts
forshtat: A mistake. It did not crash anything because the client's methods are immediately stubbed.

src/relayclient/RelayClient.ts
forshtat: There is nothing in the defaultLoggerConfiguration worth overriding.

drortirosh (approved):

PR #543, opened 2020-12-09
MVP gsn status revamp

forshtat:


forshtat:
A jira issue to implement the remaining vision is here:
https://opengsn.atlassian.net/browse/OG-383

This is a better-than-nothing MVP for network status tool.

package.json
```diff
@@ -57,6 +58,7 @@
     "@truffle/hdwallet-provider": "1.0.34",
     "@types/chai": "^4.2.12",
     "@types/chai-as-promised": "^7.1.3",
+    "@types/cli-table": "^0.3.0",
```
drortirosh, line 61: duplicate: you also add it as a devDependency
(and anyway ,need to remove many deps since they are devDeps)

drortirosh (changes requested):
(review from running the app, not from the code)
1. wrong (or no) hub address gives stack dump, with no obvious reason (the cause - there is a local "build/gsn" file, but I looked at kovan. need explanation
2. took nearly 30 seconds with no visible response, not even "gathering data"
3. default should not attempt to "gather everything" - especially if it takes that long
4. output format is terrible. see atached picture:
image
    -  not sure what "Transactions" (with colored "accepted/rejected") means
    -  table too wide to fit whatever screen you're using - 
    - the table itself is far, far too long - it is supported to be all relayers, but there are ~100 entries there, so there's some mistake
    - more likely: you're listing all relayers that ever registered, but ignore any "active window"
    - 

5. usability: its problematic that you have to specify hub address, but we currently don't have better option (maybe we should put a json file of all "official" RelayHubs, and the utility will reference that by default)
6. short addresses are good for display, but useless for further work. must have "full" mode, to dump full addresses (and probably, with everything in json format)
7. maybe entire tool should be split into 2:
  - generate data (one big json, but with flags if you don't want everything, only partial)
  - format data (again, can select portion of selected data, whether full/short addresses)

Currently it is not an MVP - its a pile of too much irrelevant data. better be "minimal", in order to be "viable"

src/common/types/GSNContractsDataTypes.ts
```diff
@@ -0,0 +1,121 @@
...
+  relayUrl: string
+}
+
+export interface RelayRegisteredEventInfo extends RelayInfoUrl, GNSContractsEvent {
+  relayManager: Address
+  baseRelayFee: IntString
+  pctRelayFee: IntString
+}
+
+export interface TransactionRelayedEventInfo extends GNSContractsEvent {
```
drortirosh, line 31: what's GNS ?

src/common/types/GSNStatistics.ts
```diff
@@ -0,0 +1,125 @@
...
+  hubUnauthorizedEvents: Array<EventTransactionInfo<HubUnauthorizedEventInfo>>
+}
+
+export interface RelayServerRegistrationInfo {
+  lastRegisteredUrl: string
+  lastRegisteredBaseFee: IntString
+  lastRegisteredPctFee: IntString
+  pingResult: PingResult
+  registeredWorkers: Address[]
+  workerBalances: ObjectMap<IntString>
```
drortirosh, line 61: not sure what type is this.
but workers can be held as <address,intString>. no need for array at all (workersMap.keys() returns just the addresses)
actually, value can be a small struct, currently contains  "balance", and in the future might have more data.

src/common/types/GSNStatistics.ts
```diff
@@ -0,0 +1,125 @@
...
+  stakePenalizedEvents: Array<EventTransactionInfo<StakePenalizedEventInfo>>
+  hubAuthorizedEvents: Array<EventTransactionInfo<HubAuthorizedEventInfo>>
+  hubUnauthorizedEvents: Array<EventTransactionInfo<HubUnauthorizedEventInfo>>
+}
+
+export interface RelayServerRegistrationInfo {
+  lastRegisteredUrl: string
+  lastRegisteredBaseFee: IntString
+  lastRegisteredPctFee: IntString
+  pingResult: PingResult
```
drortirosh, line 59: ping is the only field that is "last" ... url, fee, pct are truly the last value, but are much less temporary than the ping result..
I think the "last" prefix should not be used, as it doesn't add anything, just confusing.

src/common/types/GSNStatistics.ts
```diff
@@ -0,0 +1,125 @@
...
+  address: Address
+  transactionCount: number
+}
+
+export interface SenderInfo {
+  address: Address
+  transactionCount: number
+}
+
+export interface RelayHubConstructorParams {
```
drortirosh, line 81: this was named `RelayHubConfiguration`. why duplicate same type?

src/common/statistics/StatisticsManager.ts
```diff
@@ -0,0 +1,394 @@
...
+      paymasters,
+      senders,
+      recipients,
+      relayServers,
+      totalGasPaidViaGSN
+    }
+  }
+
+  async fetchRelayHubEvents (): Promise<void> {
+    const transactionDepositedEventsData =
```
drortirosh, line 109: I lost you here: you have a nice `extractTransactionInfos()` to filter out specific type from "all events" - but still, you do getPastEventsForHub for each event separately...
do gpefh once, and filter its output 5 times
(ok, if you really want, then split the TransactionRelayed out  - but all the rest can be slurped together)

src/common/statistics/StatisticsManager.ts
```diff
@@ -0,0 +1,394 @@
...
+        address,
+        relayHubBalance,
+        acceptedTransactionsCount,
+        rejectedTransactionsCount
+      })
+    }
+    return paymasterInfos
+  }
+
+  extractUnique (events: Array<EventTransactionInfo<StakeChangeEvent>>): Address[] {
```
drortirosh, line 163: extractUniqueRelayers

drortirosh: and please, return a Set, not array.
the "`.includes()`" you use is O(n), and `set.has()` is O(1), and you do it in a loop.

src/common/statistics/StatisticsManager.ts
```diff
@@ -0,0 +1,394 @@
...
+
+    const relayHubEvents: RelayHubEvents = {
+      relayRegisteredEvents,
+      transactionRelayedEvents,
+      transactionRejectedEvents
+    }
+
+    const isRegistered = stakeStatus === RelayServerStakeStatus.STAKE_LOCKED && relayRegisteredEvents.length !== 0
+    const relayHubEarningsBalance = (await this.contractInteractor.hubBalanceOf(managerAddress)).toString()
+    const stakeInfo = await this.contractInteractor.stakeManagerStakeInfo(managerAddress)
```
drortirosh, line 243: AFAIK, you collected all info from all events, ever. now you do direct calls to collect the info again from the hub..

src/common/statistics/StatisticsManager.ts
```diff
@@ -0,0 +1,394 @@
...
+    const ownerBalance = await this.contractInteractor.getBalance(stakeInfo.owner)
+    const managerBalance = await this.contractInteractor.getBalance(managerAddress)
+    if (isRegistered) {
+      const lastRegisteredUrl = relayRegisteredEvents[relayRegisteredEvents.length - 1].returnValues.relayUrl
+      const lastRegisteredBaseFee = relayRegisteredEvents[relayRegisteredEvents.length - 1].returnValues.baseRelayFee
+      const lastRegisteredPctFee = relayRegisteredEvents[relayRegisteredEvents.length - 1].returnValues.pctRelayFee
+      const pingResult = await this.attemptPing(lastRegisteredUrl)
+      const registeredWorkers: Address[] = await this.contractInteractor.getRegisteredWorkers(managerAddress)
```
drortirosh, line 251: this was also collected before through events.

src/common/statistics/StatisticsManager.ts
```diff
@@ -0,0 +1,394 @@
...
+      relayHubEarningsBalance,
+      relayHubEvents,
+      registrationInfo,
+      stakeManagerEvents
+    }
+  }
+
+  getStakeManagerEvents (managerAddress?: Address): StakeMangerEvents {
+    const stakeAddedEvents =
+      this.allStakeManagerEvents.stakeAddedEvents.filter(it => it.returnValues.relayManager === managerAddress)
```
drortirosh, line 282: it aches me in the eye... 
why not:

```js
const mgrFilter = it => it.returnValues.relayManager === managerAddress

return {
  allEvents: this.allStakeManagerEvents.allEvents.filter(mgrFilter),
  stakeAddedEvents: this.allStakeManagerEvents.stakeAddedEvents.filter(mgrFilter),
  ...
}
```

src/common/statistics/StatisticsManager.ts
```diff
@@ -0,0 +1,394 @@
...
+      stakePenalizedEvents,
+      hubAuthorizedEvents,
+      hubUnauthorizedEvents
+    }
+  }
+
+  async attemptPing (url: string): Promise<PingResult> {
+    let relayPing: PingResult
+    try {
+      const pingResponse = await this.httpClient.getPingResponse(url)
```
drortirosh, line 330: must be a way to handle timeouts (yes, not here, at the API level of this class)

src/common/statistics/StatisticsManager.ts
```diff
@@ -0,0 +1,394 @@
...
+        eventData => {
+          return {
+            eventData,
+            returnValues: eventData.returnValues as T
+          }
+        }
+      )
+  }
+
+  async getAuthorizedHubsAndVersions (stakeManagerEvents: StakeMangerEvents): Promise<ObjectMap<SemVerString>> {
```
drortirosh, line 351: this is an overkill - and also performance issue: for each entity (relayer), you query all relayhubs it is registered for.
Also, you're stats is for specific hub - so why bother collect into on all hubs
(or vice-versa: if you're collect for all hubs, then start by collecting general hubs info, and then you don't have to query it here)

src/cli/CommandLineStatisticsPresenterConfig.ts
```diff
@@ -0,0 +1,18 @@
+export interface CommandLineStatisticsPresenterConfig {
```
drortirosh, line 1: deliberately DIDNT look into the presenter/cmdline stuff: they don't belong to this huge PR. please split)

shahafn: why not?

src/common/statistics/StatisticsManager.ts
shahafn: What do you mean? errors are caught in this code, do you mean that httpClient should handle timeouts inside `getPingResponse` function? If so, it shouldn't be part of this PR.

drortirosh: It waits for all relayers, meaning a dead relayer can make the API dead for a long time. As a user (eg admin that runs the cli) I need to be able to select hours much time to wait. Right now, there's no such parameter

package.json
forshtat: not relevant after 'workspaces' PR

src/cli/CommandLineStatisticsPresenterConfig.ts
forshtat: cool. don't.

src/common/ContractInteractor.ts
```diff
@@ -439,11 +425,11 @@ export default class ContractInteractor {
     return await this._getPastEvents(stakeManager.contract, names, extraTopics, options)
   }
 
-  // eslint-disable-next-line @typescript-eslint/require-await
   async _getPastEvents (contract: any, names: EventName[], extraTopics: string[], options: PastEventOptions): Promise<EventData[]> {
     const topics: string[][] = []
     const eventTopic = event2topic(contract, names)
     topics.push(eventTopic)
+    // TODO: AFAIK this means only the first parameter of the event is supported
```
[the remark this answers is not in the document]
forshtat: No, it is not. this code does **not** allow you to filter events by the 2nd, 3rd of 4th indexed parameter. All values passed as 'extraTopics' will be seen as "OR" for a first indexed parameter.

This is a "known" issue but we do not need to filter on other parameters anywhere in the code so far.

src/common/Utils.ts
```diff
@@ -168,24 +165,22 @@ export function randomInRange (min: number, max: number): number {
   return Math.floor(Math.random() * (max - min) + min)
 }
 
-export function isSecondEventLater (a: EventData, b: EventData): boolean {
+export function eventsComparator (a: EventData, b: EventData): number {
```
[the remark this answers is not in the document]
forshtat: Will do (separate PR)

src/common/types/GSNContractsDataTypes.ts
forshtat: will fix (separate PR)

src/common/types/GSNStatistics.ts
forshtat: Will remove

src/common/types/GSNStatistics.ts
forshtat: Ok, ObjectMap is just an object, but yes, it can be a single map. Will fix.

src/common/types/GSNStatistics.ts
forshtat: My mistake. Will fix.

drortirosh (approved):

PR #544, opened 2020-12-10
OG-349 Shortcut: disable storage for Reputation Manager (and OG-331)

forshtat:
This will prevent the bug where paymaster reputation is permanently
affected by the same accepted/rejected event on every server restart.

This should also prevent crash on update to version >2.0.3 (OG-331)

### DIFF
--- a/src/relayserver/RelayServer.ts
+++ b/src/relayserver/RelayServer.ts
@@ -108,7 +108,8 @@ export class RelayServer extends EventEmitter {
 
   async pingHandler (paymaster?: string): Promise<PingResponse> {
     if (this.config.runPaymasterReputations && paymaster != null) {
-      const status = await this.reputationManager.getPaymasterStatus(paymaster)
+      const currentBlockNumber = await this.contractInteractor.getBlockNumber()
+      const status = await this.reputationManager.getPaymasterStatus(paymaster, currentBlockNumber)
       if (status === PaymasterStatus.BLOCKED || status === PaymasterStatus.ABUSED) {
         throw new Error(`This paymaster will not be served, status: ${status}`)
       }
@@ -169,7 +170,8 @@ export class RelayServer extends EventEmitter {
   }
 
   async validatePaymasterReputation (paymaster: Address): Promise<void> {
-    const status = await this.reputationManager.getPaymasterStatus(paymaster)
+    const currentBlockNumber = await this.contractInteractor.getBlockNumber()
+    const status = await this.reputationManager.getPaymasterStatus(paymaster, currentBlockNumber)
     if (status === PaymasterStatus.GOOD) {
       return
     }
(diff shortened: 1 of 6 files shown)


src/relayserver/ReputationManager.ts
```diff
@@ -100,9 +133,36 @@ export class ReputationManager {
       .filter(it => Date.now() - it.timestamp < this.config.abuseTimeWindowMs)
       .reduce((previousValue: number, currentValue: ReputationChange) => previousValue + currentValue.change, 0)
     if (-changeInAbuseWindow >= this.config.abuseReputationChange) {
+      /*
       await this.reputationStoreManager.setAbuseFlag(paymaster)
+       */
+      entry.abuseStartedTs = Date.now()
```
drortirosh, line 139: this is bad: when loading a batch, all are initialized to "now".
should use block-based reputations (to avoid reading the timestamps of all blocks.
we can "calibrate" the block time at startup, using
```js
async averageBlockTime() {
  latest = await getBlock('latest')
  if ( latest.number<1000 ) return 1; //ganache: no average time..
  older = await getBlock(latest.number-1000)
  blockTimeMs = (latest.timestamp - older.timestamp)
  return blockTimeMs
}
```

forshtat: Well
a) reputation abuse mechanism was defined as time-based and does not require block times to function
b) this PR does not modify the logic inside the mechanism, but disconnects it from persistent storage. 

We can discuss change to the Reputation Mechanism separately, but that will not be a blocker by default.

src/relayserver/ReputationManager.ts
```diff
@@ -82,27 +96,81 @@ export class ReputationManager {
...
     }
-    const oldChangesExpirationTs = Date.now() - this.config.abuseTimeWindowMs
+    const oldChangesExpirationBlock = eventBlockNumber - this.config.abuseTimeWindowBlocks
+    /*
     await this.reputationStoreManager.updatePaymasterReputation(paymaster, change, oldChangesExpirationTs)
+     */
+    this.updatePaymasterReputation(paymaster, change, oldChangesExpirationBlock, eventBlockNumber)
+  }
+
+  updatePaymasterReputation (
```
drortirosh, line 152: why this method is duplicated?

forshtat: One is in the Storage Manager. I do not remove the persistent storage code yet as I believe we will reenable it quite soon. 
The other is "re-implementation" with a `Map` instance inside the Reputation Manager.

src/relayserver/RelayServer.ts
```diff
@@ -624,18 +624,21 @@ latestBlock timestamp   | ${latestBlock.timestamp}
...
+      await this.reputationManager.updatePaymasterStatus(paymaster, true, eventBlockNumber)
     }
   }
 
-  async _handleTransactionRejectedByPaymasterEvent (paymaster: Address, blockNumber: number): Promise<void> {
-    this.alerted = true
-    this.alertedBlock = blockNumber
-    this.logger.error(`Relay entered alerted state. Block number: ${blockNumber}`)
+  async _handleTransactionRejectedByPaymasterEvent (paymaster: Address, currentBlockNumber: number, eventBlockNumber: number): Promise<void> {
+    // TODO: do not call this method when events are processed already (stateful server thing)
```
drortirosh, line 634: comment irrelevant.

src/relayserver/RelayServer.ts
```diff
@@ -624,18 +624,21 @@ latestBlock timestamp   | ${latestBlock.timestamp}
...
+    if (currentBlockNumber - eventBlockNumber < 30) {
```
drortirosh, line 635: what's this "30" ?

forshtat: removed

src/relayserver/RelayServer.ts
forshtat: I'll keep it because I want to stop calling 'handes' for past events

drortirosh (approved):

PR #545, opened 2020-12-10
OG-234 Make 'sleep' for alerted mode be the first thing done

forshtat:
The purpose of the random 'sleep' is to make it harder for attackers to
trick the relays into broadcasting the transaction that would otherwise
fail validation. If sleep is done after the broadcst, it has no effect.

### DIFF
--- a/src/relayserver/RelayServer.ts
+++ b/src/relayserver/RelayServer.ts
@@ -275,6 +275,10 @@ returnValue        | ${viewRelayCallRet.returnValue}
     if (!this.isReady()) {
       throw new Error('relay not ready')
     }
+    if (this.alerted) {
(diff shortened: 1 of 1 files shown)


src/relayserver/RelayServer.ts
```diff
@@ -275,6 +275,10 @@ returnValue        | ${viewRelayCallRet.returnValue}
     if (!this.isReady()) {
       throw new Error('relay not ready')
     }
+    if (this.alerted) {
```
drortirosh, line 278: not really sure if this should be before all static (non-async) parameter validation, but also it doesn't really matter.

drortirosh (approved):

PR #547, opened 2020-12-11
OG-334 Client choke on gas limit calculation on mainnet

forshtat:
We could introduce a more advanced code to calculate external gas limit,
but for the purposes of the view function passing the worker balance if
it is lower then block gas limit sounds fine.


drortirosh:
LGTM, but no test...

drortirosh (approved):

PR #551, opened 2020-12-12
OG-331 Relayer's reputation manager throws exception after restart

forshtat:


drortirosh (approved):
tested on relayer locally (SHOULD have CI test..)

PR #553, opened 2020-12-14
OG-361 remove: "beta ignore version compatibility"

drortirosh:
add explicit required version range (relevant when actual version
differs by more than patch version)

### DIFF
--- a/src/common/ContractInteractor.ts
+++ b/src/common/ContractInteractor.ts
@@ -39,7 +39,7 @@ import { Address, IntString, Web3ProviderBaseInterface } from './types/Aliases'
 import GsnTransactionDetails from './types/GsnTransactionDetails'
 
 import { Contract, TruffleContract } from '../relayclient/LightTruffleContract'
-import { gsnRuntimeVersion } from './Version'
+import { gsnRequiredVersion, gsnRuntimeVersion } from './Version'
 import Common from 'ethereumjs-common'
 import { GSNContractsDeployment } from './GSNContractsDeployment'
 import TransactionDetails = Truffle.TransactionDetails
@@ -106,7 +106,7 @@ export default class ContractInteractor {
       deployment = {}
     }: ConstructorParams) {
     this.logger = logger
-    this.versionManager = versionManager ?? new VersionsManager(gsnRuntimeVersion)
+    this.versionManager = versionManager ?? new VersionsManager(gsnRuntimeVersion, gsnRequiredVersion)
     this.web3 = new Web3(provider as any)
     this.deployment = deployment
     this.provider = provider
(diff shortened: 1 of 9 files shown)


src/common/ContractInteractor.ts
```diff
@@ -217,7 +217,7 @@ export default class ContractInteractor {
   }
 
   async _validateCompatibility (): Promise<void> {
-    if (this.deployment == null) {
+    if (this.deployment == null || this.relayHubInstance == null) {
```
forshtat, line 220: how would you have an `instance` if the `deployment` is `null`? I think we initialize the instances from the deployment, don't we?

src/relayclient/GSNConfigurator.ts
```diff
@@ -36,6 +36,7 @@ export const defaultGsnConfig: GSNConfig = {
   sliceSize: 3,
   relayTimeoutGrace: DEFAULT_RELAY_TIMEOUT_GRACE_SEC,
   methodSuffix: '',
+  requiredVersionRange: '^2.0.0',
```
forshtat, line 39: Why would we need to make it overridable though?
Among all configurable fields, at least defining the required version is kind of our responsibility IMO.

src/relayclient/RelayClient.ts
```diff
@@ -398,7 +398,7 @@ export class RelayClient {
         this.logger.info(`Audit call failed for relay at URL: ${auditor}. Failed audit calls: ${failedAuditorsCount}/${auditors.length}`)
       }
     }
-    if (auditors.length === failedAuditorsCount) {
+    if (auditors.length === failedAuditorsCount && failedAuditorsCount !== 0) {
```
forshtat, line 401: I think it is better to check `this.config.auditorsCount === 0` cause otherwise there will be no log indicating that, i.e. there was only 1 relay so we could not audit a transaction with any other relay, etc.

forshtat (changes requested):

src/common/ContractInteractor.ts
drortirosh: added because it crashed.. we create ContractInteractor without any RH, but still call validateCompatibility on it. it used to pass because the "beta ignore" ignored also "can't find versionHub of undefined"

src/relayclient/GSNConfigurator.ts
drortirosh: possible future bugs, or troubleshooting.
I'm don't want to document it, since its basically internal (much like the relayLookupWindowBlocks)
there still a possibility that a client that wasn't supposed to support a contract, in some edge cases, is still usable.

src/relayclient/RelayClient.ts
drortirosh: If both are the same, there is no difference if I check either for nonzero.
I just wanted to remove "all zero auditors failed" log..

forshtat: well, it is not same. you are comparing actual auditors array. it is NOT GUARANTEED to be equal in size to `config.auditorsCount`

drortirosh: the `getAuditors()` method returns at most `config.auditorsCount` elements
so this code only has to filter out the "all [zero] auditors failed" log

forshtat (approved):

PR #555, opened 2020-12-14
Remove test publish

drortirosh:


scripts/prepublish
```diff
@@ -25,3 +25,4 @@ done
 tsc
 
 cp -r types dist/
+rm -rf dist/test
```
forshtat, line 28: I am not sure this is required. 
Did you try to configure it through 'package.json'?

forshtat (changes requested):

scripts/prepublish
drortirosh: well, it can't be done with package.json - but it can be done with .npmignore: its an ignore file with all the rules of .gitignore, so we can include/exclude at will - but the language is reversed...

forshtat (approved):

PR #561, opened 2020-12-21
Enforcing max gas price for client requests w.i.p OG-364

shahafn:


forshtat:
@shahafn please refer to the Jira issue number in the title :-)

src/relayserver/RelayServer.ts
```diff
@@ -140,9 +140,14 @@ export class RelayServer extends EventEmitter {
...
     // Check that the gasPrice is initialized & acceptable
-    if (this.gasPrice > parseInt(req.relayRequest.relayData.gasPrice)) {
+    const requestGasPrice = parseInt(req.relayRequest.relayData.gasPrice)
+    if (this.minGasPrice > requestGasPrice) {
       throw new Error(
-        `Unacceptable gasPrice: relayServer's gasPrice:${this.gasPrice} request's gasPrice: ${req.relayRequest.relayData.gasPrice}`)
+        `gasPrice too low: relayServer's gasPrice:${this.minGasPrice} request's gasPrice: ${requestGasPrice}`)
+    }
+    if (parseInt(this.config.maxGasPrice) < requestGasPrice) {
+      throw new Error(
```
forshtat, line 149: please
1. update comment
2. update error text messages

src/relayserver/RelayServer.ts
```diff
@@ -540,8 +545,8 @@ latestBlock timestamp   | ${latestBlock.timestamp}
 
   async _refreshGasPrice (): Promise<void> {
     const gasPriceString = await this.gasPriceFetcher.getGasPrice()
-    this.gasPrice = Math.floor(parseInt(gasPriceString) * this.config.gasPriceFactor)
-    if (this.gasPrice === 0) {
+    this.minGasPrice = Math.floor(parseInt(gasPriceString) * this.config.gasPriceFactor)
```
forshtat, line 548: please add a check here that there is a sane gap between a new `minGasPrice` and a hard-capped `maxGasPriced`.

I think if the difference is small it should emit a warning, if max < min - crash or something, I don't know (it will not be able to relay anything)

forshtat (changes requested):

src/relayserver/RelayServer.ts
drortirosh: how about range check, e.g.

`expected gas price X to be in range YYY .. ZZZ`

shahafn: fixed

src/relayserver/RelayServer.ts
shahafn: fixed

forshtat (approved):

PR #564, opened 2021-01-07
Blacklisted paymaster OG-351 wip

shahafn:


src/relayserver/RelayServer.ts
```diff
@@ -49,7 +49,7 @@ export class RelayServer extends EventEmitter {
   lastSuccessfulRounds = Number.MAX_SAFE_INTEGER
   readonly managerAddress: PrefixedHexString
   readonly workerAddress: PrefixedHexString
-  gasPrice: number = 0
+  minGasPrice: number = 0
```
drortirosh, line 52: seems like it appeared in another issue..

forshtat (approved):

PR #570, opened 2021-01-27
OG-394: forwarder frontrunning

drortirosh:


contracts/forwarder/Forwarder.sol
```diff
@@ -59,8 +59,10 @@ contract Forwarder is IForwarder {
         _verifySig(req, domainSeparator, requestTypeHash, suffixData, sig);
         _updateNonce(req);
 
+        bytes memory callData = abi.encodePacked(req.data, req.from);
+        require( gasleft()*63/64 >= req.gas, "FWD: insufficient gas" );
```
forshtat, line 63: in the Hub we have a `gas reserve`, I think we might want to consider it for Forwarder as well if we cannot be sure that 1/64 of gas limit will be enough to make a `transfer` and a `return` .

contracts/forwarder/Forwarder.sol
```diff
@@ -70,7 +72,7 @@ contract Forwarder is IForwarder {
 
 
     function _verifyNonce(ForwardRequest memory req) internal view {
-        require(nonces[req.from] == req.nonce, "nonce mismatch");
+        require(nonces[req.from] == req.nonce, "FWD: nonce mismatch");
```
forshtat, line 75: If we are adding a prefix, I would add error numbers, what do you think?
Like `Error: FWD-401: nonce mismatch`
would make for easier documentation, searching etc. and can get away with shorter error messages

test/forwarder/Forwarder.test.ts
```diff
@@ -341,7 +341,33 @@ contract('Forwarder', ([from]) => {
       const logs = await recipient.getPastEvents('TestForwarderMessage')
       assert.equal(logs.length, 1, 'TestRecipient should emit')
       assert.equal(logs[0].args.realSender, senderAddress, 'TestRecipient should "see" real sender of meta-tx')
-      assert.equal('1', (await fwd.getNonce(senderAddress)).toString(), 'verifyAndCall should increment nonce')
+      assert.equal('1', (await fwd.getNonce(senderAddress)).toString(), 'execute should increment nonce')
+    })
+
+    it('should revert if not given enough gas', async () => {
```
forshtat, line 347: as forwarder is a critical component, maybe add a call with 'value'
this makes it run a different, more interesting path (see previous comments)

forshtat (changes requested):

contracts/forwarder/Forwarder.sol
drortirosh: we don't really care if for transfer and return succeed - it will revert entire outder transaction.
The attack is only related to failure INSIDE the addr.call(), where only inner transaction is reverted.

drortirosh: Note that the `should revert if not given enough gas` tests validates that even 1 gas too little causes a revert of the outer transaction

contracts/forwarder/Forwarder.sol
drortirosh: string is unique. I did want to scope all forwarder errors with FWD, but I hate numbers - they add nothing.

test/forwarder/Forwarder.test.ts
drortirosh: there is a separate test for value. as I said - a single [outer] wei too little crashes the entire TX.

forshtat (approved):

PR #571, opened 2021-01-27
OG-396 fix msg data

drortirosh:


contracts/BaseRelayRecipient.sol
```diff
@@ -47,19 +47,8 @@ abstract contract BaseRelayRecipient is IRelayRecipient {
...
-                let ptr := mload(0x40)
-                // copy only size-20 bytes
-                let size := sub(calldatasize(),20)
-                // structure RLP data as <offset> <length> <bytes>
-                mstore(ptr, 0x20)
-                mstore(add(ptr,32), size)
-                calldatacopy(add(ptr,64), 0, size)
-                return(ptr, add(size,64))
-            }
+        if (msg.data.length >= 20 && isTrustedForwarder(msg.sender)) {
```
forshtat, line 50: Why did you change 24 to 20?

Do we support calling fallback functions?

contracts/test/TestRecipient.sol
```diff
@@ -77,6 +77,7 @@ contract TestRecipient is BaseRelayRecipient, IKnowForwarderAddress {
     //function with no return value (also test revert with no msg.
     function checkNoReturnValues(bool doRevert) public view {
         (this);
+
```
forshtat, line 80: no big thing, but I try to leave less files changed by a PR according to git history.

test/forwarder/BaseRelayRecipient.test.ts
```diff
@@ -56,4 +56,17 @@ contract('BaseRelayRecipient', ([from]) => {
     assert.equal(await callMsgData(fwd.address, extra + sender.slice(2)), encoded + extra,
       'should extract msg.data if called through trusted forwarder')
   })
+
+  it('should extract msgSender and msgData in transaction', async () => {
+    // trust "from" as forwarder (using real forwarder requires signing
+    const recipient = await TestForwarderTarget.new(from)
+    const encoded = recipient.contract.methods.emitMessage('hello').encodeABI() as string
+    const encodedWithSender = encoded + '1'.repeat(40)
```
forshtat, line 64: oh just use some address, you have 10 of those for free

what if it like gave you address in reverse? would not catch it with 0x111111111111111111111111111111111111111111, you know...

forshtat (changes requested):

contracts/BaseRelayRecipient.sol
drortirosh: GSN doesn't support callback functions, but there is no reason for the forwarder not to.

contracts/test/TestRecipient.sol
drortirosh: ok. will remove.

test/forwarder/BaseRelayRecipient.test.ts
drortirosh: ok

contracts/BaseRelayRecipient.sol
forshtat: Ok. Why this contract is not in a Forwarder folder then? Just asking, if you agree it is part of Forwarder project will add a jira task to put it there.

drortirosh: Right. It should be moved

forshtat (approved):

PR #573, opened 2021-01-31
OG-401 and OG-402 StakeManager voulnerabilities fix

forshtat:
Note: unrelated gas calculations do not pass - reasearch.

### DIFF
--- a/packages/cli/src/CommandsLogic.ts
+++ b/packages/cli/src/CommandsLogic.ts
@@ -31,6 +31,10 @@ import { GSNContractsDeployment } from '@opengsn/common/dist/GSNContractsDeploym
 require('source-map-support').install({ errorFormatterForce: true })
 
 export interface RegisterOptions {
+  /** ms to sleep if waiting for RelayServer to set its owner */
+  sleep: number
+  /** number of times to sleep before timeout*/
+  sleepCount: number
   from: Address
   gasPrice: string | BN
   stake: string | BN
@@ -73,7 +77,10 @@ export default class CommandsLogic {
     deployment: GSNContractsDeployment,
     mnemonic?: string
   ) {
-    let provider: HttpProvider | HDWalletProvider = new Web3.providers.HttpProvider(host)
+    let provider: HttpProvider | HDWalletProvider = new Web3.providers.HttpProvider(host, {
+      keepAlive: true,
+      timeout: 120000
+    })
     if (mnemonic != null) {
       // web3 defines provider type quite narrowly
       provider = new HDWalletProvider(mnemonic, provider) as unknown as HttpProvider
@@ -194,39 +201,7 @@ export default class CommandsLogic {
 
       if (owner !== constants.ZERO_ADDRESS && !isSameAddress(owner, options.from)) {
         // throw new Error(`Already owned by ${owner}, our account=${options.from}`)
-        throw new Error(`Already owned by ${owner}, our account=`)
-      }
-
-      if (toBN(unstakeDelay).gte(toBN(options.unstakeDelay)) &&
-        toBN(stake).gte(toBN(options.stake.toString()))
-      ) {
-        console.log('Relayer already staked')
-      } else {
-        const stakeValue = toBN(options.stake.toString()).sub(toBN(stake))
-        console.log(`Staking relayer ${fromWei(stakeValue, 'ether')} eth`,
-          stake === '0' ? '' : ` (already has ${fromWei(stake, 'ether')} eth)`)
-
-        const stakeTx = await stakeManager
-          .stakeForAddress(relayAddress, options.unstakeDelay.toString(), {
-            value: stakeValue,
-            from: options.from,
-            gas: 1e6,
-            gasPrice: options.gasPrice
-          })
-        transactions.push(stakeTx.tx)
-      }
-
-      if (isSameAddress(owner, options.from)) {
-        console.log('Relayer already authorized')
-      } else {
-        console.log('Authorizing relayer for hub')
-        const authorizeTx = await stakeManager
-          .authorizeHubByOwner(relayAddress, relayHubAddress, {
-            from: options.from,
-            gas: 1e6,
-            gasPrice: options.gasPrice
-          })
-        transactions.push(authorizeTx.tx)
+        throw new Error(`Already owned by ${owner}, our account=${options.from}`)
       }
 
       const bal = await this.contractInteractor.getBalance(relayAddress)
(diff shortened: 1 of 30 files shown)


packages/cli/src/CommandsLogic.ts
```diff
@@ -31,6 +31,10 @@ import { GSNContractsDeployment } from '@opengsn/common/dist/GSNContractsDeploym
 require('source-map-support').install({ errorFormatterForce: true })
 
 export interface RegisterOptions {
+  /** ms to sleep if waiting for RelayServer to set its owner */
+  sleep: number
```
drortirosh, line 35: 1. time should have `Ms` or `Sec` suffix
2. params should be more clear to their target
3. no values for real network.

So its better to have **setOwnerTimoutSec** (1 sec for test, 60sec for real network), and **setOwnerPollSec** (0.1sec on test, 10sec on real network)
The count is not interesting

packages/cli/src/CommandsLogic.ts
```diff
@@ -73,7 +77,10 @@ export default class CommandsLogic {
     deployment: GSNContractsDeployment,
     mnemonic?: string
   ) {
-    let provider: HttpProvider | HDWalletProvider = new Web3.providers.HttpProvider(host)
+    let provider: HttpProvider | HDWalletProvider = new Web3.providers.HttpProvider(host, {
+      keepAlive: true,
```
drortirosh, line 81: thought keepAlive was a default... 
if it isn't, then its important to add - but to real server, not the command-line tool.

packages/cli/src/CommandsLogic.ts
```diff
@@ -194,7 +201,45 @@ export default class CommandsLogic {
...
+        }
+        transactions.push(fundTx.transactionHash)
+      }
+
+      if (owner === constants.ZERO_ADDRESS) {
+        let i = 0
+        while (true) {
+          await sleep(options.sleep)
+          const newStakeInfo = await stakeManager.getStakeInfo(relayAddress)
+          if (isSameAddress(newStakeInfo.owner, options.from)) {
```
drortirosh, line 235: should test for non-zero, non-from.
currently if it is configured with wrong address, you'll have to wait the entire timeout (and then crash on the next owner transaction)

packages/cli/src/commands/gsn-relayer-register.ts
```diff
@@ -24,6 +24,8 @@ const commander = gsnCommander(['n', 'f', 'm', 'g'])
   const logger = createCommandsLogger(commander.loglevel)
   const logic = await new CommandsLogic(host, logger, {}, mnemonic).init()
   const registerOptions = {
+    sleep: 100,
```
drortirosh, line 27: you can't default to half-second on testnet/mainnet.

packages/dev/test/RegistrationManager.test.ts
```diff
@@ -45,23 +45,29 @@ contract('RegistrationManager', function (accounts) {
...
       })
       latestBlock = await env.web3.eth.getBlock('latest')
+      managerBalance = await relayServer.getManagerBalance()
+      assert.equal(managerBalance.cmp(toBN(expectedBalance)), 0, 'should have balance now')
       transactionHashes = await relayServer._worker(latestBlock.number)
-      assert.equal(transactionHashes.length, 0)
+      assert.equal(transactionHashes.length, 1, 'should only set owner')
+      const tx = await web3.eth.getTransaction(transactionHashes[0])
+      assert.equal(tx.to, env.stakeManager.address, 'should only set owner')
+      assert.equal(tx.input.indexOf('0xfece3dd4'), 0, 'should only set owner')
```
drortirosh, line 69: yuck. why not test emitted event?

packages/dev/test/RegistrationManager.test.ts
```diff
@@ -143,7 +153,7 @@ contract('RegistrationManager', function (accounts) {
 
     let newServer: RelayServer
     it('should initialize relay after staking and funding it', async function () {
-      await env.newServerInstanceNoInit({}, undefined, unstakeDelay)
+      await env.newServerInstance({}, undefined, unstakeDelay)
```
drortirosh, line 156: its a skipped test. does this help?

packages/dev/test/RelayServer.test.ts
```diff
@@ -7,7 +7,7 @@ import sinon from 'sinon'
 import sinonChai from 'sinon-chai'
 import chaiAsPromised from 'chai-as-promised'
 
-import { GSNConfig } from '@opengsn/relayclient/GSNConfigurator'
+import { GSNConfig } from '@opengsn/provider/dist/GSNConfigurator'
```
drortirosh, line 10: huh?

packages/dev/test/StakeManager.test.ts
```diff
@@ -70,29 +79,64 @@ contract('StakeManager', function ([_, relayManager, anyRelayHub, owner, nonOwne
...
           from: relayManager
         }),
-        'relayManager cannot stake for itself'
+        'already owned'
+      )
+    })
+
+    it('should not allow not owner to schedule unlock', async function () {
+      await expectRevert(
+        stakeManager.unlockStake(nonOwner, { from: owner }),
```
drortirosh, line 112: didn't you mean
```js
stakeManager.unlockStake(relayManager , {from: nonOwner })
```
?

packages/dev/test/StakeManager.test.ts
```diff
@@ -7,12 +7,21 @@ import { StakeManagerInstance } from '../../../types/truffle-contracts'
...
 
-contract('StakeManager', function ([_, relayManager, anyRelayHub, owner, nonOwner]) {
+contract('StakeManager', function ([_, relayManager, anyRelayHub, owner, nonOwner, secondRelayManager]) {
   const initialUnstakeDelay = new BN(4)
   const initialStake = ether('1')
   let stakeManager: StakeManagerInstance
 
-  function testCanStake (relayManager: string): void {
+  function testNotOwnerCannotStake (): void {
+    it('should not allow not owner to stake for owned relayManager address', async function () {
```
drortirosh, line 16: nitpic: the title should be "non-owner", and the member "nonOwner" (these are nouns, and in the "not owner" error its verb...
(but you really don't have to...)

packages/dev/test/TestUtils.ts
```diff
@@ -42,6 +42,9 @@ export async function startRelay (
   args.push('--relayHubAddress', relayHubAddress)
   const configFile = path.resolve(__dirname, './server-config.json')
   args.push('--config', configFile)
+  args.push('--stakeManagerAddress', stakeManager.address)
```
drortirosh, line 45: why? the SM was always retrieved from the RH.

packages/contracts/src/StakeManager.sol
```diff
@@ -20,37 +20,37 @@ contract StakeManager is IStakeManager {
...
+        stakes[msg.sender].owner = owner;
+        emit OwnerSet(msg.sender, owner);
+    }
+
+    /// Put a stake for a relayManager and set its unstake delay. Only the owner can call this function.
     /// @param relayManager - address that represents a stake entry and controls relay registrations on relay hubs
     /// @param unstakeDelay - number of blocks to elapse before the owner can retrieve the stake after calling 'unlock'
-    function stakeForAddress(address relayManager, uint256 unstakeDelay) external override payable {
-        require(stakes[relayManager].owner == address(0) || stakes[relayManager].owner == msg.sender, "not owner");
+    function stakeForAddress(address relayManager, uint256 unstakeDelay) external override payable ownerOnly(relayManager) {
```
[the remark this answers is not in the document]
forshtat: Is it worth it to add a rename here as well? Does not change semantics

packages/cli/src/CommandsLogic.ts
forshtat: added parameters for 'gsn-relayer-register' (no automated tests, will test manually on register branch)

packages/contracts/src/StakeManager.sol
```diff
@@ -20,37 +20,37 @@ contract StakeManager is IStakeManager {
...
 
-    function withdrawStake(address relayManager) external override {
+    function withdrawStake(address relayManager) external override ownerOnly(relayManager) {
         StakeInfo storage info = stakes[relayManager];
-        require(info.owner == msg.sender, "not owner");
         require(info.withdrawBlock > 0, "Withdrawal is not scheduled");
         require(info.withdrawBlock <= block.number, "Withdrawal is not due");
         uint256 amount = info.stake;
-        delete stakes[relayManager];
+        stakes[relayManager].stake = 0;
```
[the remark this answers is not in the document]
forshtat: fixed

packages/cli/src/CommandsLogic.ts
forshtat: debugger was disconnecting. not sure if this helped in fact

packages/dev/test/RegistrationManager.test.ts
forshtat: This is a bit tricky with the way Transaction Manager just broadcasts transactions, so I would have to manually decode logs after getting raw transaction from the node. 
Not sure if this is worth it for a single test

packages/dev/test/RegistrationManager.test.ts
forshtat: it does compile :-)

packages/dev/test/RelayServer.test.ts
forshtat: It failed to import and `tsc` is not run for test files (executed inside `ts-node`) so it just went on with `GSNConfig` being `any`.

It is easy to make 'tsc' run for test files but will try to output them somewhere, not sure how useful is that. You won't need to run tests to see compilation errors, so probably will add `tsc` for tests later.

packages/dev/test/TestUtils.ts
forshtat: seemed like the easiest way, why read it if can pass it?

packages/dev/test/StakeManager.test.ts
forshtat: yes, but it does not matter technically - all these addresses are not in any kind of relationship. Fixed.

shahafn (approved):

PR #578, opened 2021-02-01
OG-395 Relaycall max calldata size, OG-412 accounting calldata gas cost in maxPossibleGas

shahafn:


src/relayserver/RelayServer.ts
```diff
@@ -153,10 +162,12 @@ export class RelayServer extends EventEmitter {
     }
     // Check that the fee is acceptable
     if (parseInt(req.relayRequest.relayData.pctRelayFee) < this.config.pctRelayFee) {
-      throw new Error(`Unacceptable pctRelayFee: ${req.relayRequest.relayData.pctRelayFee} relayServer's pctRelayFee: ${this.config.pctRelayFee}`)
```
drortirosh, line 156: you repeat yourself. how about "checkRange(title,min,max,value)"

src/relayserver/RelayServer.ts
```diff
@@ -701,7 +721,11 @@ latestBlock timestamp   | ${latestBlock.timestamp}
   }
 
   _isTrustedPaymaster (paymaster: string): boolean {
-    return this.trustedPaymastersGasLimits.get(paymaster.toLocaleLowerCase()) != null
+    return this.trustedPaymastersGasAndDataLimits.get(paymaster.toLocaleLowerCase()) != null
```
drortirosh, line 724: toLowerCase. (not sure why the "locale" version was used here)

test/PaymasterCommitment.test.ts
```diff
@@ -189,6 +189,53 @@ contract('Paymaster Commitment', function ([_, relayOwner, relayManager, relayWo
...
+        request: {
+          // nonce: '4',
+          data: recipientContract.contract.methods.emitMessage('').encodeABI()
+        },
+        relayData: { paymaster }
+
+      }, sharedRelayRequestData, chainId, forwarderInstance)
+
+      const gasAndDataLimits = await paymasterContract.getGasAndDataLimits()
+      const hugeApprovalData = '0x' + 'ef'.repeat(parseInt(gasAndDataLimits.calldataSizeLimit) - 1000)
```
drortirosh, line 224: does it mean that the relay request (without the huge approval ) is 500 bytes long?

src/relayserver/RelayServer.ts
shahafn: Huh? That's not part of this PR

src/relayserver/RelayServer.ts
shahafn: Not Part of this PR, not sure who put `Locale` there

contracts/BasePaymaster.sol
```diff
@@ -32,19 +32,21 @@ abstract contract BasePaymaster is IPaymaster, Ownable {
     uint256 constant public PRE_RELAYED_CALL_GAS_LIMIT = 100000;
     uint256 constant public POST_RELAYED_CALL_GAS_LIMIT = 110000;
     uint256 constant public PAYMASTER_ACCEPTANCE_BUDGET = PRE_RELAYED_CALL_GAS_LIMIT + FORWARDER_HUB_OVERHEAD;
+    uint256 constant public CALLDATA_SIZE_LIMIT = 22000;
 
-    function getGasLimits()
+    function getGasAndDataLimits()
```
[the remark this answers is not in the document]
forshtat: I suggest keeping the original name

contracts/BasePaymaster.sol
```diff
@@ -32,19 +32,21 @@ abstract contract BasePaymaster is IPaymaster, Ownable {
     uint256 constant public PRE_RELAYED_CALL_GAS_LIMIT = 100000;
     uint256 constant public POST_RELAYED_CALL_GAS_LIMIT = 110000;
     uint256 constant public PAYMASTER_ACCEPTANCE_BUDGET = PRE_RELAYED_CALL_GAS_LIMIT + FORWARDER_HUB_OVERHEAD;
+    uint256 constant public CALLDATA_SIZE_LIMIT = 22020;
```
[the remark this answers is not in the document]
forshtat: I was under impression we agreed the limit will be nominated in gas, not in bytes.
So this would be calldata cost limit, was that changed?

contracts/RelayHub.sol
```diff
@@ -114,24 +117,26 @@ contract RelayHub is IRelayHub {
...
     view
-    returns (IPaymaster.GasLimits memory gasLimits, uint256 maxPossibleGas) {
-        gasLimits =
-            IPaymaster(relayRequest.relayData.paymaster).getGasLimits{gas:50000}();
-
-        require(paymasterMaxAcceptanceBudget >= gasLimits.acceptanceBudget, "unexpected high acceptanceBudget");
+    returns (IPaymaster.GasAndDataLimits memory gasAndDataLimits, uint256 maxPossibleGas) {
+        gasAndDataLimits =
+            IPaymaster(relayRequest.relayData.paymaster).getGasAndDataLimits{gas:50000}();
+        require(msg.data.length <= gasAndDataLimits.calldataSizeLimit, "msg.data exceeded limit" );
```
forshtat, line 130: If the next version of a relay hub uses x2 the `_dataGasCostPerByte`, this Paymaster will not be able to update to use it.

contracts/RelayHub.sol
```diff
@@ -217,7 +222,7 @@ contract RelayHub is IRelayHub {
     {
         if (!vars.success) {
             //Failure cases where the PM doesn't pay
-            if ( (vars.innerGasUsed < vars.gasLimits.acceptanceBudget ) && (
+            if ( (vars.innerGasUsed < vars.gasAndDataLimits.acceptanceBudget ) && (
                     vars.status == RelayCallStatus.RejectedByPreRelayed ||
```
forshtat, line 226: This was supposed to be changed so that `RejectedByPreRelayed` is considered top priority rule.

test/PaymasterCommitment.test.ts
```diff
@@ -189,6 +189,53 @@ contract('Paymaster Commitment', function ([_, relayOwner, relayManager, relayWo
...
+      await expectRevert(
+        relayHubInstance.relayCall(10e6, r.req, r.sig, hugeApprovalData, externalGasLimit, {
+          from: relayWorker,
+          gas: externalGasLimit,
+          gasPrice
+        }), 'msg.data exceeded limit'
+      )
+    })
+
+    it('paymaster should not pay for requests with max msg.data size if it rejects in pre', async () => {
```
forshtat, line 212: I do not see a corresponding change in the Relay Hub, indicating this test is not correct.

test/RelayHubGasCalculations.test.ts
```diff
@@ -337,6 +345,88 @@ contract('RelayHub gas calculations', function ([_, relayOwner, relayWorker, rel
...
+  describe('relayCall()\'s msg.data cost calculations', function () {
+    enum RelayCallDynamicArg {
+      APPROVAL_DATA = 'approvalData',
+      ENCODED_FUNCTION = 'encodedFunction',
+      PAYMASTER_DATA = 'paymasterData'
+    }
+
+    before(async function () {
+
+    });
```
forshtat, line 357: empty before is useless

test/RelayHubGasCalculations.test.ts
```diff
@@ -337,6 +345,88 @@ contract('RelayHub gas calculations', function ([_, relayOwner, relayWorker, rel
       })
   })
 
+  describe('relayCall()\'s msg.data cost calculations', function () {
+    enum RelayCallDynamicArg {
+      APPROVAL_DATA = 'approvalData',
```
forshtat, line 350: giving string aliases to these enums does not seem useful

test/RelayHubGasCalculations.test.ts
```diff
@@ -337,6 +345,88 @@ contract('RelayHub gas calculations', function ([_, relayOwner, relayWorker, rel
...
+          // console.log('gasUsed: ', gassesUsed)
+          const id = (await snapshot()).result
+          const senderNonce = (await forwarderInstance.getNonce(senderAddress)).toString()
+          let approvalData = '0x'
+          let paymasterData = '0x'
+          let encodedFunction = recipient.contract.methods.dontEmitMessage('').encodeABI()
+          if (dynamicArg === RelayCallDynamicArg.APPROVAL_DATA) {
+            approvalData = '0x' + 'ff'.repeat(dataLength)
+          } else if (dynamicArg === RelayCallDynamicArg.ENCODED_FUNCTION) {
+            // encoded function arg is an ascii string, so each char is a byte. We append half the length then.
```
forshtat, line 371: I don't see what 'half the length' refers to here

test/RelayHubGasCalculations.test.ts
```diff
@@ -337,6 +345,88 @@ contract('RelayHub gas calculations', function ([_, relayOwner, relayWorker, rel
...
+          })
+          gassesUsed.push(receipt.gasUsed - calculateCalldataCost(relayCall.encodeABI()))
+          // console.log('relayCall encodeABI len', relayCall.encodeABI().length / 2)
+          // console.log('gasUsed is', receipt.gasUsed)
+          // console.log('calculateCalldataCost is', calculateCalldataCost(relayCall.encodeABI()))
+          if (gassesUsed.length > 1) {
+            const diff = gassesUsed[gassesUsed.length - 1] - gassesUsed[0]
+            // console.log('diff is', diff)
+            // console.log('diff per byte is', diff / dataLength)
+            if (diff / dataLength > hubDataGasCostPerByte) {
```
forshtat, line 420: there is an `assert.closeTo` check. I think it will be better fit as if our value is waay to big it may also be a problem

test/RelayHubGasCalculations.test.ts
```diff
@@ -350,82 +440,84 @@ contract('RelayHub gas calculations', function ([_, relayOwner, relayWorker, rel
...
-                  nonce: senderNonce,
-                  value: '0',
-                  gas: gasLimit.toString()
-                },
-                relayData: {
-                  baseRelayFee,
-                  pctRelayFee,
-                  gasPrice: gasPrice.toString(),
-                  relayWorker,
+            it(`should compensate relay with requested fee of ${requestedFee.toString()}% with ${messageLength.toString()} calldata size`,
```
forshtat, line 443: why does this tests appear as entirely new?

forshtat (changes requested):

src/relayserver/RelayServer.ts
drortirosh: I see it part of the diff.
you should create a PR against release, and then it will contain only its own content, and not other PRs, otherwise, I can't tell.

test/PaymasterCommitment.test.ts
```diff
@@ -189,6 +189,53 @@ contract('Paymaster Commitment', function ([_, relayOwner, relayManager, relayWo
...
+      await paymasterContract.setRevertPreRelayCall(true)
```
drortirosh, line 213: this is not the case Oleksii raised: preRelayedCall should succeed, and then forwarder should fail (on nonce error)
in this case, 22000 bytes of approvalData makes the paymaster pay (=relayCall emits "TransactionRelayed" instead of "RejectedByPaymaster"

src/relayserver/RelayServer.ts
shahafn: The only change is indentation.

contracts/BasePaymaster.sol
```diff
@@ -32,19 +32,21 @@ abstract contract BasePaymaster is IPaymaster, Ownable {
     uint256 constant public PRE_RELAYED_CALL_GAS_LIMIT = 100000;
     uint256 constant public POST_RELAYED_CALL_GAS_LIMIT = 110000;
     uint256 constant public PAYMASTER_ACCEPTANCE_BUDGET = PRE_RELAYED_CALL_GAS_LIMIT + FORWARDER_HUB_OVERHEAD;
+    uint256 constant public CALLDATA_SIZE_LIMIT = 22000;
```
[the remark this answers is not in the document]
shahafn: Reduced to 10K, please check with Augur etc. for a reasonable value

contracts/RelayHub.sol
shahafn: Not in this PR

contracts/RelayHub.sol
shahafn: This only validates paymasters' own requirement of data length, how does it change through hubs?

test/PaymasterCommitment.test.ts
shahafn: See line 136 in RelayHub.sol. We added max calldata size, hub as this `require`. This is literally the edge case You asked me to add, when wee send the maximum size and paymaster rejects.

contracts/RelayHub.sol
```diff
@@ -114,24 +117,30 @@ contract RelayHub is IRelayHub {
...
-        gasLimits =
-            IPaymaster(relayRequest.relayData.paymaster).getGasLimits{gas:50000}();
-
-        require(paymasterMaxAcceptanceBudget >= gasLimits.acceptanceBudget, "unexpected high acceptanceBudget");
+    returns (IPaymaster.GasAndDataLimits memory gasAndDataLimits, uint256 maxPossibleGas) {
+        gasAndDataLimits =
+            IPaymaster(relayRequest.relayData.paymaster).getGasAndDataLimits{gas:50000}();
+        require(msg.data.length <= gasAndDataLimits.calldataSizeLimit, "msg.data exceeded limit" );
+        uint256 dataGasCost = calldataGasCost(msg.data.length);
+        require(maxRelayExposure >= gasAndDataLimits.acceptanceBudget.add(dataGasCost), "unexpected high maxRelayExposure");
```
drortirosh, line 136: the unexpected value is the acceptanceBudget+dataCost, not the maxRelayExposure

contracts/interfaces/IPaymaster.sol
```diff
@@ -82,7 +83,7 @@ interface IPaymaster {
      *              Note that in most cases the paymaster shouldn't try use it at all. It is always checked
      *              by the forwarder immediately after preRelayedCall returns.
      *  @param approvalData - extra dapp-specific data (e.g. signature from trusted party)
-     *  @param maxPossibleGas - based on values returned from {@link getGasLimits},
+     *  @param maxPossibleGas - based on values returned from {@link getGasAndDataLimits},
```
drortirosh, line 86: .. and actual data size.

src/relayserver/ServerConfigParams.ts
```diff
@@ -79,9 +81,11 @@ const serverDefaultConfiguration: ServerConfigParams = {
   alertedBlockDelay: 0,
   minAlertedDelayMS: 0,
   maxAlertedDelayMS: 0,
-  maxAcceptanceBudget: 2e5,
+  // set to paymasters' default acceptanceBudget + RelayHub.calldataGasCost(<paymasters' default calldataSizeLimit>)
+  maxRelayExposure: defaultPaymasterConfiguration.acceptanceBudget + defaultRelayHubConfiguration.dataGasCostPerByte * defaultPaymasterConfiguration.calldataSizeLimit,
```
drortirosh, line 85: why? the whole idea is have a number here, e.g. 300k

Our tests then backtrack and make sure this number is viable.

contracts/BasePaymaster.sol
```diff
@@ -28,23 +28,25 @@ abstract contract BasePaymaster is IPaymaster, Ownable {
     //overhead of forwarder verify+signature, plus hub overhead.
     uint256 constant public FORWARDER_HUB_OVERHEAD = 50000;
 
-    //These parameters are documented in IPaymaster.GasLimits
+    //These parameters are documented in IPaymaster.GasAndDataLimits
     uint256 constant public PRE_RELAYED_CALL_GAS_LIMIT = 100000;
     uint256 constant public POST_RELAYED_CALL_GAS_LIMIT = 110000;
     uint256 constant public PAYMASTER_ACCEPTANCE_BUDGET = PRE_RELAYED_CALL_GAS_LIMIT + FORWARDER_HUB_OVERHEAD;
+    uint256 constant public CALLDATA_SIZE_LIMIT = 10020;
```
forshtat, line 35: Why 10'0**2**0? Why not 10'000?

contracts/RelayHub.sol
```diff
@@ -114,24 +117,30 @@ contract RelayHub is IRelayHub {
         emit Withdrawn(account, dest, amount);
     }
 
-    function verifyGasLimits(
-        uint256 paymasterMaxAcceptanceBudget,
+    function calldataGasCost(uint256 length) public override view returns (uint256) {
```
forshtat, line 120: Does it have to be part of an interface? I think this method can be internal.

src/relayserver/RelayServer.ts
```diff
@@ -188,17 +199,23 @@ export class RelayServer extends EventEmitter {
...
+    relayExposure: number
   }> {
     const paymaster = req.relayRequest.relayData.paymaster
-    let gasLimits = this.trustedPaymastersGasLimits.get(paymaster)
-    let acceptanceBudget: number
-    if (gasLimits == null) {
+    let gasAndDataLimits = this.trustedPaymastersGasAndDataLimits.get(paymaster)
+    let relayExposure: number
+
+    const maxBlockGas = 12e6
```
forshtat, line 210: hey hey hey, this is a magic number! please use server parameters to define something like that.
as an option, can ask for block limit on a latest block when initializing

src/relayserver/RelayServer.ts
```diff
@@ -209,26 +226,28 @@ export class RelayServer extends EventEmitter {
...
-      const paymasterAcceptanceBudget = parseInt(gasLimits.acceptanceBudget)
-      if (paymasterAcceptanceBudget > acceptanceBudget) {
+      const paymasterAcceptanceBudget = parseInt(gasAndDataLimits.acceptanceBudget)
+      // TODO remove, since it's redundant. This check is also done in relayCall() on-chain, so the server will fail
+      // on view call if these requirements aren't met.
+      if (paymasterAcceptanceBudget + dataGasCost > relayExposure) {
         if (!this._isTrustedPaymaster(paymaster)) {
           throw new Error(
-            `paymaster acceptance budget too high. given: ${paymasterAcceptanceBudget} max allowed: ${this.config.maxAcceptanceBudget}`)
+            `paymaster acceptance budget too high. given: ${paymasterAcceptanceBudget} max allowed: ${this.config.maxRelayExposure}`)
```
forshtat, line 235: also depends on `dataGasCost`, isn't it?

src/relayserver/ServerConfigParams.ts
```diff
@@ -30,11 +31,12 @@ export interface ServerConfigParams {
   readyTimeout: number
   devMode: boolean
   registrationBlockRate: number
-  maxAcceptanceBudget: number
+  maxRelayExposure: number
   alertedBlockDelay: number
   minAlertedDelayMS: number
   maxAlertedDelayMS: number
   trustedPaymasters: Address[]
+  blacklistedPaymasters: Address[]
```
forshtat, line 39: This is coming from an unapproved branch, doesn't it? Where is the rest of 'blacklist' commit here?

test/PaymasterCommitment.test.ts
forshtat: The description is incorrect then. This test "paymaster should not pay for requests with max msg.data size if it rejects in pre" is hard or impossible to implement. I cannot understand what is actually checked here.

test/PaymasterCommitment.test.ts
```diff
@@ -189,10 +189,83 @@ contract('Paymaster Commitment', function ([_, relayOwner, relayManager, relayWo
...
+        from: relayWorker,
+        gas: externalGasLimit,
+        gasPrice
+      }
+      relayCallParams.push(txdetails)
+      const res = await relayHubInstance.relayCall(...relayCallParams)
+      expectEvent(res, 'TransactionRejectedByPaymaster', { reason: encodeRevertReason('You asked me to revert, remember?') })
+    })
+
+    it('paymaster should not pay for requests with max msg.data size if it accepts in pre but forwarder fails', async () => {
```
forshtat, line 239: What is new in this test? How does it relate to the issue we are discussing?

test/PaymasterCommitment.test.ts
```diff
@@ -207,16 +280,19 @@ contract('Paymaster Commitment', function ([_, relayOwner, relayManager, relayWo
...
+      const dataGasCost = await relayHubInstance.calldataGasCost(relayCall.length)
+      const maxRelayExposure = parseInt(gasLimits.acceptanceBudget) + dataGasCost.toNumber()
       // fail if a bit lower
-      expectRevert(relayHubInstance.relayCall(parseInt(gasLimits.acceptanceBudget) - 1, r.req, r.sig, '0x', externalGasLimit, {
+      expectRevert(relayHubInstance.relayCall(maxRelayExposure - 1, r.req, r.sig, '0x', externalGasLimit, {
         from: relayWorker,
         gas: externalGasLimit,
         gasPrice
-      }), 'unexpected high acceptanceBudget')
+      }), 'unexpected high maxRelayExposure')
```
forshtat, line 292: as mentioned earlier, this is not exposure that is high, that is call data size/ low Paymaster acceptanceBudget. Now this is an interesting test to add - raise this exception on call data size.

forshtat (changes requested):

test/PaymasterCommitment.test.ts
drortirosh: my calculations (for OG-416) show a minimum (dynamic buffers empty) relayCall is 22 words (4+ 22*32=708 bytes)
every non-empty string adds 1-word length and actual data ( **signature** is always non-empty, **req.data** is almost always non-empty (at least from solidity calls), **paymasterData**, **approvalData** are optional)

bottom line: I'm not sure you got to this "1000"

contracts/RelayHub.sol
shahafn: Currently it's also being used in the server, and in general I believe that it might vary between hubs, and might be needed outside, whether on or off chain.

contracts/RelayHub.sol
shahafn: fixed

src/relayserver/RelayServer.ts
shahafn: Right, though it's just a place holder for uint256 as parameter, since we only care about `msg.data.length` here.

src/relayserver/RelayServer.ts
```diff
@@ -209,26 +226,28 @@ export class RelayServer extends EventEmitter {
         }
         throw new Error(message)
       }
-      acceptanceBudget = this.config.maxAcceptanceBudget
-      const paymasterAcceptanceBudget = parseInt(gasLimits.acceptanceBudget)
-      if (paymasterAcceptanceBudget > acceptanceBudget) {
+      const paymasterAcceptanceBudget = parseInt(gasAndDataLimits.acceptanceBudget)
+      // TODO remove, since it's redundant. This check is also done in relayCall() on-chain, so the server will fail
+      // on view call if these requirements aren't met.
+      if (paymasterAcceptanceBudget + dataGasCost > relayExposure) {
```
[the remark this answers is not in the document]
shahafn: I didn't copy it. Just fixed an already existing check in the server. If we want to remove it, we should decide that outside of this PR.

src/relayserver/RelayServer.ts
shahafn: Right.

src/relayserver/ServerConfigParams.ts
shahafn: Yes, it's part of another PR that this one is branched out from.

src/relayserver/ServerConfigParams.ts
shahafn: I don't necessarily agree that it should be detached from how we got to that number. This default should be in line with the paymasters' and hubs' constants.

test/PaymasterCommitment.test.ts
shahafn: You're the one who specifically asked me to add this test for the edge case where the msg.data.length reaches limit and paymaster rejects. Do you want me to remove it?

test/PaymasterCommitment.test.ts
shahafn: See that below

test/PaymasterCommitment.test.ts
shahafn: This is literally the attack raised by the audit.

test/PaymasterCommitment.test.ts
shahafn: fixed

test/RelayHubGasCalculations.test.ts
shahafn: fixd

test/RelayHubGasCalculations.test.ts
shahafn: fixed

test/RelayHubGasCalculations.test.ts
shahafn: Removed the comment anyway

test/RelayHubGasCalculations.test.ts
shahafn: Fixed

test/RelayHubGasCalculations.test.ts
shahafn: It doesn't, it's indentation

test/PaymasterCommitment.test.ts
forshtat: Oh, ok. I understand now. Test description was not clear, or maybe I was drunk, I don't know.

contracts/RelayHub.sol
forshtat: accepted

test/PaymasterCommitment.test.ts
forshtat: That is correct, I misread the sentence

src/relayserver/RelayServer.ts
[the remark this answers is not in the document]
forshtat: agreed

drortirosh:
I don't see any place that attempt to calculate `dataGasCostPerByte` -  I'd expect a  test to show in the logs how close is the value in the environment to the expected value.
I see tests that fail on `callDataSizeLimit` - but none that are based on the data-size calculation

shahafn:
> `dataGasCostPerByte`

Look closer then.

forshtat:
> > I don't see any place that attempt to calculate `dataGasCostPerByte` - I'd expect a test to show in the logs how close is the value in the environment to the expected value.
> > I see tests that fail on `callDataSizeLimit` - but none that are based on the data-size calculation
> 
> Look closer then.

Screenshot 2021-02-11 at 04 13 09

contracts/BasePaymaster.sol
forshtat: put 10'500 please

forshtat (dismissed):

forshtat:
@drortirosh please re-review as well

test/RelayHubGasCalculations.test.ts
```diff
@@ -340,6 +348,91 @@ contract('RelayHub gas calculations', function ([_, relayOwner, relayWorker, rel
...
+          gassesUsed.push(receipt.gasUsed - calculateCalldataCost(relayCall.encodeABI()))
+          // console.log('relayCall encodeABI len', relayCall.encodeABI().length / 2)
+          // console.log('gasUsed is', receipt.gasUsed)
+          // console.log('calculateCalldataCost is', calculateCalldataCost(relayCall.encodeABI()))
+          const slack = 2
+          if (gassesUsed.length > 1) {
+            const diff = gassesUsed[gassesUsed.length - 1] - gassesUsed[0]
+            // console.log('diff per byte is', diff / dataLength)
+            // console.log('diff is', diff)
+            const costPerByte = diff / dataLength
```
drortirosh, line 421: You calculate `diffGas / absoluteDataLength`
you need to keep previous `dataLength` too for a meaningful calculation.

forshtat: @shahafn this sounds correct

contracts/RelayHub.sol
```diff
@@ -30,6 +30,7 @@ contract RelayHub is IRelayHub {
     uint256 public override postOverhead;
     uint256 public override gasReserve;
     uint256 public override maxWorkerCount;
+    uint256 public override dataGasCostPerByte;
```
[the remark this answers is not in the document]
shahafn: fixed

test/RelayHubGasCalculations.test.ts
forshtat: UPD: `dataLength` is not `absoluteDataLength` as it is just the 'additional' bytes. As this formula is always comparing with 0th gas usage value, it is actually a `diff` in this case

forshtat (approved):

PR #580, opened 2021-02-02
OG-413 Minor audit issues

forshtat:
6.8-6.10 - add '@notice' comments to  GsnTypes about gas price, fees
6.12 - inconsistent natspec compliance in comments
6.20 - _msgSender only works on msg.data >= 24
6.28 - workerCount[relayManager] read thrice in this function
6.38 - the unstakeDelay could be accidentally set to unrealistically
long periods of time resulting in a forever locked stake

### DIFF
--- a/migrations/2_deploy_contracts.js
+++ b/migrations/2_deploy_contracts.js
@@ -5,7 +5,7 @@ const SampleRecipient = artifacts.require('TestRecipient')
 const Forwarder = artifacts.require('Forwarder')
 
 module.exports = async function (deployer) {
-  await deployer.deploy(StakeManager)
+  await deployer.deploy(StakeManager, 30000)
   await deployer.deploy(Penalizer)
   await deployer.deploy(RelayHub, StakeManager.address, Penalizer.address, 0, 0, 0, 0, 0, 0, 0)
   await deployer.deploy(Forwarder)
--- a/src/common/Environments.ts
+++ b/src/common/Environments.ts
@@ -9,15 +9,16 @@ interface Environment {
   readonly chainId: number
   readonly mintxgascost: number
   readonly relayHubConfiguration: RelayHubConfiguration
+  readonly maxUnstakeDelay: number
 }
 
 /**
  * With about 6000 blocks per day, maximum unstake delay is defined at around 5 years for the mainnet.
  * This is done to prevent mistakenly setting an unstake delay to millions of years.
  */
-export const defaultStakeManagerMaxUnstakeDelay: number = 10000000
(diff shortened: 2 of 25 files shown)


src/common/Environments.ts
```diff
@@ -11,6 +11,12 @@ interface Environment {
   readonly relayHubConfiguration: RelayHubConfiguration
 }
 
+/**
+ * With about 6000 blocks per day, maximum unstake delay is defined at around 5 years for the mainnet.
+ * This is done to prevent mistakenly setting an unstake delay to millions of years.
+ */
+export const defaultStakeManagerMaxUnstakeDelay: number = 10000000
```
drortirosh, line 18: this is not a global parameter, but clearly an environment issue, so it should be in "environment" structure

test/StakeManager.test.ts
```diff
@@ -114,18 +115,28 @@ contract('StakeManager', function ([_, relayManager, anyRelayHub, owner, nonOwne
       )
     })
 
-    it('with owner set', async function () {
-      before(async function () {
+    context('with owner set', function () {
+      beforeEach(async function () {
```
drortirosh, line 119: don't "beforeEach" something that is needed once. we already have too many "beforeEach" which slow us down.
just keep it "before"

migrations/2_deploy_contracts.js
```diff
@@ -5,7 +5,7 @@ const SampleRecipient = artifacts.require('TestRecipient')
 const Forwarder = artifacts.require('Forwarder')
 
 module.exports = async function (deployer) {
-  await deployer.deploy(StakeManager)
+  await deployer.deploy(StakeManager, 30000)
```
drortirosh, line 8: you defined a constant for it, didn't you?

(not for this issue, but I thnk "truffle migrate" is MUCH more robust for installing contracts on networks than our deployer tool. 
Environments are great, and we should have one for each network, and that's great way to document the params used in each network. stakeManagerMaxUnstake is just one environment param

forshtat: It cannot be resolved in the 'migrations' script. Migrations are not used anymore but now is not the time to remove those.

test/StakeManager.test.ts
forshtat: Here it did not work. There are not so many tests in this `context`, so will research later.

src/common/Environments.ts
forshtat: Fixed.

drortirosh (approved):

PR #581, opened 2021-02-04
OG-410: forwarder frontrun value

drortirosh:
Make sure forwarder is deterministic on using value:
The forwrader will only handle value explicitly mentioned in
request.value, and value returned by the recipient call.

### DIFF
--- a/contracts/forwarder/Forwarder.sol
+++ b/contracts/forwarder/Forwarder.sol
@@ -59,12 +59,25 @@ contract Forwarder is IForwarder {
         _verifyAndUpdateNonce(req);
 
         bytes memory callData = abi.encodePacked(req.data, req.from);
+        uint preBalance = address (this).balance;
+        require(preBalance >= req.value, "FWD: insufficient value");
         require( gasleft()*63/64 >= req.gas, "FWD: insufficient gas" );
         // solhint-disable-next-line avoid-low-level-calls
         (success,ret) = req.to.call{gas : req.gas, value : req.value}(callData);
-        if ( address(this).balance>0 ) {
+
+        //this is a signed number (can be nagative). all params are known values,
+        // so can't overflow.
+        uint excessBalance = address (this).balance - preBalance + req.value;
+
+        // preBalance is the balance before this TX (and value sent to it as payable)
+        // preBalance-req.value - the expected balance after the call, assuming it didn't send anything back
+        // this.balance-(preBalance-req.value) = this.balance-preBalance+req.value
+        //   positive: the value sent by the method back to us: transfer to "from" address
+        //   negative: too much was sent to the forwarder (preBalance > req.value)
+        //         this excess is inaccessible.
+        if ( int(excessBalance)>0 ) {
             //can't fail: req.from signed (off-chain) the request, so it must be an EOA...
-            payable(req.from).transfer(address(this).balance);
+            payable(req.from).transfer(excessBalance);
         }
         return (success,ret);
     }
--- a/contracts/forwarder/Forwarder.sol
+++ b/contracts/forwarder/Forwarder.sol
@@ -67,7 +67,7 @@ contract Forwarder is IForwarder {
 
         //this is a signed number (can be nagative). all params are known values,
         // so can't overflow.
-        uint excessBalance = address (this).balance - preBalance + req.value;
+        int excessBalance = int(address (this).balance - preBalance + req.value);
 
         // preBalance is the balance before this TX (and value sent to it as payable)
         // preBalance-req.value - the expected balance after the call, assuming it didn't send anything back
@@ -75,9 +75,9 @@ contract Forwarder is IForwarder {
         //   positive: the value sent by the method back to us: transfer to "from" address
         //   negative: too much was sent to the forwarder (preBalance > req.value)
         //         this excess is inaccessible.
-        if ( int(excessBalance)>0 ) {
+        if ( excessBalance>0 ) {
             //can't fail: req.from signed (off-chain) the request, so it must be an EOA...
-            payable(req.from).transfer(excessBalance);
+            payable(req.from).transfer(uint(excessBalance));
         }
         return (success,ret);
     }
--- a/contracts/forwarder/Forwarder.sol
+++ b/contracts/forwarder/Forwarder.sol
@@ -59,15 +59,13 @@ contract Forwarder is IForwarder {
         _verifyAndUpdateNonce(req);
 
         bytes memory callData = abi.encodePacked(req.data, req.from);
-        uint preBalance = address (this).balance;
+        uint256 preBalance = address(this).balance;
         require(preBalance >= req.value, "FWD: insufficient value");
-        require( gasleft()*63/64 >= req.gas, "FWD: insufficient gas" );
+        require(gasleft()*63/64 >= req.gas, "FWD: insufficient gas" );
         // solhint-disable-next-line avoid-low-level-calls
-        (success,ret) = req.to.call{gas : req.gas, value : req.value}(callData);
+        (success,ret) = req.to.call{gas: req.gas, value: req.value}(callData);
 
-        //this is a signed number (can be nagative). all params are known values,
-        // so can't overflow.
-        int excessBalance = int(address (this).balance - preBalance + req.value);
+        int256 excessBalance = int256(address(this).balance - preBalance + req.value);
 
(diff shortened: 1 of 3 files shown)


contracts/forwarder/Forwarder.sol
```diff
@@ -59,12 +59,25 @@ contract Forwarder is IForwarder {
         _verifyAndUpdateNonce(req);
 
         bytes memory callData = abi.encodePacked(req.data, req.from);
+        uint preBalance = address (this).balance;
+        require(preBalance >= req.value, "FWD: insufficient value");
```
shahafn, line 63: Why preBalance and not msg.value here?

contracts/forwarder/Forwarder.sol
```diff
@@ -59,12 +59,25 @@ contract Forwarder is IForwarder {
...
         require( gasleft()*63/64 >= req.gas, "FWD: insufficient gas" );
         // solhint-disable-next-line avoid-low-level-calls
         (success,ret) = req.to.call{gas : req.gas, value : req.value}(callData);
-        if ( address(this).balance>0 ) {
+
+        //this is a signed number (can be nagative). all params are known values,
+        // so can't overflow.
+        uint excessBalance = address (this).balance - preBalance + req.value;
```
shahafn, line 70: Why not use `int`?

contracts/forwarder/Forwarder.sol
drortirosh: there are 2 ways to get value into the transaction:
1. sending eth to the forwarder just before calling execute (that's how a paymaster works)
2. passing "value" into execute()
checking `this.balance` uses the SUM of those two methods, so the rest of the code doesn't care what method was used to transfer in value.

shahafn: Why do we want to pass value before the `excute`?

contracts/forwarder/Forwarder.sol
drortirosh: fixed.

contracts/forwarder/Forwarder.sol
drortirosh: because a paymaster doesn't call `execute()`. Its the relayhub that calls it.
the the paymaster's preRelayedCall checks the `request.value` and transfer money into the forwarder, knowing the forwarder will use that money in the relayed call.
the "payable" mode is there for "completeness", but its unused by GSN.

forshtat: it will revert anyway on attempt to transfer more value than the current forwarder balance, why bother with this require?

contracts/forwarder/Forwarder.sol
```diff
@@ -59,12 +59,25 @@ contract Forwarder is IForwarder {
...
+        uint preBalance = address (this).balance;
+        require(preBalance >= req.value, "FWD: insufficient value");
         require( gasleft()*63/64 >= req.gas, "FWD: insufficient gas" );
         // solhint-disable-next-line avoid-low-level-calls
         (success,ret) = req.to.call{gas : req.gas, value : req.value}(callData);
-        if ( address(this).balance>0 ) {
+
+        //this is a signed number (can be nagative). all params are known values,
+        // so can't overflow.
+        int excessBalance = int(address (this).balance - preBalance + req.value);
```
forshtat, line 70: I don't see a point of doing it either. If someone 'forgot' value at forwarder, why keep it there? 
What if ETH is sent BACK to forwarder as part of `execute`, do we want to disable that? 

If the problem is insufficient gas for transfer, the solution sounds to be checking we have enough gas, not adding extra logic.

forshtat (changes requested):

contracts/forwarder/Forwarder.sol
drortirosh: the problem is not "adding extra gas", but "making transaction atomic and predictable."
We handle `req.value` eth passed into the forwarder as part of this transaction - either in payable value, or sent inside outer transaction (as paymaster does).
We don't care for other value.

contracts/forwarder/Forwarder.sol
drortirosh: not strictly required. it just gives better reason, instead of empty string revert (we also add a revert string on our notion of "OOG")

yoavw:
Overall looks fine, although it did add some logic.  The logic seems safe since we ensure that any value included in the transaction will leave the contract, so gas stays predictable.  One side effect is that there could be a contract that becomes unusable with GSN because it sends back a non-predictable value, e.g. an exchange that returns ETH based on the current exchange rate and slippage, although I don't see why such transaction should go through GSN at all.

If we do care about such cases, there may be a simpler solution.  The frontrunning problem is due to value being out-of-band (outside the signed message), and therefore someone could mess with the value of a signed meta-tx.  Why don't we just force it to be part of the signed message?  Add a value field which is normally zero.  If it's non-zero, the contract verifies that the transaction value is identical to the value specified in the signed message.  Makes everything explicit.

Wouldn't it solve the problem in a cleaner way?  The only downside I see is that it adds an extra int to all transactions (unless we can stash a flag in some existing parameter, and check for has_value before looking for the extra int.

drortirosh:
> One side effect is that there could be a contract that becomes unusable with GSN because it sends back a non-predictable value, e.g. an exchange that returns ETH based on the current exchange rate and slippage

This is still predictable. that is, if you call (say) Uniswap, you do expect to get some eth back (though the exact amount is unknown), and the current implementation handles it.

> Why don't we just force it to be part of the signed message? Add a value field which is normally zero. If it's non-zero, the contract verifies that the transaction value is identical to the value specified in the signed message. Makes everything explicit.

This is what we do: the `request.value` is the value we pass with the call, and the paymaster must make sure in advance that this value exist in the forwarder.

The whole mechanism is the handling any extra value, above what is specified as part of the `request.value`

contracts/forwarder/Forwarder.sol
```diff
@@ -59,16 +59,25 @@ contract Forwarder is IForwarder {
...
+        require(preBalance >= req.value, "FWD: insufficient value");
+        require(gasleft()*63/64 >= req.gas, "FWD: insufficient gas" );
         require(req.validUntil == 0 || req.validUntil > block.number, "FWD: request expired");
-
         // solhint-disable-next-line avoid-low-level-calls
-        (success,ret) = req.to.call{gas : req.gas, value : req.value}(callData);
-        if ( address(this).balance>0 ) {
+        (success,ret) = req.to.call{gas: req.gas, value: req.value}(callData);
+
+        int256 excessBalance = int256(address(this).balance - preBalance + req.value);
```
forshtat, line 69: By the way, you don't check if the call succeeded or not. If the recipient reverted, the entire `req.value` will be stuck in Forwarder, won't it?

forshtat:
The calculation does seem to break if `msg.value != preBalance`, not only if value was sent outside the call frame/transaction.
In both success and reverted inner call flow, this `excessBalance` calculation is only correct with `require(msg.value == this.balance)`.
Please comment.

contracts/forwarder/Forwarder.sol
```diff
@@ -59,16 +59,25 @@ contract Forwarder is IForwarder {
...
+
+        // preBalance is the balance before this TX (and value sent to it as payable)
+        // preBalance-req.value - the expected balance after the call, assuming it didn't send anything back
+        // this.balance-(preBalance-req.value) = this.balance-preBalance+req.value
+        //   positive: the value sent by the method back to us: transfer to "from" address
+        //   negative: too much was sent to the forwarder (preBalance > req.value)
+        //         this excess is inaccessible.
+        if ( excessBalance>0 ) {
```
shahafn, line 77: When is it ever non zero? There is no test that handles this from what I saw.

forshtat:
@drortirosh I am compiling a Mitigations document. If @shahafn chooses to merge this in despite my objections I will need you to describe the motivation and architecture for this change in writing, very readable. Please provide. Thank you.

contracts/forwarder/Forwarder.sol
shahafn, line 69: When recipient reverts, what happens?

shahafn: If recipient reverts, you send req.value to req.from, instead of paymaster, if anyone should receive it.

forshtat:
an alternative mitigation has been merged https://github.com/opengsn/gsn/pull/593

PR #582, opened 2021-02-04
OG-408 forwarder validUntil

drortirosh:
initial fix
(note that fixing all GSN tests requires a "shortgun" change on all tests, to add the new param

### DIFF
--- a/contracts/forwarder/Forwarder.sol
+++ b/contracts/forwarder/Forwarder.sol
@@ -60,7 +60,8 @@ contract Forwarder is IForwarder {
 
         bytes memory callData = abi.encodePacked(req.data, req.from);
         require( gasleft()*63/64 >= req.gas, "FWD: insufficient gas" );
-        require(req.validUntil==0 || req.validUntil < block.timestamp, "FWD: request expired");
+        // solhint-disable-next-line not-rely-on-time
+        require(req.validUntil==0 || req.validUntil > block.timestamp, "FWD: request expired");
 
         // solhint-disable-next-line avoid-low-level-calls
         (success,ret) = req.to.call{gas : req.gas, value : req.value}(callData);
@@ -116,11 +117,6 @@ contract Forwarder is IForwarder {
         typeHashes[requestTypehash] = true;
         emit RequestTypeRegistered(requestTypehash, requestType);
     }
-
-    event DomainRegistered(bytes32 indexed domainSeparator, bytes domainValue);
-
-    event RequestTypeRegistered(bytes32 indexed typeHash, string typeStr);
-
 
     function _verifySig(
         ForwardRequest calldata req,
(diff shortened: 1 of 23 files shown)


contracts/forwarder/Forwarder.sol
```diff
@@ -60,12 +60,16 @@ contract Forwarder is IForwarder {
 
         bytes memory callData = abi.encodePacked(req.data, req.from);
         require( gasleft()*63/64 >= req.gas, "FWD: insufficient gas" );
+        // solhint-disable-next-line not-rely-on-time
+        require(req.validUntil==0 || req.validUntil > block.timestamp, "FWD: request expired");
```
forshtat, line 64: Inconsistent spacing for boolean expressions. Unfortunately, solhint does not seem opinionated about that.

contracts/forwarder/Forwarder.sol
```diff
@@ -60,12 +60,16 @@ contract Forwarder is IForwarder {
...
+
         // solhint-disable-next-line avoid-low-level-calls
         (success,ret) = req.to.call{gas : req.gas, value : req.value}(callData);
         if ( address(this).balance>0 ) {
             //can't fail: req.from signed (off-chain) the request, so it must be an EOA...
             payable(req.from).transfer(address(this).balance);
         }
+
```
forshtat, line 72: ?

contracts/forwarder/Forwarder.sol
```diff
@@ -8,7 +8,7 @@ import "./IForwarder.sol";
 contract Forwarder is IForwarder {
     using ECDSA for bytes32;
 
-    string public constant GENERIC_PARAMS = "address from,address to,uint256 value,uint256 gas,uint256 nonce,bytes data";
+    string public constant GENERIC_PARAMS = "address from,address to,uint256 value,uint256 gas,uint256 nonce,bytes data,uint256 validUntil";
```
forshtat, line 11: I believe placement of 'bytes' parameter in the end of the struct was deliberate. Is there a downside to adding 'validUntil' field in the middle, with all the uint256 parameters?

contracts/forwarder/Forwarder.sol
forshtat, line 64: Also, we have specifically agreed to use block numbers, not timestamps, for validity. As well as everything else as we have agreed with the rule of not relying on block time. 
What changed since?

forshtat: Also, 'block' should really appear in the field's name

forshtat (changes requested):

contracts/forwarder/Forwarder.sol
forshtat: may I suggest 'expirationBlock' (or 'validUntilBlock', but that is 3 words)

forshtat (approved):

PR #583, opened 2021-02-04
Calculate transfer gas cost

drortirosh:
* calculate cost of calling just innerRelayedCall. assume this is the cost of any external function (forwarder, preRelayedCall)
* estimate if expected gas after preRelayedCall is enough for Forwarder verifications (including both code and data)


drortirosh:
doesn't work as expected.

PR #586, opened 2021-02-06
Remove penalization without commitment flow

forshtat:


src/relayserver/penalizer/PenalizerService.ts
```diff
@@ -140,15 +176,81 @@ export class PenalizerService {
...
+      commitHash,
+      type: PenalizationTypes.REPEATED_NONCE,
+      methodArgs: penalizationArguments
+    }
+    const commitTxHash = await this.commitAndScheduleReveal(delayedPenalization)
+    return { commitTxHash }
+  }
+
+  calculateCommitHash (method: any): PrefixedHexString {
+    const msgData = method.encodeABI()
```
drortirosh, line 198: need to have a local "nonce" appended to the encoded-abi. should use the same value here, and when sending the TX

forshtat: Fixed.

src/relayserver/penalizer/PenalizerService.ts
```diff
@@ -56,7 +59,29 @@ function createWeb3Transaction (transaction: Web3CoreTransaction, rawTxOptions:
...
+  readyBlockNumber?: number
+  type: PenalizationTypes
+  commitHash: PrefixedHexString
+  methodArgs: PrefixedHexString[]
+}
+
 export class PenalizerService {
+  private workerTask?: Timeout
+
+  // TODO: TransactionManager is not integrated with Penalizer Service, so there is a duplication here
```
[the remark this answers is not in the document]
forshtat: I believe the Penalizer Service that is part of the  Relay Service cannot depart from the way the Relay Server is implemented for many reasons.

src/relayserver/penalizer/PenalizerService.ts
```diff
@@ -140,15 +176,82 @@ export class PenalizerService {
       throw Error(`Failed to get transaction ${minedTransactionData.hash} from node`)
     }
     const minedTxBuffers = createWeb3Transaction(minedTx, rawTxOptions)
-    const method = this.getPenalizeRepeatedNonceMethod(minedTxBuffers, requestTx)
+    const randomValue = randomInRange(0, 0xffffffff).toString(16)
+    const penalizationArguments = this.getPenalizeRepeatedNonceArguments(minedTxBuffers, requestTx, randomValue)
```
drortirosh, line 180: I really don't understand why you use getXXarguments and getMethod(XX):
instead, keep the method object itself in the delayedPenalization.

forshtat: I hate the concept of 'method object', I think it is best to avoid using it if there is a simple way to use a plain JavaScript object.

drortirosh: You have the "object method" so much, that you have to create helper method for the params of each solidity function, and another switch/case helper method to select the right function....
I don't think its a good tradeoff.

drortirosh (approved):

PR #588, opened 2021-02-08
remove 0x trace/profile... not working

drortirosh:
0x profiling/tracing tools are 2 years old, and unmaintained
They have open issues for not supporting 0.6 compiler version, and now
obviously fail on 0.7 options..

They add 400Mb to installed packages and considerate time to "yarn
install", so they are now removed.

### DIFF
--- a/package.json
+++ b/package.json
@@ -104,10 +104,6 @@
     "winston": "^3.3.3"
   },
   "devDependencies": {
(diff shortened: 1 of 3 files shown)


forshtat:
no need - will migrate to workspaces immediately after release. can be changed there.

PR #589, opened 2021-02-09
OG-416: validate no extra bytes in msg.data

drortirosh:
- validate the packing of the relayCall request.
- make transaction penalizeable if the packing is not cannonicall, and
contains "rubbish" in excess of actual parameters (either between or
after parametrs)
- relayer now also make the test against the Penalizer, to avoid
submitting transactions that are penalizeable (either because of a bug
in the relayer, or malicious data from client)

### DIFF
--- a/contracts/utils/RelayHubValidator.sol
+++ b/contracts/utils/RelayHubValidator.sol
@@ -1,0 +1,24 @@
+// SPDX-License-Identifier:MIT
+pragma solidity ^0.7.5;
+pragma abicoder v2;
+
+import "../interfaces/GsnTypes.sol";
+library RelayHubValidator {
+
+    // validate that encoded relayCall is properly packed without any extra bytes
+    function verifyTransactionPacking(
+        GsnTypes.RelayRequest calldata relayRequest,
+        bytes calldata signature,
+        bytes calldata approvalData
+    ) internal pure {
+        uint expectedMsgDataLen = 4 + 22*32 +
+        len1(signature) + len1(approvalData) + len1(relayRequest.request.data) + len1(relayRequest.relayData.paymasterData);
+        require(signature.length <= 65, "invalid signature length");
+        require(expectedMsgDataLen == msg.data.length, "extra msg.data bytes" );
+    }
+
+    // helper method for verifyTransactionPacking
+    function len1(bytes calldata buf) internal pure returns (uint) {
+        return 32 + ((buf.length+31) & uint(~31));
+    }
+}
--- a/contracts/utils/RelayHubValidator.sol
+++ b/contracts/utils/RelayHubValidator.sol
@@ -11,14 +11,28 @@ library RelayHubValidator {
         bytes calldata signature,
         bytes calldata approvalData
     ) internal pure {
-        uint expectedMsgDataLen = 4 + 22*32 +
-        len1(signature) + len1(approvalData) + len1(relayRequest.request.data) + len1(relayRequest.relayData.paymasterData);
+        // abicoder v2: https://docs.soliditylang.org/en/latest/abi-spec.html
+        // each static param/member is 1 word
+        // struct (with dynamic members) has offset to struct which is 1 word
+        // dynamic member is 1 word offset to actual value, which is 1-word length and ceil(length/32) words for data
+        // relayCall has 5 method params,
+        // relayRequest: 2 members
+        // relayData 8 members
+        // ForwardRequest: 7 members
+        // total 22 32-byte words if all dynamic params are zero-length.
+        uint expectedMsgDataLen = 4 + 22 * 32 +
+            dynamicParamSize(signature) +
+            dynamicParamSize(approvalData) +
+            dynamicParamSize(relayRequest.request.data) +
+            dynamicParamSize(relayRequest.relayData.paymasterData);
         require(signature.length <= 65, "invalid signature length");
         require(expectedMsgDataLen == msg.data.length, "extra msg.data bytes" );
     }
 
-    // helper method for verifyTransactionPacking
-    function len1(bytes calldata buf) internal pure returns (uint) {
-        return 32 + ((buf.length+31) & uint(~31));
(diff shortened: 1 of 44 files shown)


contracts/utils/RelayHubValidator.sol
```diff
@@ -0,0 +1,24 @@
...
+import "../interfaces/GsnTypes.sol";
+library RelayHubValidator {
+
+    // validate that encoded relayCall is properly packed without any extra bytes
+    function verifyTransactionPacking(
+        GsnTypes.RelayRequest calldata relayRequest,
+        bytes calldata signature,
+        bytes calldata approvalData
+    ) internal pure {
+        uint expectedMsgDataLen = 4 + 22*32 +
```
forshtat, line 14: please use explicit integer sizes

forshtat, line 14: please define named constants

contracts/utils/RelayHubValidator.sol
```diff
@@ -0,0 +1,24 @@
...
+        len1(signature) + len1(approvalData) + len1(relayRequest.request.data) + len1(relayRequest.relayData.paymasterData);
+        require(signature.length <= 65, "invalid signature length");
+        require(expectedMsgDataLen == msg.data.length, "extra msg.data bytes" );
+    }
+
+    // helper method for verifyTransactionPacking
+    function len1(bytes calldata buf) internal pure returns (uint) {
```
forshtat, line 21: please give meaningful names to methods

contracts/utils/RelayHubValidator.sol
```diff
@@ -0,0 +1,24 @@
...
+        return 32 + ((buf.length+31) & uint(~31));
```
forshtat, line 22: please define named constants

test/RelayHub.test.ts
```diff
@@ -422,6 +422,17 @@ contract('RelayHub', function ([_, relayOwner, relayManager, relayWorker, sender
           )
         })
 
+        it('should revert if encoded function contains extra bytes', async () => {
```
forshtat, line 425: please add test for signature require

test/RelayHubValidator.test.ts
```diff
@@ -0,0 +1,72 @@
...
+const TestRelayHubValidator = artifacts.require('TestRelayHubValidator')
+
+contract('RelayHubValidator', ([from, senderAddress, target, paymaster, relayWorker, forwarder]) => {
+  let validator: any
+  before(async () => {
+    validator = await TestRelayHubValidator.new()
+  })
+
+  it('#len1 should return dynamic structure size', async () => {
+    const len1 = async function (size: number): Promise<number> { return validator.len1('0x' + '11'.repeat(size)).then((x: any) => x.toNumber()) }
```
forshtat, line 12: please give meaningful names to lambda functions

test/RelayHubValidator.test.ts
```diff
@@ -0,0 +1,72 @@
...
+
+const TestRelayHubValidator = artifacts.require('TestRelayHubValidator')
+
+contract('RelayHubValidator', ([from, senderAddress, target, paymaster, relayWorker, forwarder]) => {
+  let validator: any
+  before(async () => {
+    validator = await TestRelayHubValidator.new()
+  })
+
+  it('#len1 should return dynamic structure size', async () => {
```
forshtat, line 11: please start with 'should' at all costs

test/RelayHubValidator.test.ts
```diff
@@ -0,0 +1,72 @@
...
+    { approvalData: '0x123456' },
+    { approvalData: '0x123456', signature: '0xabcd' },
+    { signature: '0xab' },
+    { data: '0x1234' },
+    { data: '0x1234', signature: '0xabcdef' },
+    { paymasterData: '0x12345678' },
+    { suffix: '123456' },
+    { data: '0x1234', signature: '0xabcdef', suffix: '123456' }
+  ].forEach((appended: any) => {
+    it(`test verify length with "${JSON.stringify(appended)}" `, async () => {
```
forshtat, line 33: please start with 'should' at all costs

forshtat (changes requested):

contracts/utils/RelayHubValidator.sol
drortirosh: what name? `mask_ffffe0` ? I don't have a better name for this mask...

test/RelayHub.test.ts
drortirosh: there is, but it the Validator test. here is just a single test, just to verify the "verify" method is called.

test/RelayHubValidator.test.ts
drortirosh: sorry, English no good for that. its a sanity check that the method does its work.
`#len1` should be enough, though

test/RelayHubValidator.test.ts
drortirosh: localWrapperForSolidityLen1ToBeEasilyCalledFromAssert ?

contracts/utils/RelayHubValidator.sol
drortirosh: the old elaborated description was lost. added back.

test/RelayHub.test.ts
forshtat: there is no check for a `require('invalid signature length'`

test/RelayHubValidator.test.ts
```diff
@@ -0,0 +1,101 @@
...
+            baseRelayFee: 2,
+            relayWorker,
+            paymaster: paymaster,
+            paymasterData: '0x',
+            clientId: 3,
+            forwarder
+          }
+        },
+        '0x' + '11'.repeat(66),
+        '0x', 0), 'invalid sig')
```
forshtat, line 98: "invalid signature length"

forshtat (approved):

PR #595, opened 2021-02-14
OG-418 prioritize paymaster pre rejection over acceptance budget

shahafn:


contracts/test/TestPaymasterConfigurableMisbehavior.sol
```diff
@@ -44,6 +45,9 @@ contract TestPaymasterConfigurableMisbehavior is TestPaymasterEverythingAccepted
     function setExpensiveGasLimits(bool val) public {
         expensiveGasLimits = val;
     }
+    function setPreGasHigherThanAcceptance(bool val) public {
```
drortirosh, line 48: you have `setGasLimits` for this, setting all 3 gas params

contracts/test/TestPaymasterConfigurableMisbehavior.sol
```diff
@@ -124,6 +128,13 @@ contract TestPaymasterConfigurableMisbehavior is TestPaymasterEverythingAccepted
             return IPaymaster.GasAndDataLimits(limits.acceptanceBudget * 9, limits.preRelayedCallGasLimit, limits.postRelayedCallGasLimit,
             limits.calldataSizeLimit);
         }
+        if (preHigherThanAcceptance) {
+            return IPaymaster.GasAndDataLimits(
+                limits.acceptanceBudget,
+                limits.preRelayedCallGasLimit + limits.acceptanceBudget + relayHub.calldataGasCost(limits.calldataSizeLimit),
```
drortirosh, line 134: too much magic in contract instead of in the test itself.

shahafn: Fixing, though this is a test contract, so does it matter where this magic happens?

contracts/test/TestPaymasterConfigurableMisbehavior.sol
shahafn: fixed

drortirosh (approved):

PR #596, opened 2021-02-15
ver-2.2.0-rc.1

drortirosh:


migrations/1_initial_migration.js
```diff
@@ -1,5 +0,0 @@
-var Migrations = artifacts.require('./Migrations.sol')
```
forshtat, line 1: Is there a reason to include this removal with the version change commit?

forshtat (changes requested):

PR #601, opened 2021-02-24
OG-431: max possible gas tx data cost; OG-429 - put boundaries on externalGasLimit

drortirosh:
calculate maxPossibleGas based on actual tx data cost (cost of putting the TX on-chain)
 - since this value is based on externalGasLimit, make sure it is within  the valid range (each byte cost is 4..16)

### DIFF
--- a/contracts/RelayHub.sol
+++ b/contracts/RelayHub.sol
@@ -32,8 +32,12 @@ contract RelayHub is IRelayHub {
     uint256 public immutable override gasReserve;
     uint256 public immutable override maxWorkerCount;
     uint256 public immutable override dataGasCostPerByte;
+    uint256 public immutable override relayCallDataOverhead;
     IStakeManager immutable  override public stakeManager;
     address immutable override public penalizer;
+
+    uint256 public constant Gzero = 4;
+    uint256 public constant Gnonzero = 16;
 
     // maps relay worker's address to its manager's address
     mapping(address => address) public override workerToManager;
@@ -53,7 +57,8 @@ contract RelayHub is IRelayHub {
         uint256 _maximumRecipientDeposit,
         uint256 _minimumUnstakeDelay,
         uint256 _minimumStake,
-        uint256 _dataGasCostPerByte
+        uint256 _dataGasCostPerByte,
+        uint256 _relayCallDataOverhead
     ) {
         stakeManager = _stakeManager;
         penalizer = _penalizer;
@@ -65,6 +70,7 @@ contract RelayHub is IRelayHub {
         minimumUnstakeDelay = _minimumUnstakeDelay;
         minimumStake =  _minimumStake;
         dataGasCostPerByte = _dataGasCostPerByte;
+        relayCallDataOverhead = _relayCallDataOverhead;
     }
 
     function registerRelayServer(uint256 baseRelayFee, uint256 pctRelayFee, string calldata url) external override {
@@ -126,7 +132,8 @@ contract RelayHub is IRelayHub {
     function verifyGasAndDataLimits(
         uint256 maxRelayExposure,
         GsnTypes.RelayRequest calldata relayRequest,
-        uint256 initialGas
+        uint256 initialGasLeft,
+        uint256 externalGasLimit
     )
     private
     view
@@ -135,6 +142,10 @@ contract RelayHub is IRelayHub {
             IPaymaster(relayRequest.relayData.paymaster).getGasAndDataLimits{gas:50000}();
         require(msg.data.length <= gasAndDataLimits.calldataSizeLimit, "msg.data exceeded limit" );
         uint256 dataGasCost = calldataGasCost(msg.data.length);
+        uint256 txDataCost = externalGasLimit - initialGasLeft - relayCallDataOverhead;
+        uint256 txDataCostPerByte = txDataCost/msg.data.length;
+        require(txDataCostPerByte >= Gzero && txDataCostPerByte <= Gnonzero, "invalid externalGasLimit")
+
         require(maxRelayExposure >= gasAndDataLimits.acceptanceBudget.add(dataGasCost), "pm budget + dataGasCost too high");
 
         maxPossibleGas =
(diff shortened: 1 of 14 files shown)


contracts/RelayHub.sol
```diff
@@ -32,9 +32,13 @@ contract RelayHub is IRelayHub {
     uint256 public immutable override gasReserve;
     uint256 public immutable override maxWorkerCount;
     uint256 public immutable override dataGasCostPerByte;
+    uint256 public immutable override relayCallDataOverhead;
```
forshtat, line 35: Is this a constant value? How is it calculated?

contracts/RelayHub.sol
```diff
@@ -135,19 +142,24 @@ contract RelayHub is IRelayHub {
             IPaymaster(relayRequest.relayData.paymaster).getGasAndDataLimits{gas:50000}();
         require(msg.data.length <= gasAndDataLimits.calldataSizeLimit, "msg.data exceeded limit" );
         uint256 dataGasCost = calldataGasCost(msg.data.length);
+        uint256 txDataCost = externalGasLimit - initialGasLeft - relayCallDataOverhead;
+        uint256 txDataCostPerByte = txDataCost/msg.data.length;
+        require(txDataCostPerByte >= G_ZERO && txDataCostPerByte <= G_NONZERO, "invalid externalGasLimit");
```
forshtat, line 147: I am not sure this check is that useful. It requires adding 3 constants to the RelayHub for the benefit of setting very wide boundaries on the `externalGasLimit` which is still penalizeable if lied about. 

With false `externalGasLimit` the `actualCharge` can become higher than `maxPossibleGas` with or without this check.

Let's discuss this check further.

contracts/RelayHub.sol
```diff
@@ -135,19 +142,24 @@ contract RelayHub is IRelayHub {
...
         require(maxRelayExposure >= gasAndDataLimits.acceptanceBudget.add(dataGasCost), "pm budget + dataGasCost too high");
 
         maxPossibleGas =
             gasOverhead.add(
             gasAndDataLimits.preRelayedCallGasLimit).add(
             gasAndDataLimits.postRelayedCallGasLimit).add(
             relayRequest.request.gas).add(
-            dataGasCost);
+            dataGasCost).add(
+            txDataCost);
```
forshtat, line 157: these two names are confusing, let's look for a better couple, maybe?

contracts/RelayHub.sol
```diff
@@ -135,19 +142,24 @@ contract RelayHub is IRelayHub {
...
 
         // This transaction must have enough gas to forward the call to the recipient with the requested amount, and not
         // run out of gas later in this function.
         require(
-            initialGas >= maxPossibleGas,
+            externalGasLimit >= maxPossibleGas,
```
forshtat, line 162: Maybe if we are touching the line, this is a good place to add another `gasLimit * (63/64)^3` check? Just asking. It is on relay's account if the gas is insufficient in most cases

src/common/Utils.ts
```diff
@@ -104,21 +105,42 @@ export async function getEip712Signature (
...
+      msgDataGasCost +
+      txDataCost
+
+  // maxPossibleGas =
+  //     gasOverhead.add(
+  //         gasAndDataLimits.preRelayedCallGasLimit).add(
+  //         gasAndDataLimits.postRelayedCallGasLimit).add(
+  //         relayRequest.request.gas).add(
+  //         dataGasCost).add(
+  //         txDataCost);
```
forshtat, line 143: \--

test/RelayHubPenalizations.test.ts
```diff
@@ -442,7 +442,7 @@ contract('RelayHub Penalizations', function ([_, relayOwner, relayWorker, commit
             from: other,
             value: ether('1')
           })
-          const relayCallTx = await relayHub.relayCall(10e6, relayRequest, signature, '0x', gasLimit.add(new BN(2e6)), {
+          const relayCallTx = await relayHub.relayCall(10e6, relayRequest, signature, '0x', gasLimit.add(new BN(1e6 - 1)), {
```
forshtat, line 445: Why is this test affected?

test/TestBatchForwarder.test.ts
```diff
@@ -105,7 +105,8 @@ contract('BatchForwarder', ([from, relayManager, relayWorker, relayOwner]) => {
       )
 
       const ret = await hub.relayCall(10e6, relayRequest, signature, '0x', 7e6, {
-        from: relayWorker
+        from: relayWorker,
+        gas: 7e6
```
forshtat, line 109: Why is this test affected?

forshtat (changes requested):

contracts/RelayHub.sol
drortirosh: calculated here: https://bit.ly/3uAiZVe

When calculating external datacost as `externalGasLimit-gasleft()`, this is a fixed overhead added. verified to be the same value on ganache and on geth.
It includes the 21000 stipend, and the cost of calling the `relayCall()`. 
Note that it is used just to validate our txDataCost is accurate. this value is later part of the gasOverhead, but I didn't want to change the meaning of gasOverhead and separate it from this value, since it would complicate a lot of the code.

contracts/RelayHub.sol
drortirosh: Assuming we can't think of a relayer attack based on providing too little gas, we can remove the G_ZERO test
the G_NONZERO is there to prevent the OG-429 (externalGasLimit attack)

Its true that OG-429 is not strictly part of this issue, and the check could be a separate

contracts/RelayHub.sol
drortirosh: onChainDataCost and externalDataCost ?

contracts/RelayHub.sol
drortirosh: its a separate issue, and requires its own test (the change is simple, but the test isn't..)

src/common/Utils.ts
drortirosh: ok (though this line is a match to a solidity code too bad both use some unreadable construct - and each is different...

test/RelayHubPenalizations.test.ts
drortirosh: because the original test failed on the invalid externalGasLimit... had to make a very small change instead.

test/TestBatchForwarder.test.ts
drortirosh: again: default gas was too different from actual gas we pass. they must be close enough , so better put the same value.

contracts/RelayHub.sol
drortirosh: (also, didn't really touch it: I renamed the "initialGas" parameter to its proper name, "externalGasLimit", since there was no point giving it a new name..

contracts/RelayHub.sol
forshtat: good enough

test/TestBatchForwarder.test.ts
forshtat: oh, so all these transactions had a wrong gas limit parameter? I see

forshtat (approved):

PR #610, opened 2021-03-29
OG-474 update dockers build scripts

drortirosh:


forshtat:
@drortirosh did you test these? I can't really review this change, can I?

forshtat (approved):

drortirosh:
> @drortirosh did you test these? I can't really review this change, can I?

The main issue that they can be built, using **dbuild.sh** in each of the folders relaydc, jsrelay, prefixrouter, cli. 
(there is no "master" script. each script doesn't "publish" the generated docker, just dump on screen its command, so I never just build them all blindly)

The cli can be executed using the "**./dockers/cli/gsn**" wrapper script  (after dbuild, of course.)

The relayer can be tested with:

`docker run --rm -ti -v /var/run/docker.sock:/var/run/docker.sock -e ROOT=$PWD -v $PWD/.env:/.env opengsn/relaydc up`
in a folder with "**.env**" and "**config/**" . 

note that each of the above commands can work even without building - but it would attempt to download the images.
to make sure you're using locally built images, do `docker images|grep opengsn` and remove them e.g. with
`docker rmi opengsn/cli:2.1.0 opengsn/cli:latest`
(or use [this](https://stackoverflow.com/questions/32490229/how-can-i-delete-docker-images-by-tag-preferably-with-wildcarding) to rename by wildcard.. )

PR #629, opened 2021-04-18
use --gaslimit including versionRegistry.addVersion

drortirosh:


packages/cli/src/CommandsLogic.ts
```diff
@@ -345,7 +345,7 @@ export class CommandsLogic {
 
     const regInstance = await this.getContractInstance(VersionRegistryAbi, {}, deployOptions.registryAddress, Object.assign({}, options), deployOptions.skipConfirmation)
     if (deployOptions.registryHubId != null) {
-      await regInstance.methods.addVersion(string32(deployOptions.registryHubId), string32('1'), rInstance.options.address).send({ from: deployOptions.from })
+      await regInstance.methods.addVersion(string32(deployOptions.registryHubId), string32('1'), rInstance.options.address).send({ ...options })
```
forshtat, line 348: this gasLimit will be 5M for an addVersion... you sure? also, as consistency is more important then correctness IMO, please use `Object.assign` or change all uses to `...`. Thanks!

forshtat (changes requested):

packages/cli/src/CommandsLogic.ts
drortirosh: I prefer the shorter idiom - its not less (or more) understandable that Object.assign({},x) format - but its cleaner and shorter...

do you have a better suggestion for gas? I don't want another parameter, and I don't have a good value to hard-code..

forshtat (approved):

PR #632, opened 2021-05-03
OG-513 cache registered relayers events

drortirosh:
- fixed the dead timeout on mumbai, by caching relayHub events
- scan SM managers from first block of those events, not from fromBlock:1
- improve performance on any network (reduce repeated queries)

required relayer configuration modification:
- re-register every lookup window, even if processed  requests
(TransactionRelayed )
(this is just one monthly required TX, and is required anyway for a
relayer that doesn't have any traffic)

### DIFF
--- a/packages/provider/src/KnownRelaysManager.ts
+++ b/packages/provider/src/KnownRelaysManager.ts
@@ -44,6 +44,8 @@ export class KnownRelaysManager {
   private readonly scoreCalculator: AsyncScoreCalculator
 
   private latestScannedBlock: number = 0
+  cachedLatestEvents: EventData[] = []
+
   private relayFailures = new Map<string, RelayFailureInfo[]>()
 
   public relayLookupWindowParts: number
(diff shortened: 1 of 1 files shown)


packages/provider/src/KnownRelaysManager.ts
```diff
@@ -44,6 +44,8 @@ export class KnownRelaysManager {
   private readonly scoreCalculator: AsyncScoreCalculator
 
   private latestScannedBlock: number = 0
+  cachedLatestEvents: EventData[] = []
```
forshtat, line 47: Caching is dangerous so I suggest not doing it 1 day before release
Lets just look for registration events for a way smaller window

packages/provider/src/KnownRelaysManager.ts
```diff
@@ -143,11 +149,39 @@ export class KnownRelaysManager {
...
 
-  async _fetchRecentlyActiveRelayManagers (): Promise<Set<Address>> {
+  async _getLatestRelayHubEvents (eventNames?: string[], managers?: Set<Address>): Promise<EventData[]> {
     const toBlock = await this.contractInteractor.getBlockNumber()
-    const fromBlock = Math.max(0, toBlock - this.config.relayLookupWindowBlocks)
 
-    const relayEvents: any[] = await this.getPastEventsForHub(fromBlock, toBlock)
+    let relayEvents: EventData[]
+    if (toBlock - this.latestScannedBlock < 10) {
+      // re-use the same cached events for ~2 minutes.
```
forshtat, line 157: When targeting alternative networks hard-coding time-per-block assumptions is not feasible

packages/provider/src/KnownRelaysManager.ts
```diff
@@ -143,11 +149,39 @@ export class KnownRelaysManager {
...
+      relayEvents = this.cachedLatestEvents
+    } else {
+      const fromBlock = Math.max(0, toBlock - this.config.relayLookupWindowBlocks)
```
forshtat, line 160: So as I was suggesting, registration lookup window should be separate (and few times bigger) than any event lookup window

packages/provider/src/KnownRelaysManager.ts
```diff
@@ -143,11 +149,39 @@ export class KnownRelaysManager {
...
+
+    if (eventNames != null) {
+      const nameSet = new Set(eventNames)
+      relayEvents = relayEvents.filter(event => nameSet.has(event.event))
+    }
+
+    if (managers != null) {
+      const lowercaseSet = new Set(Array.from(managers).map(x => x.toLowerCase()))
+      relayEvents = relayEvents.filter(event => lowercaseSet?.has(event.returnValues.relayManager.toLowerCase()))
+    }
```
forshtat, line 175: You have an 'infrastructure' to only query these events. I don't think we are saving time by getting events we don't need and then keeping them for 10 blocks worth of time. I suggest we keep using log filters.

forshtat (changes requested):

forshtat:
Adding caching this late is deemed unwise. Will delay this change and use a workaround in https://github.com/opengsn/gsn/pull/634 so far.

PR #637, opened 2021-05-05
OG-515 support pagination for all log queries

forshtat:


packages/common/src/ContractInteractor.ts
```diff
@@ -397,16 +401,85 @@ export class ContractInteractor {
...
-    return await this._getPastEvents(stakeManager.contract, names, extraTopics, options)
+    return await this._getPastEventsPaginated(stakeManager.contract, names, extraTopics, options)
   }
 
   async getPastEventsForPenalizer (names: EventName[], extraTopics: string[], options: PastEventOptions): Promise<EventData[]> {
-    return await this._getPastEvents(this.penalizerInstance.contract, names, extraTopics, options)
+    return await this._getPastEventsPaginated(this.penalizerInstance.contract, names, extraTopics, options)
+  }
+
+  async getPagesForBlockWindow (fromBlock: BlockNumber = 1, toBlock?: BlockNumber): Promise<number> {
```
shahafn, line 416: Pages *Per* Block Window?

packages/common/src/ContractInteractor.ts
```diff
@@ -397,16 +401,85 @@ export class ContractInteractor {
...
+    }
+    return ret
+  }
+
+  /**
+   * Splits requested range into pages to avoid fetching too many blocks at once.
+   * In case 'getLogs' returned with a common error message of "more than X events" dynamically decrease page size.
+   */
+  async _getPastEventsPaginated (contract: any, names: EventName[], extraTopics: string[], options: PastEventOptions): Promise<EventData[]> {
+    let pagesCurrent: number = await this.getPagesForBlockWindow(options.fromBlock, options.toBlock)
```
shahafn, line 458: pagesPerBlockWindow?

packages/common/src/ContractInteractor.ts
```diff
@@ -397,16 +401,85 @@ export class ContractInteractor {
...
+    if (typeof fromBlock !== 'number' || typeof toBlock !== 'number') {
+      throw new Error('ContractInteractor:getPartsForBlockWindow: only number supported for block range when using pagination')
+    }
+    const rangeSize = toBlock - fromBlock + 1
+    const pagesForBlockWindow = Math.max(Math.ceil(rangeSize / this.maxPageSize), 1)
+    this.logger.info(`Splitting request for ${rangeSize} blocks into ${pagesForBlockWindow} smaller paginated requests!`)
+    return pagesForBlockWindow
+  }
+
+  splitRange (fromBlock: BlockNumber, toBlock: BlockNumber, parts: number): Array<{ fromBlock: BlockNumber, toBlock: BlockNumber }> {
```
shahafn, line 434: This function is too similar to `getPagesForBlockWindow`, and they have the same calculation inside - which is generally a bad thing. In this case, you actually have a bug that if you provide to any `getPastEvents` function in interactor `toBlock = 'latest'`, it will throw in line 441, even though `getPagesForBlockWindow` supports it.

packages/common/src/ContractInteractor.ts
forshtat: No, opposite of that. "Per" implies there are multiple 'windows' consumed, like "miles per gallon", "kilometers per hour" etc.
Here there is one "window" being split into pages. 
I will rename "window" to range and the method to `getLogsPagesForRange` to make it more clear.

dockers/relaydc/docker-compose.yml
```diff
@@ -24,15 +24,15 @@ services:
     restart: always
     ports: [ '8080:80' ] #bypass https-portal
     # add line for each relay
-    command: 
-      - gsn1 
+    command:
```
drortirosh, line 27: why this space changes?

packages/dev/package.json
```diff
@@ -4,7 +4,7 @@
   "license": "MIT",
   "main": "dist/GsnTestEnvironment.js",
   "files": [
-    "dist/*",
+    "dist/src/*",
```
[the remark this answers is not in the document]
forshtat: I have changed the `tsconfig.js` to include test files, so unless we want `./dist/test` in `@opengsn/dev`, they should be excluded.

packages/dev/test/PaymasterCommitment.test.ts
```diff
@@ -202,6 +202,7 @@ contract('Paymaster Commitment', function ([_, relayOwner, relayManager, relayWo
 
       }, sharedRelayRequestData, chainId, forwarderInstance)
       const gasAndDataLimits = await paymasterContract.getGasAndDataLimits()
+      // @ts-ignore
```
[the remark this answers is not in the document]
forshtat: Not always. Some errors do not manifest while running via `ts-node` and then it failed for me. Might be fixed already, but needs separate verification.

shahafn (approved):

packages/common/src/ContractInteractor.ts
forshtat: They do have a same calculation but do the opposite things. `splitRange` has existed before this change and is responsible for translation of 1 range into X smaller ranges. `getPagesForBlockWindow` determines X. The common code is boilerplate and can be extracted, generally speaking.

packages/common/src/ContractInteractor.ts
```diff
@@ -84,11 +86,13 @@ export class ContractInteractor {
 
   constructor (
     {
+      maxPageSize,
```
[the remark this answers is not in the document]
forshtat: We did not use this syntax previously so I think we don't have to start now. It is nice though and we can transition to it in the near future.

packages/dev/package.json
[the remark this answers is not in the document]
forshtat: Also, you made 17 writing errors in this sentence! 😂

packages/common/src/ContractInteractor.ts
```diff
@@ -397,16 +401,85 @@ export class ContractInteractor {
...
+    // save 'getBlockNumber' roundtrip for a known max value
+    if (this.maxPageSize === Number.MAX_SAFE_INTEGER) {
+      return 1
+    }
+    if (toBlock === 'latest' || toBlock == null) {
+      toBlock = await this.getBlockNumber()
+    }
+    // noinspection SuspiciousTypeOfGuard - known false positive
+    if (typeof fromBlock !== 'number' || typeof toBlock !== 'number') {
+      throw new Error('ContractInteractor:getPartsForBlockWindow: only number supported for block range when using pagination')
```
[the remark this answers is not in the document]
forshtat: Fixed

shahafn (approved):

Issue #642, opened 2021-05-25
@opengsn/cli doesn't create any binary symlink


cedoor:
@opengsn/cli doesn't create any symlink in the `node_modules/.bin` folder. This is probably because the [`bin`](https://docs.npmjs.com/cli/v6/configuring-npm/package-json#bin) field is missing in the `package.json` file.

PR #643, opened 2021-05-25
Add bin field in package.json file

cedoor:
The `bin` field was missing in the `package.json` file.  This field allows npm to simlink correctly the binary file.

#642


drortirosh:
Thanks. We really missed that.
But for the record, please change the title of the PR (and git commit) to the symptom (gsn cli command no longer works), and leave the fix in the description

forshtat (approved):

PR #647, opened 2021-05-30
Fixing server registration

shahafn:
Now server validates stake status with hub before trying to add
workers/register


drortirosh (approved):

forshtat:
@shahafn @drortirosh I see that this PR removes the 'Splitting request for ${rangeSize} blocks' logs entirely, doesn't it? Don't you think these logs might be useful for us later? Maybe it should remain but only print if rangesCount > 1 ?

PR #651, opened 2021-06-06
Inclue 'paymasters' as part of the main GSN repo

forshtat:


packages/paymasters/src/HashCashApproval.ts
```diff
@@ -0,0 +1,98 @@
...
+ * @param interval call the callback every that many iterations
+ * @param callback async callback to call. return "false" to abort. true to continue
+ * @return the approvalData value (bytes32 hash, uint256 counter)
+ */
+export async function calculateHashcash (senderAddress: string, senderNonce: string, difficulty: any, interval?: number, callback?: any): Promise<string> {
+  const diffMax = toBN(1).shln(256 - difficulty)
+  let hashNonce = 0
+  let intervalCount = 0
+  while (true) {
+    // @ts-ignore
```
drortirosh, line 22: why changed all ts-error to ignore ? ts-error **requires** an error (and thus should be removed when there's no more problem), while ts-ignore well, ignores everything.

packages/paymasters/package.json
```diff
@@ -0,0 +1,52 @@
...
+    "ethereumjs-util": "^6.1.0"
+  },
+  "devDependencies": {
+    "solhint": "^3.3.2",
+    "@types/chai": "^4.2.12",
+    "@types/mocha": "^8.2.0",
+    "@types/web3": "1.2.2",
+    "@types/ethereumjs-util": "^6.1.0",
+    "ethereumjs-tx": "2.1.2",
+    "solidity-string-utils": "^0.0.4"
```
drortirosh, line 42: remove. this is a test library I have. 
it adds a StringUtils library, which allows you to do
`"asdasd".concat(something).concat("varname", varname)`

(useful during debugging, but I can't easily  mark it as "NOGIT" in the package.json and it slipped in)

packages/paymasters/package.json
```diff
@@ -0,0 +1,52 @@
+{
+  "name": "@opengsn/paymasters",
+  "license": "MIT",
```
drortirosh, line 3: GPL

packages/paymasters/package.json
```diff
@@ -0,0 +1,52 @@
...
+  "publishConfig": {
+    "access": "public"
+  },
+  "dependencies": {
+    "@openzeppelin/contracts": "^3.4.0-solc-0.7",
+    "@opengsn/dev": "^2.2.2",
+    "@opengsn/common": "^2.2.2",
+    "@opengsn/provider": "^2.2.2",
+    "@opengsn/contracts": "^2.2.2",
+    "@typechain/truffle-v5": "^4.0.1",
```
drortirosh, line 32: I think types are only dev-dependencies. we should "export" such deps (general issue, not only this package.

packages/paymasters/tsconfig.json
```diff
@@ -0,0 +1,14 @@
+{
+  "extends": "../../tsconfig.packages.json",
+  "include": [
+    "../paymasters/src/**/*.ts",
```
drortirosh, line 4: I don't understand this. this is the `paymasters` folder. why not `./src/**/*.ts`

packages/paymasters/package.json
forshtat: not used in any actual test? removing

packages/paymasters/package.json
forshtat: Right... And put a LICENSE file, I guess?..

packages/paymasters/package.json
forshtat: Moved `@typechain/truffle-v5` to `devDependencies` as it should not be referenced AFAIK.

packages/paymasters/tsconfig.json
forshtat: Idea refactoring outsmarted itself, fixing

packages/paymasters/contracts/AcceptEverythingPaymaster.sol
```diff
@@ -0,0 +1,41 @@
+//SPDX-License-Identifier: MIT
+pragma solidity ^0.7.6;
+pragma experimental ABIEncoderV2;
```
drortirosh, line 3: nitpic: its `pragma abicoder v2;`
(not critical, since I think we'll have to release 0.8 contracts soon...)

packages/paymasters/tsconfig.json
```diff
@@ -0,0 +1,14 @@
+{
+  "extends": "../../tsconfig.packages.json",
+  "include": [
+    "./src/**/*.ts",
+    "./test/**/*.ts",
+    "./types/truffle-contracts/index.d.ts",
```
drortirosh, line 6: all other projects reference `contracts/types`. why this project has its own?

PR #652, opened 2021-06-07
Verify datas lengths

forshtat:


packages/paymasters/contracts/HashcashPaymaster.sol
```diff
@@ -36,10 +36,11 @@ contract HashcashPaymaster is AcceptEverythingPaymaster {
     virtual
     relayHubOnly
     returns (bytes memory, bool revertOnRecipientRevert) {
+        (maxPossibleGas, signature);
 
-        (relayRequest, approvalData, maxPossibleGas, signature);
+        require(approvalData.length == 64, "invalid approval data length");
```
drortirosh, line 41: please name the field. the best use of a comment is for search. by renaming (event adding a space) you break the possibility to use it directly
e.g.
`invalid approvalData length`

packages/paymasters/contracts/VerifyingPaymater.sol
```diff
@@ -37,7 +37,8 @@ contract VerifyingPaymaster is Ownable, BasePaymaster {
     returns (bytes memory context, bool revertOnRecipientRevert) {
         (signature, maxPossibleGas);
 
-        require(approvalData.length == 65, "invalid approvalData signature");
+        require(approvalData.length == 65, "invalid approval data length");
```
drortirosh, line 40: it is a signature. why remove that fact from the error message?

packages/paymasters/contracts/TokenPaymaster.sol
```diff
@@ -102,7 +96,11 @@ contract TokenPaymaster is BasePaymaster {
     virtual
     relayHubOnly
     returns (bytes memory context, bool revertOnRecipientRevert) {
-        (relayRequest, signature, approvalData, maxPossibleGas);
+        (signature);
+
+        require(approvalData.length == 0, "invalid approval data length");
+        require(relayRequest.relayData.paymasterData.length == 32, "invalid paymaster data length");
```
drortirosh, line 102: here its even worse: `invalid paymaster data length` give no clue as to what it should contain (in this case, a uniswap address, which is used in another place (in _getToken)
please restore the meaningful error messages.

drortirosh (changes requested):

drortirosh (approved):

PR #659, opened 2021-06-21
OG-566 Hub penalize fix

shahafn:


drortirosh (approved):

drortirosh (changes requested):
the hub penalization fix looks OK, but it depends on a complex boosting issue, which looks much more complex (and unrelated..)

shahafn:
> the hub penalization fix looks OK, but it depends on a complex boosting issue, which looks much more complex (and unrelated..)

Yes, once we merge the previous PR, I'll rebase this one and you'll see only the diff.

drortirosh (approved):

drortirosh (changes requested):
missing version change

shahafn:
> missing version change

Do you mean contract version?

packages/provider/src/GSNConfigurator.ts
```diff
@@ -37,7 +37,7 @@ export const defaultGsnConfig: GSNConfig = {
   sliceSize: 3,
   relayTimeoutGrace: DEFAULT_RELAY_TIMEOUT_GRACE_SEC,
   methodSuffix: '',
-  requiredVersionRange: '2.2.0',
+  requiredVersionRange: '2.2.3',
```
drortirosh, line 40: this is the one version we should not update: you basically say that you don't want to run against contracts with older version.

drortirosh (approved):

Issue #660, opened 2021-07-02
change `getChainID` from pure to view


hack3r-0m:
https://github.com/opengsn/gsn/blob/master/packages/contracts/src/utils/GsnEip712Library.sol#L113-L118

since solidity 0.8.0 chainId retrieval should be `view` and not `pure`
reference: https://github.com/ethereum/solidity/releases/tag/v0.8.0

and if contracts are upgraded to 0.8 in the future, it is good to replace it with `block.chainid`

drortirosh:
Yes, this is part of the changes needed when converting to 0.8 ..

PR #661, opened 2021-07-02
change `getChainID` from pure to view (#660)

hack3r-0m:


shahafn:
Thanks for the fix :+1: 
We'll discuss and introduce it if needed soon

shahafn:
@hack3r-0m it's now part of PR #674

PR #664, opened 2021-07-16
Mumbai support

Sulejman:
Updating truffle wallet provider removes issue with replay protected transactions when using cli to deploy relay for mumbai/polygon network.


shahafn:
Thanks for bringing that to our attention. We will look at the issue and fix it

forshtat:
It appears that GSN is currently running on Mumbai network. Ping me if this has to be reopened.

PR #668, opened 2021-07-27
Update ethereumjs tx

shahafn:


packages/dev/test/provider/RelayClient.test.ts
```diff
@@ -541,7 +541,7 @@ contract('RelayClient', function (accounts) {
         maxPageSize,
         deployment: { paymasterAddress: paymaster.address }
       }).init()
-      const badHttpClient = new BadHttpClient(logger, false, false, false, pingResponse, '0x123')
+      const badHttpClient = new BadHttpClient(logger, false, false, false, pingResponse, '0xc6808080808080')
```
drortirosh, line 544: what is this hex value, and how comes it appears in 2 unrelated tests?

drortirosh (approved):

packages/dev/test/provider/RelayClient.test.ts
shahafn: Previously, it was '0x123', which is not a valid rlp encoding of a transaction. Since the constructor didn't complain and just filled the fields with defaults, it passed. '0xc6808080808080' is `rlp.encode([0,0,0,0,0,0]) ` since Transaction now expects a valid rlp encoding.

PR #671, opened 2021-08-10
Update doc comment on _msgData

tzann:
The doc comment in BaseRelayRecipient.sol is outdated compared to IRelayRecipient.sol. It also trails off in the middle of a sentence.


shahafn:
@tzann Thank you very much!

shahafn (approved):

PR #674, opened 2021-08-11
OG-587 Solidity 0.8 upgrade

shahafn:


packages/contracts/src/BaseRelayRecipient.sol
```diff
@@ -1,6 +1,6 @@
 // SPDX-License-Identifier: GPL-3.0-only
 // solhint-disable no-inline-assembly
-pragma solidity >=0.7.6;
+pragma solidity ^0.8.0;
```
drortirosh, line 3: interfaces (and base classes) are better set using `>=`, so that they can be used with later versions.
If they don't change, can even leave them as `>= 0.7.6`, so that clients with earlier version can work.

packages/contracts/truffle.js
```diff
@@ -6,7 +6,7 @@ module.exports = {
   contracts_directory: './src',
   compilers: {
     solc: {
-      version: '0.7.6',
+      version: '0.8.7-nightly.2021.8.9+commit.74c804d8',
```
drortirosh, line 9: 0.8.7 is released (no need for "nightly") - with `block.basefee` (yay!)

packages/dev/test/RelayHubPenalizations.test.ts
```diff
@@ -481,7 +481,7 @@ contract('RelayHub Penalizations', function ([_, relayOwner, committer, nonCommi
           const method = await commitPenalizationAndReturnMethod(
             penalizer.contract.methods.penalizeRepeatedNonce, txDataSigA.data, txDataSigA.signature, txDataSigB.data, txDataSigB.signature, relayHub.address)
           await expectRevert(
-            method.send({ from: reporterRelayManager }),
+            method.send({ from: reporterRelayManager, gas: 5e5 }),
```
drortirosh, line 484: forward looking: EIP1559's support in `hardhat node` as minimum basefee of 8 gwei, so if you specify gas, at least put a meaningful number like 10gwei, to avoid changing it again...)

packages/paymasters/contracts/helpers/TestProxy.sol
```diff
@@ -30,7 +30,7 @@ contract TestProxy is BaseRelayRecipient, Ownable  {
         require(success, string(ret));
     }
 
-    function _msgSender() internal override(Context, BaseRelayRecipient) view returns (address payable) {
```
drortirosh, line 33: msg`.sender` is no longer payable in solidity 0.8, so `_msgSender()` should be non-payable just the same.

packages/contracts/src/BaseRelayRecipient.sol
superical: I agree that it would be good if we keep them as `>=0.7.6` for the contracts if possible just so at least it works with the earlier version. Some projects may still rely on an earlier version.

superical:
@shahafn Can I check will this update be released soon so that we can start using with v0.8? The current version of v2.2.2 still doesn’t include this PR.

PR #682, opened 2021-08-23
OG-597 Create a BaseRelayFee bidding mode for GSN

forshtat:
In Arbitrum, gas limit can be divided into 4 components: L1Transaction, L1Calldata, L2Storage and L2Computation.  It means the same Solidity code cannot be used to calculate gas used both on EVM and on Arbitrum.

For this change we stop calculating gasUsed, maxPossibleGas, gasUseWithoutPost values on-chain. We do not need to remove those from code, just setting request’s gasPrice to 0 makes these values multiply by zero eventually. Paymasters should not rely on them.

Perform all the calculations about the transaction cost on the client. Put this number in the baseRelayFee field. RelayServer makes the same calculations and decides whether he is making a profit by sending this transaction.


packages/contracts/src/interfaces/IRelayHub.sol
```diff
@@ -26,6 +26,10 @@ interface IRelayHub {
         uint256 dataGasCostPerByte;
         // relayCalls() minimal gas overhead when calculating cost of putting tx on chain.
         uint256 externalCallDataCostOverhead;
+        // maximum cost of sending a calldata byte, prevents Relays from malicious claim of enormous external gas limit
+        uint256 maxGasCostPerCalldataByte;
+        // if set to true, requested gas price will be forced to be 0, making gas used calculation redundant
+        bool baseRelayFeeBidMode;
```
drortirosh, line 32: what's this? do we need a config param, or simply a zero fee value in the relayrequest?

packages/contracts/src/RelayHub.sol
```diff
@@ -54,29 +52,11 @@ contract RelayHub is IRelayHub, Ownable {
...
-        uint256 _maxWorkerCount,
-        uint256 _gasReserve,
-        uint256 _postOverhead,
-        uint256 _gasOverhead,
-        uint256 _maximumRecipientDeposit,
-        uint256 _minimumUnstakeDelay,
-        uint256 _minimumStake,
-        uint256 _dataGasCostPerByte,
-        uint256 _externalCallDataCostOverhead
+        RelayHubConfig memory _config
```
drortirosh, line 55: did you try to use a console to deploy a relayer this way?

packages/common/src/ContractInteractor.ts
```diff
@@ -586,9 +615,111 @@ export class ContractInteractor {
     return await this.web3.eth.estimateGas(gsnTransactionDetails)
   }
 
+  async estimateGasWithoutCalldata (gsnTransactionDetails: GsnTransactionDetails): Promise<number> {
+    const originalGasEstimation = await this.web3.eth.estimateGas(gsnTransactionDetails)
+    const calldataGasCost = this.calculateCalldataCost(gsnTransactionDetails.data)
+    const adjustedEstimation = originalGasEstimation - calldataGasCost
+    this.logger.info(`estimateGasWithoutCalldata: original estimation: ${originalGasEstimation}; calldata cost: ${calldataGasCost}; adjusted estimation: ${adjustedEstimation}`)
```
drortirosh, line 622: this is debug.

packages/common/src/Environments.ts
```diff
@@ -26,19 +22,22 @@ const defaultStakeManagerMaxUnstakeDelay: number = 10000000
...
+  gasOverhead: 32048,
+  postOverhead: 12036,
   gasReserve: 100000,
   maxWorkerCount: 10,
   minimumStake: 1e18.toString(),
   minimumUnstakeDelay: 1000,
   maximumRecipientDeposit: 2e18.toString(),
   dataGasCostPerByte: 13,
-  externalCallDataCostOverhead: 22402
+  maxGasCostPerCalldataByte: 16,
```
shahafn, line 38: Please rename to gtxdatanonzero as this is exactly what it is.

forshtat:
Closing this PR as this should not be merged as-is. A separate PR will be derived from it.

PR #683, opened 2021-08-23
OG-588 Hardhat migration

shahafn:


package.json
```diff
@@ -58,7 +58,10 @@
     "lerna-rm-dist": "lerna run rm-dist",
     "lerna-run-test": "lerna run test --stream --no-prefix",
     "lerna-run-test-fork": "lerna run test-fork --stream --no-prefix",
-    "lerna-run-test-with-testrpc": "run-with-testrpc --chainId 1337 --hardfork 'istanbul' --port 8544 --gasLimit 12000000 --defaultBalanceEther 1000 --deterministic 'yarn lerna-run-test'",
+    "lerna-run-test1": "lerna run test1 --stream",
```
forshtat, line 61: is this task needed?

packages/cli/src/GsnTestEnvironment.ts
```diff
@@ -75,7 +77,7 @@ class GsnTestEnvironmentClass {
       stake: ether('1'),
       funds: ether('1'),
       relayUrl: relayUrl,
-      gasPrice: '1e9',
+      gasPrice: 1e9.toString(),
```
forshtat, line 80: If I understand this change correctly, ganache didn't care about the inputs but hardhat does?

packages/common/src/Utils.ts
```diff
@@ -95,12 +106,13 @@ export async function getEip712Signature (
       // @ts-ignore
       method = web3.currentProvider.send
     }
-    method.bind(web3.currentProvider)({
-      method: 'eth_signTypedData' + methodSuffix,
+    const paramBlock = {
+      method: `eth_signTypedData${methodSuffix}`,
       params: [senderAddress, dataToSign],
-      from: senderAddress,
+      jsonrpc: '2.0',
```
forshtat, line 112: no, wait, this is a problem, how do you not need a 'from' address? What if the provider controls multiple addresses, which is totally part of an RPC spec afaik?

packages/contracts/src/test/TestPaymasterConfigurableMisbehavior.sol
```diff
@@ -116,7 +116,7 @@ contract TestPaymasterConfigurableMisbehavior is TestPaymasterEverythingAccepted
         if (expensiveGasLimits) {
             uint sum;
             //memory access is 700gas, so we waste ~50000
-            for ( int i=0; i<60000; i+=700 ) {
+            for ( int i=0; i<100000; i+=700 ) {
```
forshtat, line 119: is this loop actually cheaper on hardhat? :-)

packages/dev/package.json
```diff
@@ -15,8 +15,10 @@
     "truffle-compile": "truffle compile",
     "lint": "eslint -f unix .",
     "test": "lerna run tsc && truffle --compile-all --network npmtest test",
+    "truffle-console": "truffle console",
     "truffle-test": "truffle test --compile-all ",
     "test1": "truffle --network npmtest test",
+    "test2": "hardhat --network npmtest test",
```
forshtat, line 21: this is getting out of control...

packages/dev/test/HttpWrapper.test.ts
```diff
@@ -29,16 +29,10 @@ describe('HttpWrapper', () => {
...
-    const res =
-      // @ts-ignore
-      http.sendPromise(web3.currentProvider.host, {
-        jsonrpc: '2.0',
-        method: 'net_version',
-        id: 123
-      })
-    return await expect(res).to.be.eventually.rejectedWith('timeout of 1ms exceeded')
+  it('should pass timeout to provider', async () => {
+    // This test should be removed. It checks axios functionality.
```
forshtat, line 33: lets remove it then; asserting a value is not cool
also, why can't you keep the expect.to.be.reverted check?

packages/dev/test/PaymasterCommitment.test.ts
```diff
@@ -185,10 +185,10 @@ contract('Paymaster Commitment', function ([_, relayOwner, relayManager, relayWo
       // gasPrice is '1', so price=gasUsed...
       expectEvent(res, 'TransactionRelayed', { status: '0' })
 
-      const gasUsed = res.receipt.gasUsed
+      // const txCost = toBN(res.receipt.gasUsed).mul(toBN(gasPrice))
```
forshtat, line 188: why is it commented out?

packages/dev/test/PaymasterCommitment.test.ts
```diff
@@ -330,7 +330,8 @@ contract('Paymaster Commitment', function ([_, relayOwner, relayManager, relayWo
       expectEvent(res, 'TransactionRelayed', { status: '0' })
 
       const paymasterPaid = paymasterBalance.sub(await relayHubInstance.balanceOf(paymaster)).toNumber()
-      assert.closeTo(paymasterPaid, parseInt(gasUsed) + 15000, 50)
+      // TODO understand why there's a 2100 gas decrease (cost of sload?)
+      assert.closeTo(paymasterPaid - parseInt(gasUsed), 17100, 50)
```
forshtat, line 334: I do not understand this test. Also, wasn't gas refund removed in the latest hard-fork? Does hardhat still refund gas?

packages/dev/test/RelayHub.test.ts
```diff
@@ -82,20 +82,20 @@ contract('RelayHub', function ([_, relayOwner, relayManager, relayWorker, sender
...
+      expectEvent.inLogs(res.logs, 'Deposited', {
         paymaster,
         from: sender,
         amount
       })
-
+      const txCost = (new BN(res.receipt.gasUsed)).mul(gasPrice)
       expect(await relayHubInstance.balanceOf(paymaster)).to.be.bignumber.equal(amount)
-      expect(await senderBalanceTracker.delta()).to.be.bignumber.equal(amount.neg())
+      expect(await senderBalanceTracker.delta()).to.be.bignumber.equal(amount.neg().sub(txCost))
```
forshtat, line 98: .add().neg() maybe? took me a second to understand wtf is going on here

packages/dev/test/ServerTestUtils.ts
```diff
@@ -56,7 +56,8 @@ export async function assertRelayAdded (transactionHashes: PrefixedHexString[],
 
 export async function getTotalTxCosts (transactionHashes: PrefixedHexString[], gasPrice: string): Promise<BN> {
   const receipts = await resolveAllReceipts(transactionHashes)
-  return receipts.map(r => toBN(r.gasUsed).mul(toBN(gasPrice))).reduce(
+  // @ts-ignore
```
forshtat, line 59: I don't see what has to be ts-ignored here, is it needed?

packages/dev/test/common/StatisticsManager.test.ts
```diff
@@ -46,11 +46,11 @@ contract('StatisticsManager', function (accounts) {
     await env.relayServer.createRelayTransaction(await env.createRelayHttpRequest())
     await env.relayServer.createRelayTransaction(await env.createRelayHttpRequest())
     await env.relayServer.createRelayTransaction(await env.createRelayHttpRequest())
+    await misbehavingPaymaster.setRevertPreRelayCallOnEvenBlocks(true)
     block = await web3.eth.getBlockNumber()
     if (block % 2 !== 0) {
       await evmMine()
     }
```
forshtat, line 53: how does this help? :-)

packages/dev/test/provider/AccountManager.test.ts
```diff
@@ -21,7 +21,7 @@ contract('AccountManager', function (accounts) {
   const address = '0x982a8CbE734cb8c29A6a7E02a3B0e4512148F6F9'
   const privateKey = '0xd353907ab062133759f149a3afcb951f0f746a65a60f351ba05a3ebf26b67f5c'
   const config = configureGSN({
-    methodSuffix: '',
+    methodSuffix: '_v4',
```
forshtat, line 24: the default should work, shouldn't it?

packages/dev/test/provider/RelayClient.test.ts
```diff
@@ -139,7 +141,8 @@ contract('RelayClient', function (accounts) {
     logger = createClientLogger(loggerConfiguration)
     relayClient = new RelayClient({ provider: underlyingProvider, config: gsnConfig })
     await relayClient.init()
-    gasLess = await web3.eth.personal.newAccount('password')
+    gasLessAccount = relayClient.newAccount()
+    gasLess = toChecksumAddress(gasLessAccount.address)
```
forshtat, line 145: this is a bit of an issue. in all cases you replace signing with provider-controlled account (major case) with an ephemeral key (edge case)

I suggest this test transfers all ether from account 9 to a single bit, and then the test is done again with actual RPC call to sign transactions.

packages/dev/test/provider/RelayProvider.test.ts
```diff
@@ -143,6 +144,9 @@ contract('RelayProvider', function (accounts) {
       // so changing the global one is not enough.
       // @ts-ignore
       TestRecipient.web3.setProvider(relayProvider)
+      gasLessAccount = relayProvider.newAccount()
```
forshtat, line 147: same here

packages/dev/truffle.js
```diff
@@ -1,7 +1,12 @@
 require('ts-node/register/transpile-only')
 
 const HDWalletProvider = require('@truffle/hdwallet-provider')
-const mnemonic = 'digital unknown jealous mother legal hedgehog save glory december universe spread figure custom found six'
+let mnemonic = 'digital unknown jealous mother legal hedgehog save glory december universe spread figure custom found six'
+
+if (process.env.MNEMONIC_FILE) {
+  console.error(`== reading mnemonic file: ${process.env.MNEMONIC_FILE}`)
+  mnemonic = require('fs').readFileSync(process.env.MNEMONIC_FILE, 'utf-8').replace(/(\r\n|\n|\r)/gm, '')
```
forshtat, line 8: why is it added to a truffle config?

patches/truffle+5.1.26.patch
```diff
@@ -10,3 +14,17 @@ index 5991472..eab1676 100755
...
+       res && typeof res === "object" && res.error && res.error.data;
+     const isString =
+       res && typeof res === "object" && typeof res.result === "string";
++    const isHardHat =
++      res && typeof res === "object" && res.error && res.error.message && !res.error.data;
++
++    if (isHardHat) {
++      if (res.error.message.includes('Transaction reverted without a reason string')) return 'revert'
++      return res.error.message.match('VM Exception while processing transaction: reverted with reason string \'(.*)\'')[1]
++    }
```
forshtat, line 27: can you elaborate and write a very descriptive comment in this patchfile? 
I was hoping we can get rid of patchfiles one day...

scripts/hardhat-node
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+#start hardhat-node in a way that "run-with-testrpc" will work
+# (it looks up for "Listening on" string, to determine when the node is up)
+trap "pkill -9 -f 'hardhat node'" EXIT
```
forshtat, line 4: how does it work? is it needed on local environment?

forshtat (changes requested):

package.json
shahafn: I run it to run tests faster without compiling, when changing only tests.

packages/cli/src/GsnTestEnvironment.ts
shahafn: No, ganahce read it as 0x1e9 === 489, and it was fine. Hardhat has baseFee higher than that due to london fork, so it needs more than 489 gasPrice.

packages/common/src/Utils.ts
shahafn: The firsr param is the "from" address, see here:
https://eips.ethereum.org/EIPS/eip-712#specification-of-the-eth_signtypeddata-json-rpc

packages/contracts/src/test/TestPaymasterConfigurableMisbehavior.sol
shahafn: Probably cheaper due to berlin and london gas costs update

packages/dev/package.json
shahafn: That I can remove for now, I tried running hardhat as test framework instead of truffle as well, and that's a left over from that.

packages/dev/test/HttpWrapper.test.ts
shahafn: https://github.com/opengsn/gsn/blob/995647ebf8e34ac183d5b99c06c385bc1995d6dd/packages/dev/test/HttpWrapper.test.ts#L33

This test used to rely on a race condition that worked with ganahce since it's slow. Now it doesn't, since hardhat is quicker to respond.

packages/dev/test/PaymasterCommitment.test.ts
shahafn: Initially I fixed the gasPrice to be higher than 1 since it's needed post london, then I realized that the charge is being calculated with the client's gasPrice in request, and the relayer can simply give actual gasPrice which is higher. So, gasPrice of 1 removes the need to multiply everywhere as the comment says. 

TLDR; this line should be removed.

packages/dev/test/PaymasterCommitment.test.ts
shahafn: It wasn't removed, but changed criteria and reduced in some cases. see 
https://eips.ethereum.org/EIPS/eip-3529
https://eips.ethereum.org/EIPS/eip-2929

I talked to Dror about this test actually as well, and it's definitely a candidate for removal.

packages/dev/test/RelayHub.test.ts
shahafn: Same difference :laughing:

packages/dev/test/ServerTestUtils.ts
shahafn: receipt interface doesn't have "effectiveGasPrice", so ts complains, at least until web3 version update I guess.

packages/dev/test/common/StatisticsManager.test.ts
shahafn: That's a difference between ganahce and hardhat, setting revert on even blocks is advancing the block number as well.

packages/dev/test/provider/AccountManager.test.ts
shahafn: Another difference between ganahce and hardhat, method suffix needs to be "_v4" in hardhat node.

packages/dev/test/provider/RelayClient.test.ts
shahafn: I thought about that solution of emptying an account as well, but Dror suggested adding an account, so I did. I can change it though.

packages/dev/truffle.js
shahafn: I wanted to connect to a public network with truffle console, with my own deployer, to test something on kovan. Is it an issue?

patches/truffle+5.1.26.patch
shahafn: On eth_call, ganache returns error object that has message, code, data fields. Truffle reads the revert reason from res.error.data in this case. But hardhat returns error object only with message and code, so to return the revert reason, I needed to parse res.error.message.

In general, when eth_call returns an error, the spec doesn't specify a data field containing the revert reason, so truffle is tailored specifically for ganache here

scripts/hardhat-node
shahafn: trap command gets a bash command and signal to work on, so whenever this script exists, it kills "hardhat node" that the script itself ran, so we don't have a running node in the background once we finish.

hardhat.config.ts
```diff
@@ -0,0 +1,17 @@
+/**
+ * @type import('hardhat/config').HardhatUserConfig
+ */
+import '@nomiclabs/hardhat-truffle5'
+import '@nomiclabs/hardhat-web3'
+
+module.exports = {
+  solidity: '0.7.3',
```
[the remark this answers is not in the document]
shahafn: Yes that was הכנה למזגן קצת

hardhat.config.ts
```diff
@@ -0,0 +1,17 @@
...
+  networks: {
+    hardhat: { chainId: 1337 },
+    npmtest: { // used from "npm test". see package.json
+      verbose: process.env.VERBOSE,
```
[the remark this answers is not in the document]
shahafn: Weird, for me it doesn't complain. I can remove though

hardhat.config.ts
```diff
@@ -0,0 +1,17 @@
...
+      url: 'http://127.0.0.1:8544',
+      network_id: '*'
```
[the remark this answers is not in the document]
shahafn: I know, forgot to remove after copy pasta from truffle.js

packages/common/src/Utils.ts
shahafn: Yes that's what I said :)

packages/dev/test/RelayHubPenalizations.test.ts
```diff
@@ -49,9 +51,9 @@ contract('RelayHub Penalizations', function ([_, relayOwner, committer, nonCommi
   let transactionOptions: TxOptions
 
   let forwarder: string
-  const relayWorkerPrivateKey = Buffer.from('6370fd033278c143179d81c5526140625662b8daa446c22ee2d73db3707e620c', 'hex')
+  const relayWorkerPrivateKey = Buffer.from('92db14e403b83dfe3df233f83dfa3a0d7096f21ca9b0d6d6b8d88b2b4ec1564e', 'hex')
```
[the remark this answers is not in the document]
shahafn: It used to be a ganache (-d) account with ether, now it's a hardhat account.

packages/dev/test/ReputationFlow.test.ts
```diff
@@ -1,39 +1,37 @@
...
+    testRecipient = await TestRecipient.new(forwarderInstance.address)
 
     misbehavingPaymaster = await TestPaymasterConfigurableMisbehavior.new()
     await misbehavingPaymaster.setRevertPreRelayCallOnEvenBlocks(true)
     await misbehavingPaymaster.setTrustedForwarder(forwarderInstance.address)
     await misbehavingPaymaster.setRelayHub(relayHub.address)
     await misbehavingPaymaster.deposit({ value: web3.utils.toWei('1', 'ether') })
 
-    const relayProvider = await RelayProvider.newProvider({
+    relayProvider = await RelayProvider.newProvider({
```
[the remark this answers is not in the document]
shahafn: to have access to relayProvider in the test, to stub validateRelayCall

packages/dev/test/ServerTestEnvironment.ts
```diff
@@ -145,6 +144,7 @@ export class ServerTestEnvironment {
       config: mergedConfig
     })
     await this.relayClient.init()
+    this.gasLess = toChecksumAddress(this.relayClient.newAccount().address)
```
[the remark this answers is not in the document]
shahafn: You might be onto something :laughing:

packages/dev/test/provider/RelayProvider.test.ts
```diff
@@ -61,7 +61,7 @@ export async function prepareTransaction (testRecipient: TestRecipientInstance,
     relayData: {
       pctRelayFee: '1',
       baseRelayFee: '1',
-      gasPrice: '1',
+      gasPrice: '4494095',
```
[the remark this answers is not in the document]
shahafn: That's just the baseFee hardhat had, but it can be 1e9 like the rest. That's the first test I changed so...

package.json
forshtat: can you at least give it a descriptive name? thanks

packages/cli/src/GsnTestEnvironment.ts
forshtat: 😂🤦

packages/common/src/Utils.ts
forshtat: oh, ok then)) good that hardhat checks that

packages/dev/package.json
forshtat: please do, lets at least try to keep track of adding new yarn tasks))

packages/dev/test/HttpWrapper.test.ts
forshtat: I still think we should remove this test. it is useless

packages/dev/test/ServerTestUtils.ts
forshtat: ok ok

packages/dev/test/common/StatisticsManager.test.ts
forshtat: wait, every transaction increases block number on both hardhat and ganache, doesn't it?

packages/dev/test/provider/AccountManager.test.ts
forshtat: yeah, the 'getDefaultMethodSuffix' should work, right?

patches/truffle+5.1.26.patch
forshtat: can you please write this comment in the patch itself, maybe? it will be impossible to remember few months later

scripts/hardhat-node
forshtat: nice

forshtat:
https://github.com/trufflesuite/ganache/issues/939

forshtat (approved):

PR #690, opened 2021-09-30
Adding value to StoredTransaction

shahafn:
Also removing lastSuccessfulRounds

### DIFF
--- a/packages/relay/src/RelayServer.ts
+++ b/packages/relay/src/RelayServer.ts
@@ -56,7 +56,6 @@ export class RelayServer extends EventEmitter {
   lastScannedBlock: number
   lastRefreshBlock = 0
   ready = false
-  lastSuccessfulRounds = Number.MAX_SAFE_INTEGER
   readonly managerAddress: PrefixedHexString
   readonly workerAddress: PrefixedHexString
   minGasPrice: number = 0
@@ -387,7 +386,6 @@ returnValue        | ${viewRelayCallRet.returnValue}
         const timedOut = Date.now() - now
         this.logger.warn(chalk.bgRedBright(`Relay state: Timed-out after ${timedOut}`))
 
-        this.lastSuccessfulRounds = 0
       }, this.config.readyTimeout)
     }
 
@@ -407,7 +405,7 @@ returnValue        | ${viewRelayCallRet.returnValue}
         this.emit('error', e)
         const error = e as Error
         this.logger.error(`error in worker: ${error.message} ${error.stack}`)
-        this.lastSuccessfulRounds = 0
+        this.setReadyState(false)
       })
       .finally(() => {
         clearTimeout(workerTimeout)
@@ -437,7 +435,6 @@ returnValue        | ${viewRelayCallRet.returnValue}
 
     await this._worker(blockNumber)
       .then((transactions) => {
-        this.lastSuccessfulRounds++
         if (transactions.length !== 0) {
           this.logger.debug(`Done handling block #${blockNumber}. Created ${transactions.length} transactions.`)
         }
@@ -805,9 +802,6 @@ latestBlock timestamp   | ${latestBlock.timestamp}
   }
 
   isReady (): boolean {
-    if (this.lastSuccessfulRounds < this.config.successfulRoundsForReady) {
-      return false
-    }
     return this.ready
   }
 
@@ -815,12 +809,7 @@ latestBlock timestamp   | ${latestBlock.timestamp}
     if (this.isReady() !== isReady) {
       const now = Date.now()
       if (isReady) {
-        if (this.lastSuccessfulRounds < this.config.successfulRoundsForReady) {
-          const roundsUntilReady = this.config.successfulRoundsForReady - this.lastSuccessfulRounds
-          this.logger.warn(chalk.yellow(`Relayer state: almost READY (in ${roundsUntilReady} rounds)`))
-        } else {
-          this.logger.warn(chalk.greenBright('Relayer state: READY'))
-        }
+        this.logger.warn(chalk.greenBright('Relayer state: READY'))
         this.readinessInfo.totalNotReadyTime += now - this.readinessInfo.currentStateTimestamp
       } else {
         this.readinessInfo.totalReadyTime += now - this.readinessInfo.currentStateTimestamp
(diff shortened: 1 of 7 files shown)


.circleci/config.yml
```diff
@@ -9,17 +9,23 @@ jobs: # a collection of steps
       - checkout # special step to check out source code to working directory
 
       - restore_cache: # special step to restore the dependency cache
-          key: dependency-cache-{{ checksum "package.json" }}-yarn2
+          key: dependency-cache-{{ checksum "yarn.lock" }}
```
forshtat, line 12: This is dangerous as not always do we check in a modified "yarn.lock" file and in general I think it is ok not to keep it always checked in. It tends to get merge conflicts and stuff. Why did you change it?

package.json
```diff
@@ -41,13 +41,12 @@
...
   },
-  "prepublish": "yarn prepare",
   "scripts": {
-    "test": "yarn prepare && yarn lerna-lint && yarn lerna-run-test-with-hardhat",
+    "test": "yarn preprocess && yarn lerna-lint && yarn lerna-run-test-with-hardhat",
     "ganache": "ganache-cli --chainId 1337 --hardfork 'istanbul' --gasLimit 120000000 --defaultBalanceEther 1000 --deterministic --keepAliveTimeout 2147483647",
     "ganache-fork-mainnet": "ganache-cli --fork https://mainnet.infura.io/v3/f40be2b1a3914db682491dc62a19ad43 --port 8544 --chainId 1 --l 10000000  --unlock 0x47ac0fb4f2d84898e4d9e7b4dab3c24507a6d503",
     "postinstall": "patch-package",
-    "prepare": "yarn lerna-rm-dist && yarn lerna-typechain-generate && yarn extract_abi && yarn lerna-tsc && yarn lerna-chmod-commands",
+    "preprocess": "yarn lerna-rm-dist && yarn lerna-typechain-generate && yarn extract_abi && yarn lerna-tsc && yarn lerna-chmod-commands",
```
forshtat, line 49: Isn't `prepare` a keyword for yarn?

packages/dev/package.json
```diff
@@ -14,7 +14,7 @@
     "tsc": "tsc",
     "truffle-compile": "truffle compile",
     "lint": "eslint -f unix .",
-    "test": "lerna run tsc && truffle --compile-all --network npmtest test",
+    "test111": "lerna run tsc && truffle --compile-all --network npmtest test",
```
forshtat, line 17: Are you sure this is a name you want?

packages/paymasters/test/HashcashPaymaster.test.ts
```diff
@@ -15,7 +15,7 @@ const HashcashPaymaster = artifacts.require('HashcashPaymaster')
 const SampleRecipient = artifacts.require('SampleRecipient')
 const IRelayHub = artifacts.require('IRelayHub')
 
-contract('HashcashPaymaster', ([from]) => {
+contract.only('HashcashPaymaster', ([from]) => {
```
forshtat, line 18: Only?

packages/paymasters/test/PermitERC20UniswapV3Paymaster.test.ts
```diff
@@ -45,12 +45,12 @@ const IQuoter = artifacts.require('IQuoter')
 // as we are using forked mainnet, we will need to impersonate an account with a lot of DAI & UNI
 const MAJOR_DAI_AND_UNI_HOLDER = '0x47ac0fb4f2d84898e4d9e7b4dab3c24507a6d503'
 
-const GAS_USED_BY_POST = 199009
+const GAS_USED_BY_POST = 197490
```
forshtat, line 48: lets make it 200000 already, maybe?

packages/relay/src/RelayServer.ts
```diff
@@ -805,22 +801,14 @@ latestBlock timestamp   | ${latestBlock.timestamp}
...
     if (this.isReady() !== isReady) {
       const now = Date.now()
       if (isReady) {
-        if (this.lastSuccessfulRounds < this.config.successfulRoundsForReady) {
-          const roundsUntilReady = this.config.successfulRoundsForReady - this.lastSuccessfulRounds
-          this.logger.warn(chalk.yellow(`Relayer state: almost READY (in ${roundsUntilReady} rounds)`))
-        } else {
-          this.logger.warn(chalk.greenBright('Relayer state: READY'))
-        }
+        this.logger.warn(chalk.greenBright('Relayer state: READY'))
```
forshtat, line 811: Does it solve anything? Like, I get it that the relay turns into 'ready' faster but the idea behind 'successfulRoundsForReady' was that if there is something wrong with an RPC to let it resolve itself.

forshtat (changes requested):

.circleci/config.yml
shahafn: Two reasons:
1. checksum only the root package.json and not all the nested package.json is incorrect
2. We care only about installed packages when caching node_modules, we don't really care about other package.json changes that don't affect the installed packages ( added scripts etc.)

I took example from circleci https://circleci.com/docs/2.0/sample-config/ using yarn.lock

package.json
shahafn: That's exactly the issue, it's being run (multiple times) even when we don't want to run it, which makes each run several minutes longer

packages/dev/package.json
shahafn: This is a temporaray name, Stop reviewing this PR :rofl:

packages/relay/src/RelayServer.ts
drortirosh: indeed. instead of removing the roundsForRedy, maybe configure it to 0 (or 1) would get the same effect?

packages/common/src/ContractInteractor.ts
```diff
@@ -566,7 +566,11 @@ export class ContractInteractor {
     let blockNumber = -1
     let attempts = 0
     while (blockNumber < this.lastBlockNumber && attempts <= 100) {
-      blockNumber = await this.web3.eth.getBlockNumber()
+      try {
+        blockNumber = await this.web3.eth.getBlockNumber()
+      } catch (e) {
+        this.logger.error(`getBlockNumber: ${(e as Error).message}`)
+      }
       await sleep(100)
```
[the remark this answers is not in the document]
forshtat: My problem is not that it is 100 ms but that it is not conditional. So this method **will** sleep for 100 ms every time it is called, whether it's necessary or not, am I right?

forshtat (approved):

PR #691, opened 2021-10-09
Polygon Compatibility

Genysys:


shahafn:
Hi @Genysys, thank you  for this well needed PR!

- The EC2 deployment looks good! we think that its place is in the [docs](https://docs.opengsn.org/relay-server/deployment-reference.html#start-a-machine) and not as readme in the source code.
- The wrapper script looks organized and well written, though the main issue is that it assumes that your mnemonic (and therefore your funds) are available on the cloud instance, which is something we wanted to avoid. We're currently working on replacing our relaydc script and making gsn-relayer-register easier to use, so hopefully this  will help this script as well once we're done.

PR #694, opened 2021-10-13
Fixing getBlockNumber

shahafn:
Removing redundant sleep()

### DIFF
--- a/packages/common/src/ContractInteractor.ts
+++ b/packages/common/src/ContractInteractor.ts
@@ -82,7 +82,6 @@ export class ContractInteractor {
   private networkId?: number
   private networkType?: string
   private paymasterVersion?: SemVerString
-  private chain!: string
 
   constructor (
     {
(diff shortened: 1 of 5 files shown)


packages/common/src/ContractInteractor.ts
```diff
@@ -476,6 +477,7 @@ export class ContractInteractor {
    * In case 'getLogs' returned with a common error message of "more than X events" dynamically decrease page size.
    */
   async _getPastEventsPaginated (contract: any, names: EventName[], extraTopics: string[], options: PastEventOptions): Promise<EventData[]> {
+    const delay = this.chain === 'private' ? 0 : 300
```
forshtat, line 480: Can you explain the idea of adding a delay here?
Also, I do not like declaring a variable on the first line of a method only to use it on the last line. Can you move the declaration to where this is used?

packages/common/src/ContractInteractor.ts
```diff
@@ -565,9 +567,17 @@ export class ContractInteractor {
   async getBlockNumber (): Promise<number> {
     let blockNumber = -1
     let attempts = 0
-    while (blockNumber < this.lastBlockNumber && attempts <= 100) {
-      blockNumber = await this.web3.eth.getBlockNumber()
-      await sleep(100)
+    const delay = this.chain === 'private' ? 0 : 1000
```
forshtat, line 570: It also seems you never actually care about 'chain' aka 'network type' but use it to decide what delay to use. Maybe you can just have a `this.rpcRetryDelay` member or something like that? Would make it all more obvious.

packages/dev/test/ReputationFlow.test.ts
```diff
@@ -42,6 +42,10 @@ contract('ReputationFlow', function (accounts) {
     TestRecipient.web3.setProvider(relayProvider)
   })
 
+  after(async function () {
```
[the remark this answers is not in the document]
forshtat: Each test is supposed to call 'stopGsn()' after it's done, if it doesn't do it it is a bug.

forshtat (dismissed):

PR #697, opened 2021-10-18
Create ArbRelayHub that overrides RelayHub

forshtat:


packages/dev/test/arbitrum/ArbRelayHub.test.ts
```diff
@@ -0,0 +1,118 @@
...
+    })
+  })
+
+  context('#relayCall()', function () {
+    const transactionCalldataGasUsed = 7e6.toString()
+
+    let relayRequest: RelayRequest
+    let signature: string
+
+    before('so much boilerplate only to get the relay request working...', async function () {
```
shahafn, line 47: Come on :laughing:

packages/dev/test/arbitrum/ArbRelayHub.test.ts
```diff
@@ -0,0 +1,118 @@
...
+
+  before(async function () {
+    forwarder = await Forwarder.new()
+    stakeManager = await StakeManager.new(Number.MAX_SAFE_INTEGER)
+    const testArbSys = await TestArbSys.new()
+    arbRelayHub = await ArbRelayHub.new(testArbSys.address, stakeManager.address, constants.ZERO_ADDRESS, environments.arbitrum.relayHubConfiguration)
+  })
+
+  context('#aggregateGasleft()', function () {
+    it('should return gas left both for execution and for L2 storage', async function () {
```
shahafn, line 35: I don't understand why this test is needed

drortirosh: this test indeed does nothing.
just like you can't really test "gasleft()" itself this way
you need to test it across some code (before/after), where we have some assumption on how much gas it should cost.
it has to be tested on arb node over a test that actually make some storage access - like the one we ran on actual arbitrum testnet.

this test is useless on hardhat.

forshtat: yes, not running arbitrum locally means I cannot "really" test anything, so I just write the tests that do nothing hoping they add a way to run Arbitrum locally and we add it to our test suite

packages/contracts/src/arbitrum/ArbRelayHub.sol
```diff
@@ -0,0 +1,24 @@
...
+
+import "../RelayHub.sol";
+import "./ArbSys.sol";
+
+contract ArbRelayHub is RelayHub {
+    ArbSys public immutable arbsys;
+
+    // note: we accept the 'ArbSys' address in the constructor to allow mocking it in tests
+    constructor(
+        ArbSys _arbsys,
```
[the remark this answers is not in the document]
forshtat: // note: we accept the 'ArbSys' address in the constructor to allow mocking it in tests

packages/dev/test/arbitrum/ArbRelayHub.test.ts
```diff
@@ -0,0 +1,118 @@
...
+        forwarder.address,
+        relayRequest
+      )
+      signature = await getEip712Signature(
+        web3,
+        dataToSign
+      )
+    })
+
+    it('should use aggregateGasleft results when calculating charge', async function () {
```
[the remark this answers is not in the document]
forshtat: Well, our account would quickly run out of Arbitrum Testnet Ether if we were to run tests on an actual network. I am optimistic that they will re-enable local running of the node, as it is probably only a scripts issue, and then these tests would make more sense

packages/dev/test/arbitrum/ArbRelayHub.test.ts
drortirosh: ok. didn't look into you TestArbSys before. I guess it is the best as it can get for a test on hardhat...

PR #707, opened 2021-11-08
fix truffle dependencies

drortirosh:
use directly the contracts from the "../contracts/src", instead of the
import by "Migrations"


packages/relay/src/ReputationManager.ts
```diff
@@ -105,7 +105,7 @@ export class ReputationManager {
     if (-changeInAbuseWindow >= this.config.abuseReputationChange) {
       await this.reputationStoreManager.setAbuseFlag(paymaster, eventBlockNumber)
     }
-    const oldChangesExpirationBlock = eventBlockNumber - this.config.abuseTimeWindowBlocks
+    const oldChangesExpirationBlock = Math.max(eventBlockNumber - this.config.abuseTimeWindowBlocks, 0)
```
forshtat, line 108: How did this change end up in this PR? Why did you have to fix the semi-abandoned reputation manager?

packages/relay/src/TransactionManager.ts
```diff
@@ -244,7 +244,7 @@ data         | 0x${transaction.data.toString('hex')}
     // Get nonce at confirmationsNeeded blocks ago
     const nonces = new Map<string, number>()
     for (const transaction of sortedTxs) {
-      const nonce = nonces.get(transaction.from) ?? await this.contractInteractor.getTransactionCount(transaction.from, blockNumber - this.config.confirmationsNeeded)
+      const nonce = nonces.get(transaction.from) ?? await this.contractInteractor.getTransactionCount(transaction.from, Math.max(0, blockNumber - this.config.confirmationsNeeded))
```
forshtat, line 247: Same question here

forshtat (changes requested):

PR #709, opened 2021-11-17
OG-576 OG-629 Eip1559 support

shahafn:


packages/common/src/Utils.ts
```diff
@@ -225,14 +217,19 @@ export function boolString (bool: boolean): string {
...
+export function getDataAndSignature (tx: TypedTransaction, chainId: number): { data: string, signature: string } {
   if (tx.to == null) {
     throw new Error('tx.to must be defined')
   }
   if (tx.s == null || tx.r == null || tx.v == null) {
     throw new Error('tx signature must be defined')
   }
-  const input: List = [bnToUnpaddedBuffer(tx.nonce), bnToUnpaddedBuffer(tx.gasPrice), bnToUnpaddedBuffer(tx.gasLimit), tx.to.toBuffer(), bnToUnpaddedBuffer(tx.value), tx.data]
+  let input: List
+  if (!tx.supports(Capability.EIP1559FeeMarket)) {
```
forshtat, line 228: I guess the arrays are 80% identical, maybe lets do 
```
input.push(bla)
if (1559){
  input.push(bla)
}
else {
  input.push(bla)
}
```

alternatively, this is too hard to read, lets go all in with
```
[
  a,
  b,
  c
]
```

packages/common/src/types/RelayTransactionRequest.ts
```diff
@@ -12,39 +12,51 @@ export interface RelayMetadata {
...
+
+const metadataShape = {
+  approvalData: ow.string,
+  relayHubAddress: ow.string,
+  relayMaxNonce: ow.number,
+  signature: ow.string,
+  maxAcceptanceBudget: ow.string
+}
+
+const baseRelayDataShape = {
```
forshtat, line 41: Do you still need this separation to base- and an override?

packages/contracts/src/RelayHub.sol
```diff
@@ -374,7 +374,13 @@ contract RelayHub is IRelayHub, Ownable {
     }
 
     function calculateCharge(uint256 gasUsed, GsnTypes.RelayData calldata relayData) public override virtual view returns (uint256) {
-        return relayData.baseRelayFee.add((gasUsed.mul(relayData.gasPrice).mul(relayData.pctRelayFee.add(100))).div(100));
+        uint256 chargeableGasPrice;
+        if (relayData.maxFeePerGas > tx.gasprice) {
```
forshtat, line 378: Is this a 'min()'?

packages/common/src/types/RelayTransactionRequest.ts
shahafn: Not anymore

packages/contracts/src/RelayHub.sol
shahafn: Indeed

packages/common/src/ContractInteractor.ts
```diff
@@ -377,7 +382,7 @@ export class ContractInteractor {
           if (revertMsg != null) {
             reject(new Error(revertMsg))
           }
-          if (err !== null) {
+          if (err as boolean) {
```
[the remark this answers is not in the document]
shahafn: That's what happens when WS provider returns false instead of null/undefined...

packages/common/src/ContractInteractor.ts
```diff
@@ -885,9 +938,9 @@ calculateTransactionMaxPossibleGas: result: ${result}
         ],
         id: Date.now()
       }, (e: Error | null, r: any) => {
-        if (e != null) {
+        if (e as any as boolean) {
```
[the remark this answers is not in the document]
shahafn: See above

packages/relay/src/TransactionManager.ts
```diff
@@ -117,50 +121,76 @@ data         | 0x${transaction.data.toString('hex')}
...
+          gasLimit: txDetails.gasLimit,
+          gasPrice: maxFeePerGas,
+          data: Buffer.from(encodedCall.slice(2), 'hex'),
+          nonce
+        }, this.rawTxOptions)
+      }
       // TODO omg! do not do this!
-      const keyManager = this.managerKeyManager.isSigner(signer) ? this.managerKeyManager : this.workersKeyManager
-      signedTransaction = keyManager.signTransaction(signer, txToSign)
+      const keyManager = this.managerKeyManager.isSigner(txDetails.signer) ? this.managerKeyManager : this.workersKeyManager
```
drortirosh, line 168: very strange API: instead of passing signer as parameter to the method, you put it as a member in the txDetails...

packages/relay/src/TransactionManager.ts
```diff
@@ -117,50 +121,76 @@ data         | 0x${transaction.data.toString('hex')}
...
+      const transactionHash = await this.contractInteractor.broadcastTransaction(signedTransaction.rawTx)
+      if (transactionHash.toLowerCase() !== storedTx.txId.toLowerCase()) {
+        throw new Error(`txhash mismatch: from receipt: ${transactionHash} from txstore:${storedTx.txId}`)
+      }
+      return {
+        transactionHash,
+        signedTx: signedTransaction.rawTx
+      }
+    } catch (e) {
+      throw new Error(`Tx broadcast failed: ${(e as Error).message}`)
```
drortirosh, line 193: thisis an old problem: if we fail to broadcast the transaction - what is next? how we "gracefully " continue ? 
at a minimum, we should flag it as "critical" error, as its not something the server can assume to resolve itself
(there are some known errors it might. currently we ignore it - and sometimes it leaves the server in limbo.)

packages/relay/src/TransactionManager.ts
```diff
@@ -183,24 +213,38 @@ data         | 0x${transaction.data.toString('hex')}
...
+          nonce: tx.nonce,
+          value: tx.value
+        },
+        this.rawTxOptions)
+    } else {
+      txToSign = new Transaction(
+        {
+          to: tx.to,
+          gasLimit: tx.gas,
+          gasPrice: newMaxFee,
```
drortirosh, line 236: please note: there are places where you resolve to legacy with the priority, and others using the fee. 
its better to stick to one (even though they should be the same!) - I suggest to stick to fee, like here.

packages/relay/src/TransactionManager.ts
```diff
@@ -213,16 +257,17 @@ data         | 0x${transaction.data.toString('hex')}
     }
   }
 
-  _resolveNewGasPrice (oldGasPrice: number): { newGasPrice: number, isMaxGasPriceReached: boolean } {
+  _resolveNewGasPrice (oldMaxFee: number, oldMaxPriorityFee: number): { newMaxFee: number, newMaxPriorityFee: number, isMaxGasPriceReached: boolean } {
     let isMaxGasPriceReached = false
-    let newGasPrice = oldGasPrice * this.config.retryGasPriceFactor
+    let newMaxFee = oldMaxFee * this.config.retryGasPriceFactor
```
drortirosh, line 262: max fee should not factor - it should stay using the same configured max, and will cause to wait until the fee goes below it.
only priority should be factored.

packages/contracts/src/RelayHub.sol
shahafn: Already fixed... wasn't pushed yet

packages/provider/src/RelayClient.ts
```diff
@@ -47,10 +47,10 @@ export const EmptyDataCallback: AsyncDataCallback = async (): Promise<PrefixedHe
 
 export const GasPricePingFilter: PingFilter = (pingResponse, gsnTransactionDetails) => {
   if (
-    gsnTransactionDetails.gasPrice != null &&
-    parseInt(pingResponse.minGasPrice) > parseInt(gsnTransactionDetails.gasPrice)
+    gsnTransactionDetails.maxPriorityFeePerGas != null &&
+    parseInt(pingResponse.minPriorityFeePerGas) > parseInt(gsnTransactionDetails.maxPriorityFeePerGas)
```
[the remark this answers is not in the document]
shahafn: Why? equality is fine here

packages/relay/src/TransactionManager.ts
```diff
@@ -117,50 +121,76 @@ data         | 0x${transaction.data.toString('hex')}
...
 
-  async sendTransaction ({ signer, method, destination, value = '0x', gasLimit, gasPrice, creationBlockNumber, serverAction }: SendTransactionDetails): Promise<SignedTransactionDetails> {
-    const encodedCall = method?.encodeABI() ?? '0x'
-    const _gasPrice = parseInt(gasPrice ?? await this.gasPriceFetcher.getGasPrice())
-    const releaseMutex = await this.nonceMutex.acquire()
-    if (isSameAddress(destination, constants.ZERO_ADDRESS)) {
-      const msg = `Preventing to send transaction with action id ${serverAction} to address(0)! Validate your configuration!`
+  async sendTransaction (txDetails: SendTransactionDetails): Promise<SignedTransactionDetails> {
+    const encodedCall = txDetails.method?.encodeABI() ?? '0x'
+    const maxFeePerGas = parseInt(txDetails.maxFeePerGas ?? await this.gasPriceFetcher.getGasPrice())
```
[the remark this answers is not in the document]
shahafn: Alchemy has a `getMaxPriorityFee()` like the one I added to contractInteractor. I don't want to add yet another "percentile" config parameter before we know we need it..

packages/relay/src/TransactionManager.ts
shahafn: What? I didn't change API here

packages/relay/src/TransactionManager.ts
shahafn: I agree

packages/contracts/src/RelayHub.sol
```diff
@@ -381,7 +382,14 @@ contract RelayHub is IRelayHub, Ownable {
...
 
     function calculateCharge(uint256 gasUsed, GsnTypes.RelayData calldata relayData) public override virtual view returns (uint256) {
-        return relayData.baseRelayFee.add((gasUsed.mul(relayData.gasPrice).mul(relayData.pctRelayFee.add(100))).div(100));
+        uint256 basefee;
+        if (relayData.maxFeePerGas == relayData.maxPriorityFeePerGas) {
+            basefee = 0;
+        } else {
+            basefee = block.basefee;
+        }
+        uint256 chargeableGasPrice = Math.min(relayData.maxFeePerGas, Math.min(tx.gasprice, basefee + relayData.maxPriorityFeePerGas));
```
forshtat, line 391: math.min accepts any number of parameters and returns the lowest one...

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/min

shahafn: Alas, this is Solidity and not Nodejs, so Math.min accepts only two parameters here...

packages/dev/test/Flows.test.ts
```diff
@@ -284,6 +315,18 @@ options.forEach(params => {
       })
     }
 
+    async function fixTxDetails (txDetails: any, relayProvider: RelayProvider): Promise<void> {
+      if (params.relay) {
+        const { maxFeePerGas, maxPriorityFeePerGas } = await relayProvider.calculateGasFees()
```
drortirosh, line 320: it makes more sense that calculateGasFees will return { mxaFeePerGas?, maxPrio?, gasPrice? }
then this "fix" will simply do:
`{ ... await relayProvider.calculateGasFees() }`

(it will help users, though, but will complicate the test, it will require an extra param "forceLegacy", which means to return gasPrice even in type-2 mode)
it is better for users in that it will not require this strange and confusing `"gasPrice = maxPriorityFeePerGas"`

packages/relay/src/TransactionManager.ts
```diff
@@ -68,41 +66,47 @@ export class TransactionManager {
...
-    const gasPriceHumanReadableOld: string = new EthVal(gasPrice).toGwei().toFixed(4)
+  printBoostedTransactionLog (txHash: string, creationBlockNumber: number, maxFeePerGas: number, maxPriorityFeePerGas: number, isMaxGasPriceReached: boolean): void {
+    const maxFeePerGasHumanReadableOld: string = new EthVal(maxFeePerGas).toGwei().toFixed(4)
+    const maxPriorityFeePerGasHumanReadableOld: string = new EthVal(maxPriorityFeePerGas).toGwei().toFixed(4)
     this.logger.info(`Boosting stale transaction:
-hash         | ${txHash}
-gasPrice     | ${gasPrice} (${gasPriceHumanReadableOld} gwei) ${isMaxGasPriceReached ? chalk.red('(MAX GAS PRICE REACHED)') : ''}
-created at   | block #${creationBlockNumber}
+hash                     | ${txHash}
+maxFeePerGas             | ${maxFeePerGas} (${maxFeePerGasHumanReadableOld} gwei) ${isMaxGasPriceReached ? chalk.red('(MAX GAS PRICE REACHED)') : ''}
```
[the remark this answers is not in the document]
shahafn: What do you mean by "how"? It displays maxFeePerGas which is equal to gasPrice on legacy networks in our case. I can add a check whether we're on legacy, then print "gasPrice" instead, but since we do want to remove legacy support eventually, I didn't think it was necessary.

packages/relay/src/RelayServer.ts
```diff
@@ -621,14 +667,21 @@ latestBlock timestamp   | ${latestBlock.timestamp}
...
+    //   const baseFee = (await this.contractInteractor.getBlock('latest')).baseFeePerGas.toString()
+    //   minPriorityFeePerGas -= parseInt(baseFee)
+    // }
+    this.minPriorityFeePerGas = Math.floor(minPriorityFeePerGas * this.config.gasPriceFactor)
+    if (this.minPriorityFeePerGas === 0) {
+      this.logger.debug(`Priority fee received from node is 0. Setting priority fee to ${this.config.defaultPriorityFee}`)
+      this.minPriorityFeePerGas = parseInt(this.config.defaultPriorityFee)
+    }
+    if (this.minPriorityFeePerGas > parseInt(this.config.maxGasPrice)) {
+      throw new Error(`network gas price ${this.minPriorityFeePerGas} is higher than max gas price ${this.config.maxGasPrice}`)
```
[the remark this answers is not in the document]
shahafn: That string was there before, fixing it now.

packages/dev/test/Flows.test.ts
```diff
@@ -18,27 +18,33 @@ import { GSNConfig } from '@opengsn/provider/dist/GSNConfigurator'
...
+  },
+  {
+    title: 'Type 2 Relayed-',
+    relay: true,
+    type: 2
   }
 ]
 
 options.forEach(params => {
+  const TestRecipient = artifacts.require('TestRecipient')
```
drortirosh, line 41: any special reason to move those "require" into the "foreach" loop? 
we usually keep them at the top of the page.

packages/relay/src/RelayServer.ts
```diff
@@ -621,14 +667,21 @@ latestBlock timestamp   | ${latestBlock.timestamp}
...
-    this.minGasPrice = Math.floor(parseInt(gasPriceString) * this.config.gasPriceFactor)
-    if (this.minGasPrice === 0) {
-      throw new Error('Could not get gasPrice from node')
-    }
-    if (this.minGasPrice > parseInt(this.config.maxGasPrice)) {
-      throw new Error(`network gas price ${this.minGasPrice} is higher than max gas price ${this.config.maxGasPrice}`)
+  async _refreshPriorityFee (): Promise<void> {
+    const minPriorityFeePerGas = parseInt(await this.contractInteractor.getMaxPriorityFee())
+    // if (this.transactionType == TransactionType.TYPE_TWO) {
+    //   // todo remove once updating to web3 1.6.1
```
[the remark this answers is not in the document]
shahafn: now done

packages/relay/src/RelayServer.ts
```diff
@@ -621,14 +667,21 @@ latestBlock timestamp   | ${latestBlock.timestamp}
...
-  async _refreshGasPrice (): Promise<void> {
-    const gasPriceString = await this.gasPriceFetcher.getGasPrice()
-    this.minGasPrice = Math.floor(parseInt(gasPriceString) * this.config.gasPriceFactor)
-    if (this.minGasPrice === 0) {
-      throw new Error('Could not get gasPrice from node')
-    }
-    if (this.minGasPrice > parseInt(this.config.maxGasPrice)) {
-      throw new Error(`network gas price ${this.minGasPrice} is higher than max gas price ${this.config.maxGasPrice}`)
+  async _refreshPriorityFee (): Promise<void> {
+    const minPriorityFeePerGas = parseInt(await this.contractInteractor.getMaxPriorityFee())
```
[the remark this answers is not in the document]
shahafn: Why?

packages/common/src/ContractInteractor.ts
```diff
@@ -222,11 +223,13 @@ export class ContractInteractor {
       relayHubAddress, forwarderAddress, paymasterVersion
     ] = await Promise.all([
       this.paymasterInstance.getHubAddr().catch((e: Error) => { throw new Error(`Not a paymaster contract: ${e.message}`) }),
-      this.paymasterInstance.trustedForwarder().catch((e: Error) => { throw new Error(`paymaster has no trustedForwarder(): ${e.message}`) }),
-      this.paymasterInstance.versionPaymaster().catch((e: Error) => { throw new Error(`Not a paymaster contract: ${e.message}`) }).then((version: string) => {
-        this._validateVersion(version, 'Paymaster')
-        return version
-      })
+      this.paymasterInstance.trustedForwarder().catch(
```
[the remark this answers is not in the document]
shahafn: Added check

packages/common/src/ContractInteractor.ts
```diff
@@ -736,11 +753,47 @@ calculateTransactionMaxPossibleGas: result: ${result}
...
+          if (e as any as boolean) {
+            reject(e)
+          } else {
+            resolve(r.result)
+          }
+        })
+    })
+  }
+
+  async getMaxPriorityFee (): Promise<string> {
```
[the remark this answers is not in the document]
shahafn: I took example from Alchemy web3 which adds this rpc call, it's just a helper function when baseFee isn't needed (relayServer)

packages/common/src/Utils.ts
shahafn: fixed

packages/common/src/ContractInteractor.ts
```diff
@@ -736,11 +753,47 @@ calculateTransactionMaxPossibleGas: result: ${result}
       return gasPriceFromNode
     }
     const gasPriceActual =
-      new BN(gasPriceFromNode)
+      toBN(gasPriceFromNode)
         .muln(this.environment.getGasPriceFactor)
     return gasPriceActual.toString()
   }
 
+  async getFeeHistory (blockCount: number | BigNumber | BN | string, lastBlock: number | BigNumber | BN | string, rewardPercentiles: number[]): Promise<FeeHistoryResult> {
```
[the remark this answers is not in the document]
shahafn: I copied the signature from web3 since it's merely a wrapper now, and I added a helper function that does have 'latest' hardcoded.

packages/dev/test/Flows.test.ts
shahafn: Resolved over the phone

packages/dev/test/provider/RelayClient.test.ts
```diff
@@ -417,18 +407,18 @@ contract('RelayClient', function (accounts) {
...
-        minGasPrice
+        minPriorityFeePerGas
       }
       const relayClient = new RelayClient({ provider: underlyingProvider, config: gsnConfig })
       await relayClient.init()
 
-      const calculatedGasPrice = await relayClient._calculateGasPrice()
-      assert.equal(calculatedGasPrice, `0x${minGasPrice.toString(16)}`)
+      const calculatedGasPrice = await relayClient._calculateGasFees()
+      assert.equal(calculatedGasPrice.maxPriorityFeePerGas, `0x${minPriorityFeePerGas.toString(16)}`)
```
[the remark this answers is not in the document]
shahafn: I tried to be consistent with the underlying api types, and I don't think it has to be changed necessarily.

packages/relay/src/RelayServer.ts
```diff
@@ -194,6 +197,23 @@ export class RelayServer extends EventEmitter {
     }
   }
 
+  validateGasFees (req: RelayTransactionRequest): void {
```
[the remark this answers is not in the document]
shahafn: Refactored to ValidateRelayFees

packages/contracts/src/RelayHub.sol
forshtat: Oh, ok... Got confused by identical look of `Math.min`

packages/relay/src/RelayServer.ts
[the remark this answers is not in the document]
forshtat: How is it that `minPriorityFeePerGas = getMaxPriorityFee`?

Is it a `minMax` type of situation? I strongly suggest we find a way to name these things first :-)

packages/common/src/ContractInteractor.ts
```diff
@@ -363,7 +367,8 @@ export class ContractInteractor {
             {
               from: relayCallABIData.relayRequest.relayData.relayWorker,
               to: relayHub.address,
-              gasPrice: toHex(relayCallABIData.relayRequest.relayData.gasPrice),
+              // TODO fix gasPrice
```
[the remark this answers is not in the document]
shahafn: Fixed

packages/common/src/ContractInteractor.ts
```diff
@@ -639,14 +647,23 @@ export class ContractInteractor {
   }
 
   async estimateGas (gsnTransactionDetails: GsnTransactionDetails): Promise<number> {
-    return await this.web3.eth.estimateGas(gsnTransactionDetails)
+    return await this.web3.eth.estimateGas({
```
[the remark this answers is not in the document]
shahafn: fixed

packages/relay/src/RelayServer.ts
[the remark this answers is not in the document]
shahafn: Yes, minPriorityFeePerGas is actually the minimum maxPriorityFeePerGas, which is annoying a bit but workable.

packages/dev/test/Flows.test.ts
shahafn: There was a reason before, now no

packages/common/src/ContractInteractor.ts
```diff
@@ -363,6 +377,17 @@ export class ContractInteractor {
     return latestBlock.gasLimit
   }
 
+  _fixGasFees (relayRequest: RelayRequest): { gasPrice?: string, maxFeePerGas?: string, maxPriorityFeePerGas?: string } {
```
forshtat, line 380: There is `RelayProvider::_fixGasFees` and `ContractInteractor::_fixGasFees` that do similar but not exactly same things. I suggest having two different names :-)

packages/common/src/ContractInteractor.ts
```diff
@@ -440,7 +466,8 @@ export class ContractInteractor {
 
   async getMaxViewableGasLimit (relayRequest: RelayRequest, maxViewableGasLimit: IntString): Promise<BN> {
     const maxViewableGasLimitBN = toBN(maxViewableGasLimit)
-    const gasPrice = toBN(relayRequest.relayData.gasPrice)
+    // TODO fix gasPrice
```
forshtat, line 469: Not sure there is anything to do here - maxFeePerGas serves its role for the max gas limit, doesn't it?

packages/common/src/ContractInteractor.ts
```diff
@@ -773,11 +803,47 @@ calculateTransactionMaxPossibleGas: result: ${result}
...
+      toBN(gasPriceFromNode)
         .muln(this.environment.getGasPriceFactor)
     return gasPriceActual.toString()
   }
 
+  async getFeeHistory (blockCount: number | BigNumber | BN | string, lastBlock: number | BigNumber | BN | string, rewardPercentiles: number[]): Promise<FeeHistoryResult> {
+    // web3 incorrectly converts blockCount type from prefixed hex string to number, hardhat doesn't handle that properly
+    // see https://github.com/ChainSafe/web3.js/pull/4529
+    // once it's fixed and web3 version is updated use:
+    // return await this.web3.eth.getFeeHistory(blockCount, lastBlock, rewardPercentiles)
```
forshtat, line 815: Note: that PR was merged and 1.6.1 was released since, can we update?

packages/common/src/Utils.ts
forshtat: > 

"We need to see" as in "see with eyes looking at code"? Not sure I agree we need to. I think we should rather rely on a runtime check. 

Especially given THERE IS a push of 3 elements just below this enormous assignment, so the point is lost...

packages/common/src/ContractInteractor.ts
shahafn: Yes

packages/common/src/ContractInteractor.ts
shahafn: Yes, they are similar in nature but different in impl, but they are both "internal" functions, so I guess having the same name is ok, no?

packages/common/src/ContractInteractor.ts
shahafn: It's merged to their 1.x branch, but wasn't part from 1.6.1 release. They will add it in the future probably.

packages/common/src/Utils.ts
shahafn: fixed... I think

packages/provider/src/RelayClient.ts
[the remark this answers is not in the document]
drortirosh: ok., the "good" value is "<=", and you test here for error case, so its OK.

packages/relay/src/TransactionManager.ts
```diff
@@ -117,59 +121,93 @@ data         | 0x${transaction.data.toString('hex')}
...
+    if (txCost.gt(toBN(signerBalance))) {
+      throw new Error(`signer ${signer} balance ${signerBalance} too low: tx cost is ${txCost.toString()}`)
+    }
+  }
+
+  async tryBroadcastTransaction (signedTx: string, verifiedTxId: string): Promise<SignedTransactionDetails> {
+    try {
+      const transactionHash = await this.contractInteractor.broadcastTransaction(signedTx)
+      if (transactionHash.toLowerCase() !== verifiedTxId.toLowerCase()) {
+        throw new Error(`txhash mismatch: from receipt: ${transactionHash} from txstore:${verifiedTxId}`)
```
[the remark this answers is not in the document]
forshtat: On an unrelated note, having a `throw` inside your own `try` is a little bit of a `goto`, if you ask me :-)

packages/relay/src/TransactionManager.ts
```diff
@@ -117,59 +121,93 @@ data         | 0x${transaction.data.toString('hex')}
...
-        gasLimit,
-        gasPrice: _gasPrice,
-        data: Buffer.from(encodedCall.slice(2), 'hex'),
-        nonce
-      }, this.rawTxOptions)
+      const nonce = await this.pollNonce(txDetails.signer)
+
+      let txToSign: TypedTransaction
+      if (this.transactionType === TransactionType.TYPE_TWO) {
+        // todo add better maxFeePerGas, maxPriorityFeePerGas defaults
```
forshtat, line 166: this todo appears to be old, isn't it?

packages/relay/src/StoredTransaction.ts
```diff
@@ -25,7 +25,8 @@ export interface StoredTransactionMetadata {
 export interface StoredTransactionSerialized {
   readonly to: Address
   readonly gas: number
-  readonly gasPrice: number
+  maxFeePerGas: number
+  maxPriorityFeePerGas: number
```
forshtat, line 29: you can actually easily keep these as `readonly` with 2 variables (for consistency and extra protection)
you can also avoid `details as StoredTransactionSerialized` this way, I think.

forshtat (approved):

packages/relay/src/TransactionManager.ts
shahafn: Yes

drortirosh (approved):

PR #712, opened 2021-11-24
clean hardhat-console solidity logs from build

drortirosh:
if TEST_CONSOLE is set, then leave contracts as-is
if not, then compile with console.log and console.sol lines removed from
code.
NOTE: it means that such logs must NOT have any side-effects!
(Also, they might modify gas calculations, so at least the test gas
should be validated with TEST_CONSOLE cleared.

the "build/src" is always a fresh copy and carries original timestamps,
so incremental build can work.

### DIFF
--- a/packages/contracts/testCopyContracts.js
+++ b/packages/contracts/testCopyContracts.js
@@ -1,0 +1,39 @@
+import fs from 'fs'
+import path from 'path'
+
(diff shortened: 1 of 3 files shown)


drortirosh:
not a good idea: logs have (gas) side-effects, and our code deeply involved with gas. 
will break gas-related tests.

PR #713, opened 2021-11-24
make BLSBatchGateway test run (and with batches)

drortirosh:


packages/common/src/LightTruffleContract.ts
```diff
@@ -12,6 +12,14 @@ function getComponent (key: string, components: AbiOutput[]): AbiOutput | undefi
 }
 
 function retypeItem (abiOutput: AbiOutput, ret: any): any {
+  // array of some type. parse each member separately
+  if (abiOutput.type.endsWith('[]')) {
```
forshtat, line 16: a reminder to fix the 'else' cases here with '[' handling code

packages/common/src/bls/CacheDecoderInteractor.ts
```diff
@@ -144,13 +144,13 @@ export class CacheDecoderInteractor {
    */
   async compressRelayRequest (_: { relayRequest: RelayRequest, cachedEncodedData?: PrefixedHexString }): Promise<RelayRequestCachingResult> {
     const nonce = toBN(_.relayRequest.request.nonce)
-    const paymaster = toBN(_.relayRequest.relayData.paymaster)
 
     // TODO: return the tuple of 'address, type' so that all addresses can be queried in one RPC request!
-    const addressesCompressed = await this.compressAddressesToIds([
+    const resolved = await this.compressAddressesToIds(
```
forshtat, line 149: Are you sure this is better as 4 parameters than with an array of arrays? 
Maybe we should do ```function foo ([a, b, c, d]: string[4]): void {}```?

packages/common/src/bls/CacheDecoderInteractor.ts
```diff
@@ -163,21 +163,20 @@ export class CacheDecoderInteractor {
     } else {
       methodData = toBuffer(_.cachedEncodedData)
     }
-    // 0 means use same decoder as the entire batch
-    const cacheDecoder = toBN(0)
+    const cacheDecoder = toBN(1)  //TODO: for un-compressed calldata
```
forshtat, line 166: Please leave `0` as most tests I do rely on calldata decoder already

packages/common/src/bls/ICalldataCacheDecoderInteractor.ts
```diff
@@ -2,7 +2,10 @@ import { PrefixedHexString } from 'ethereumjs-util'
 import BN from 'bn.js'
 
 export interface AddressesCachingResult {
-  ids: BN[][]
+  senderAsIds: BN[],
```
forshtat, line 5: It should be separated to Gateway Address Caching Result and ERC20 Address Caching Result as it is a different object for different caching purposes

packages/common/src/bls/ERC20CalldataCacheDecoderInteractor.ts
```diff
@@ -97,11 +97,17 @@ export class ERC20CalldataCacheDecoderInteractor implements ICalldataCacheDecode
   }
 
   async compressAddressesToIds (addresses: Address[][]): Promise<AddressesCachingResult> {
-    return { ids: addresses.map(it => it.map(toBN)), writeSlotsCount: 0 }
+    return {
```
forshtat, line 100: again here, for ERC20 the addresses are just the holders of the token, a single array

packages/dev/hardhat.config.ts
```diff
@@ -0,0 +1,29 @@
+/**
```
forshtat, line 1: I don't like how many commented-out lines are here, can we remove these comments?

packages/dev/migrations/1_initial_migration.js
```diff
@@ -1,5 +0,0 @@
-var Migrations = artifacts.require('./Migrations.sol')
```
forshtat, line 1: please do not remove migrations in this PR

packages/dev/test/bls/BLSBatchGateway.test.ts
```diff
@@ -31,6 +31,62 @@ const BLSBatchGateway = artifacts.require('BLSBatchGateway')
 const BLSTestHub = artifacts.require('BLSTestHub')
 const TestToken = artifacts.require('TestToken')
 
+interface DebugTransaction {
```
forshtat, line 34: I don't think we should have this code in the BLS tests, maybe in TestUtils?

packages/dev/test/bls/BLSBatchGateway.test.ts
```diff
@@ -44,11 +100,11 @@ async function createAuthorizationSignature (
   return await accountManager.createAccountAuthorization(from, registrar.address.toLowerCase())
 }
 
-contract.only('BLSBatchGateway', function ([from, from2]: string[]) {
+contract.only('BLSBatchGateway', function ([from, to, from2]: string[]) {
```
forshtat, line 103: 'to' was removed on purpose - need to use 'testToken.address'

packages/dev/test/bls/BLSBatchGateway.test.ts
```diff
@@ -185,6 +241,82 @@ contract.only('BLSBatchGateway', function ([from, from2]: string[]) {
       })
     })
 
+    ;[
```
forshtat, line 244: 😞 please put ; on the previous line

why does it only run with `2` now?

packages/dev/test/bls/BLSBatchGateway.test.ts
```diff
@@ -185,6 +241,82 @@ contract.only('BLSBatchGateway', function ([from, from2]: string[]) {
...
+        //
+        // await expectEvent.inTransaction(receipt.transactionHash, BLSBatchGateway, 'BatchRelayed', {
+        //   relayWorker: from,
+        //   batchSize: '1'
+        // })
+        //
+        // await expectEvent.inTransaction(receipt.transactionHash, BLSTestHub, 'ReceivedRelayCall', {
+        //   requestFrom: from,
+        //   requestTo: to
+        // })
```
forshtat, line 317: These were the asserts, this test does not have any asserts now

forshtat (approved):

PR #714, opened 2021-11-29
release 2.2.5

drortirosh:
fixed relayer boosting bug (after validating on real-life relayer)


forshtat:
@drortirosh what is this PR? it does not include a fix, only a version string. Should it be merged?

PR #715, opened 2021-12-05
fix batch provider

drortirosh:
support batch of multiple requests
(db sync issue if submitted multiple TXs too fast)


packages/dev/test/bls/BatchFlow.test.ts
```diff
@@ -77,21 +81,41 @@ contract.only('Batch Relaying Flow', function (accounts: string[]) {
...
   it('should relay batch', async function () {
-    const transactionResponse = await testToken.transfer(accounts[3], 200000, {
+
+    const tx1 = testToken.transfer(accounts[3], 200000, {
       from: accounts[0],
       gas: innerTransactionGas
     })
 
+    //TODO: removing this sometime causes:
+    // Can't insert key 0xb354ecf032e9e14442be590d9eaee37d2924b67a, it violates the unique constraint"
```
forshtat, line 100: This seems to indicate a bug in our storage code, it will be interesting to investigate.

PR #716, opened 2021-12-06
OG-634 Owner withdraw

shahafn:


packages/dev/test/RelayServer.test.ts
```diff
@@ -605,6 +606,212 @@ contract('RelayServer', function (accounts: Truffle.Accounts) {
...
+    })
+
+    describe('_resolveWithdrawalConfig', function () {
+      async function assertReturnedError (message: string): Promise<void> {
+        const spy = sinon.spy(env.relayServer.logger, 'error')
+        await env.relayServer.withdrawToOwnerIfNeeded(currentBlockNumber)
+        sinon.assert.calledWith(spy, `withdrawToOwnerIfNeeded: ${message}`)
+        spy.restore()
+      }
+      it('should throw with bad config file', async function () {
```
drortirosh, line 651: what is the net result ? the server stops on bad config? (maybe its not such a bad idea, since you don't have a way to signal the user the file is bad)

packages/dev/test/RelayServer.test.ts
```diff
@@ -605,6 +606,212 @@ contract('RelayServer', function (accounts: Truffle.Accounts) {
...
+        try {
+          await env.relayServer._resolveWithdrawalConfig(configFilename)
+          assert.fail()
+        } catch (e) {
+          assert.include(e.message, `leaveManagerWithAmountEth must be at least managerTargetBalance ${env.relayServer.config.managerTargetBalance}`)
+          await assertReturnedError(e.message)
+        }
+      })
+
+      it('should throw if withdrawOnEthBalanceReached - leaveManagerWithAmountEth < 1 ether  and config.smallAmount is not configured', async function () {
```
drortirosh, line 700: too much configuration.
maybe have `minimumWithdrawEth` - which is the (minimum) value which will be sent to the owner. 
it will be sent only after the manager withdraw to itself, and leave its own balance high enough. so there is no need for the extra "small amount" flag: if I specify `minimumWithdrawEth = 0.1` then its MY problem...

packages/relay/src/RelayServer.ts
```diff
@@ -428,70 +435,19 @@ returnValue        | ${viewRelayCallRet.returnValue}
...
-        this.logger.error(`error in worker: ${error.message} gotBlockNumber: ${gotBlock} ${error.stack}`)
-        this.setReadyState(false)
-      })
-      .finally(() => {
-        clearTimeout(workerTimeout)
-      })
-  }
-
   start (): void {
     this.logger.debug(`Started polling for new blocks every ${this.config.checkInterval}ms`)
```
drortirosh, line 439: new logging are better (info for one-time items, debug for repeated, non-critical).
this `debug` should match the `stop` conditions, and be `info`

packages/relay/src/RelayServer.ts
```diff
@@ -823,6 +809,57 @@ latestBlock timestamp   | ${latestBlock.timestamp}
...
+        return txHashes
+      }
+      const { withdrawalAmount, repeat } = await this._resolveWithdrawalConfig(filename)
+      if (withdrawalAmount.eqn(0)) {
+        return txHashes
+      }
+      txHashes = txHashes.concat(await this.registrationManager._sendManagerHubBalanceToOwner(blockNumber, withdrawalAmount))
+      if (!repeat) {
+        this.logger.info('Removing withdraw file.')
+        fs.rmSync(filename)
```
drortirosh, line 852: maybe instead of removing the file, rename it to filename."done" ? this way it can be repeated by moving it back (or, if it failed - check the content of the file to understand why)
if that ".done" file already exist, it should be removed first. no need for more than one backup..

packages/dev/test/RelayServer.test.ts
```diff
@@ -605,6 +606,212 @@ contract('RelayServer', function (accounts: Truffle.Accounts) {
...
+    beforeEach(async function () {
+      await env.newServerInstance({}, getTemporaryWorkdirs())
+      configFilename = `${env.relayServer.config.workdir}/withdraw.json`
+      assert.isFalse(fs.existsSync(configFilename))
+      currentBlockNumber = await env.web3.eth.getBlockNumber()
+    })
+    afterEach(async function () {
+      sinon.restore()
+    })
+    it('should not withdraw if relayer is not ready', async function () {
```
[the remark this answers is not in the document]
shahafn: The problem that I see with it is that we don't really know why it's not ready (which is a known design issue with the server...) so I cannot tell whether it's not ready because it's not connected to the node, the DB is corrupted etc. So I don't know that in all cases where the server isn't ready, it's still ok to try to withdraw. What do you think?

packages/dev/test/RelayServer.test.ts
```diff
@@ -605,6 +606,212 @@ contract('RelayServer', function (accounts: Truffle.Accounts) {
...
+      assert.deepEqual(txHashes, [])
+      sinon.assert.calledOnce(serverSpy.isReady)
+      assert.isTrue(serverSpy.isReady.returnValues[0])
+    })
+
+    describe('_resolveWithdrawalConfig', function () {
+      async function assertReturnedError (message: string): Promise<void> {
+        const spy = sinon.spy(env.relayServer.logger, 'error')
+        await env.relayServer.withdrawToOwnerIfNeeded(currentBlockNumber)
+        sinon.assert.calledWith(spy, `withdrawToOwnerIfNeeded: ${message}`)
```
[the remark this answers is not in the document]
shahafn: That's the function argument...

packages/dev/test/RelayServer.test.ts
```diff
@@ -605,6 +606,212 @@ contract('RelayServer', function (accounts: Truffle.Accounts) {
...
+          await env.relayServer._resolveWithdrawalConfig(configFilename)
+          assert.fail()
+        } catch (e) {
+          assert.include(e.message, 'withdrawOnEthBalanceReached must be at least leaveManagerWithAmountEth')
+          await assertReturnedError(e.message)
+        }
+      })
+
+      it('should throw if leaveManagerWithAmountEth < managerTargetBalance', async function () {
+        const badWithdrawConfig = JSON.stringify({
```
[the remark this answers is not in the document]
shahafn: `managerTargetBalance` is the initial server config param, which is the amount of eth the server should maintain when withdrawing from the hub to itself. `leaveManagerWithAmountEth` is the amount you leave the manager after withdrawing to the owner. I thought that it might be larger, since you might want to leave a bit more to the server when withdrawing once to the owner. I can remove it if unnecessary.

packages/dev/test/RelayServer.test.ts
shahafn: No, the server doesn't stop, it logs an error on it.

packages/dev/test/RelayServer.test.ts
shahafn: Ok

packages/relay/src/RelayServer.ts
shahafn: Ok (though I didn't touch that line here...)

packages/relay/src/RelayServer.ts
shahafn: Right.

packages/dev/test/RelayServer.test.ts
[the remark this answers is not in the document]
shahafn: Now that I added a cli to withdraw, there is no need to allow withdrawal through the relay process when it's not ready.

packages/cli/src/commands/gsn-relayer-withdraw.ts
```diff
@@ -0,0 +1,44 @@
+import { CommandsLogic, WithdrawOptions } from '../CommandsLogic'
+import { gsnCommander, getKeystorePath, getServerConfig } from '../utils'
+import { createCommandsLogger } from '../CommandsWinstonLogger'
+import { KeyManager } from '@opengsn/relay/dist/KeyManager'
+import { toBN } from 'web3-utils'
+
+const commander = gsnCommander(['f', 'g'])
+  .option('-k, --keystore-path <keystorePath>', 'relay manager keystore directory', '/app/data/gsndata/manager/')
```
drortirosh, line 8: please assume default path is ` -v $PWD:$PWD -w $PWD`, and then the default path should be `./gsndata/manager`

packages/cli/src/commands/gsn.ts
```diff
@@ -8,6 +8,7 @@ commander
   .command('start', 'all-on-one: deploy all contracts, start relay')
   .command('deploy', 'deploy RelayHub and other GSN contracts instances')
   .command('relayer-register', 'stake for a relayer and fund it')
+  .command('relayer-withdraw', 'withdraw manager hub balance to owner')
```
drortirosh, line 11: `"Withdraw relayer's manager balance from RelayHub to owner."`

(users know its a relayer, the terms "maanger", "worker" are mostly internal)

packages/cli/src/utils.ts
```diff
@@ -61,6 +62,25 @@ export function getMnemonic (mnemonicFile: string): string | undefined {
   return fs.readFileSync(mnemonicFile, { encoding: 'utf8' }).replace(/\r?\n|\r/g, '')
 }
 
+export function getKeystorePath (keystorePath: string): string {
```
drortirosh, line 65: very confusing:
if path is a valid `./folder` - you return it as is
if path is `./folder/keystore` - you return `folder`

not sure the usecase for that.

packages/dev/test/RelayServer.test.ts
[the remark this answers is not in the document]
drortirosh: is the test still valid?

packages/cli/src/commands/gsn-relayer-withdraw.ts
```diff
@@ -0,0 +1,44 @@
...
+  .option('-s, --server-config <serverConfig>', 'server config file', '/app/data/config/gsn-relay-config.json')
+  .option('-b, --broadcast', 'broadcast tx after logging it to console', false)
+  .requiredOption('-a, --amount <amount>', 'amount of funds to withdraw to owner address, in wei')
```
drortirosh, line 11: for UI, please use eth, not wei (as we do in fund, stake)

packages/cli/src/commands/gsn-relayer-withdraw.ts
```diff
@@ -0,0 +1,44 @@
...
+    withdrawAmount: toBN(commander.amount),
+    keyManager,
+    config,
+    broadcast: commander.broadcast
+  }
+  console.log('config is', config)
+  console.log('withdrawalAmount is', withdrawOptions.withdrawAmount.toString())
+  console.log('broadcast is', withdrawOptions.broadcast)
+  const result = await logic.withdrawToOwner(withdrawOptions)
+  if (result.success) {
```
drortirosh, line 32: AFAIK, cli should wait for mining, and if you want to be explicit, a cli item `--nowait` to avoid waiting for mining..
displaying the TX is good thing anyway.

packages/cli/src/commands/gsn-relayer-withdraw.ts
```diff
@@ -0,0 +1,44 @@
...
+  const withdrawOptions: WithdrawOptions = {
+    withdrawAmount: toBN(commander.amount),
+    keyManager,
+    config,
+    broadcast: commander.broadcast
+  }
+  console.log('config is', config)
+  console.log('withdrawalAmount is', withdrawOptions.withdrawAmount.toString())
+  console.log('broadcast is', withdrawOptions.broadcast)
+  const result = await logic.withdrawToOwner(withdrawOptions)
```
drortirosh, line 31: cli should also:
- display current balance (e.g. if no amount is given, or `--list`)
- withdraw all (either with `--amount all` or `--all`)

packages/cli/src/commands/gsn-relayer-withdraw.ts
shahafn: Since it's only one tx that's being sent, I don't see a reason to wait for mining. Most people will open etherscan anyway to see that this hash is there and mined.

packages/cli/src/commands/gsn-relayer-withdraw.ts
shahafn: Re displaying balance, I will add though not sure if really needed. If you use this cli, you probably checked the balance already and are supposed to know how much you want to withdraw. Otherwise, this cli is not for you imo.
Re withdraw all - adding a flag to shoot yourself in the leg isn't smart. I disgree. 
If you want to withdraw all, you may as well unstake the server and wait for it to send you all the funds anyway. Withdrawing all balance cannot guarantee the manager still have enough funds both for itself and the worker address. You might need to send the balance back to the manager afterwards. That's not smart.

packages/cli/src/CommandsLogic.ts
```diff
@@ -321,6 +333,70 @@ export class CommandsLogic {
...
+        from: relayManager,
+        to: options.config.relayHubAddress,
+        value: 0,
+        gas: gasLimit,
+        gasPrice
+      })
+      console.log('Signing tx', txToSign.toJSON())
+      const signedTx = options.keyManager.signTransaction(relayManager, txToSign)
+      console.log(`signed withdrawal hex tx: ${signedTx.rawTx}`)
+      if (options.broadcast) {
```
drortirosh, line 381: what is the purpose of NOT sending the signed TX? I think its very confusing. might be needed for testing (that is default to broadcast,)

shahafn: To test the tx in view mode only, and get the tx details in case you want to broadcast it yourself. I think that it's useful to run it only in view mode, to see the tx details, since using it on mainnet can have consequences...

packages/cli/src/commands/gsn.ts
shahafn: Ok

packages/cli/src/CommandsLogic.ts
```diff
@@ -321,6 +333,71 @@ export class CommandsLogic {
...
+      const stakeManager = await this.contractInteractor._createStakeManager(stakeManagerAddress)
+      const { owner } = await stakeManager.getStakeInfo(relayManager)
+      if (owner.toLowerCase() !== options.config.ownerAddress.toLowerCase()) {
+        // throw new Error(`Already owned by ${owner}, our account=${options.from}`)
+        throw new Error(`Owner in relayHub ${owner} is different than in server config ${options.config.ownerAddress}`)
+      }
+
+      const balance = await relayHub.balanceOf(relayManager)
+      console.log('manager hub balance is', balance.toString())
+      if (balance.lt(options.withdrawAmount)) {
```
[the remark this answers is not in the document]
shahafn: Ok

drortirosh (commented):
Please fix the message.
Currently, it dumps a lot of text, which ends with: 
*Withdrew to owner successfully!*
But of course, doing nothing...

Instead, the message should be that you need to re-run with `--broadcast` to actually execute the command

(personally, I prefer "exec" or "send" over "broadcast")

packages/dev/test/RelayServer.test.ts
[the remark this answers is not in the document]
shahafn: Yes, the relayer should now withdraw when it's not ready. If you really want to withdraw now, use the cli.

packages/dev/test/RelayServer.test.ts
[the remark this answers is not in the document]
shahafn: Not relevant anymore since it was removed.

packages/dev/test/RelayServer.test.ts
shahafn: Not relevant anymore..

packages/cli/src/commands/gsn-relayer-withdraw.ts
shahafn: Was already fixed...

packages/cli/src/utils.ts
shahafn: KeyManager actually takes `workdir` and not the actual keystore file. If given a dir, it looks for `keystore` filename inside. So I resolve that by either getting the keystore file and passing the dirname to the KeyManager, or by getting the dir itself and passing it as-is. Why does it matter that much?

packages/cli/src/commands/gsn-relayer-withdraw.ts
shahafn: Already fixed

drortirosh (approved):

PR #718, opened 2021-12-12
OG-658 Fixing selection default mechanism

shahafn:


packages/provider/src/KnownRelaysManager.ts
```diff
@@ -16,22 +16,34 @@ import {
...
 import { ContractInteractor } from '@opengsn/common/dist/ContractInteractor'
-
-export const EmptyFilter: RelayFilter = (): boolean => {
+import { MAX_INTEGER } from 'ethereumjs-util'
+import { toBN } from 'web3-utils'
+import BN from 'bn.js'
+
+export const DefaultRelayFilter: RelayFilter = function (registeredEventInfo: RelayRegisteredEventInfo): boolean {
+  const maxPctRelayFee = 100
+  const maxBaseRelayFee = 1e17
```
drortirosh, line 25: 1. should be exported constants 
2. on mainnet, might 0.1eth is a good value , but on (say) polygon, a maximum of 0.1 matic is no good. you might say "so the client should override the default" - but the relayers can't control the clients, and most clients will stick to defaults. 
3. So this makes the "maxBaseRelayFee" useless on cheap networks - even though I think it is more important there (relayer's cost has almost nothing to do with gas price, so a factor is less favorable)

packages/dev/test/provider/KnownRelaysManager.test.ts
```diff
@@ -252,6 +253,23 @@ contract('KnownRelaysManager 2', function (accounts) {
       assert.equal(relays.length, 1)
       assert.equal(relays[0].relayUrl, 'stakeAndAuthorization2')
     })
+
+    it('should use DefaultRelayFilter to remove unsuitable relays when none was provided', async function () {
+      const knownRelaysManagerWithFilter = new KnownRelaysManager(contractInteractor, logger, config)
+      // @ts-ignore
+      assert.equal(knownRelaysManagerWithFilter.relayFilter.toString(), DefaultRelayFilter.toString())
```
drortirosh, line 260: did you check toString that its not constant for all functions?

packages/provider/src/KnownRelaysManager.ts
```diff
@@ -16,22 +16,34 @@ import {
...
-  let score = Math.max(Number.MAX_SAFE_INTEGER - transactionCost, 0)
-  score = score * Math.pow(0.9, failures.length)
+export const DefaultRelayScore: AsyncScoreCalculator = async function (relay: RelayRegisteredEventInfo, txDetails: GsnTransactionDetails, failures: RelayFailureInfo[]): Promise<BN> {
+  const gasLimit = toBN(txDetails.gas ?? '0')
+  const gasPrice = toBN(txDetails.gasPrice ?? '0')
+  const pctFee = toBN(relay.pctRelayFee)
+  const baseFee = toBN(relay.baseRelayFee)
+  const transactionCost = baseFee.add(gasLimit.mul(gasPrice).muln((100 + pctFee.toNumber()) / 100)
+  )
+  let score = MAX_INTEGER.sub(transactionCost)
```
drortirosh, line 45: Why do we want to make it the max possible number? we do want to have a number higher than max possible gas so 100eth gas price * arbitrum gas limit yields 1e20 * 1e10 ~ 1e30
our current score uses only negative parameters, but we should allow algorithms to increase the score. by using MAX_INTEGER we actively prevent it.

drortirosh: (yes, and I also don't think using BN is good thing - it ties our code to some BigNumber implementation, and if/when we change to ethers, we to change the API (or carry a requirement for another BN library)
We don't need the "bit accuracy" of  big numbers at all

packages/provider/src/KnownRelaysManager.ts
shahafn: 1. Why should they be exported? it's a default implementation, by no means it has any bearing on other implementations, I don't see why we should export these.

packages/dev/test/provider/KnownRelaysManager.test.ts
shahafn: Yes, it is the function code as string.

packages/provider/src/KnownRelaysManager.ts
shahafn: I don't understand your suggestion.

drortirosh: forget it. what we need is weighted scoring, and this is not something we'll now.

drortirosh (approved):

PR #719, opened 2021-12-13
OG-662 Port releases/v2.2.x branch back to master

shahafn:


packages/provider/src/KnownRelaysManager.ts
```diff
@@ -11,22 +11,34 @@ import {
...
 import { isSameAddress } from '@opengsn/common'
-
-export const EmptyFilter: RelayFilter = (): boolean => {
+import { MAX_INTEGER } from 'ethereumjs-util'
+import { toBN } from 'web3-utils'
+import BN from 'bn.js'
+
+export const DefaultRelayFilter: RelayFilter = function (registeredEventInfo: RelayRegisteredEventInfo): boolean {
+  const maxPctRelayFee = 100
+  const maxBaseRelayFee = 1e17
```
forshtat, line 20: I am opposed to committing "magic numbers" in any case; 
At least make max- as constructor parameters, to allow easy override

forshtat: We MUST have a LOG to indicate that there was a relay and it was THROWN OUT, otherwise this is a silent censorship mechanism with a hard-coded configuration.

forshtat (changes requested):

packages/provider/src/KnownRelaysManager.ts
drortirosh: I don't really care... we need to change it anyway into configurable parameter, but that's outside the scope of this issue, so its OK right know 
(the alternative is remove completely the new filtering mechanism until we implement the full configurable config - but I think its better to have this one.)

drortirosh (approved):

forshtat:
@shahafn this PingFilter will not work at all. It is not called on an array of ping responses, it is just called on a response callback (`RelaySelectionManager::_getRelayAddressPing`) so the problem is not solved at all.

forshtat (dismissed):

forshtat (approved):

PR #721, opened 2021-12-13
Adding option to withdraw from manager eth account balance as well

shahafn:
Instead of only withdrawing from hub deposit.


drortirosh:
1. generally it works ok, both "-e" and "-d"
2. don't display "wei" balance. its pure noise
3. no command just to display balances: I do "-d 1000" so it will fail (after displaying the balance..)
4. to know both balance and deposit, must run TWO commands, which is silly.
5. naming inconsistent: 
      "-d"  is "hub balance" 
      "-e" is "account balance"

my suggestions:
* always show both deposit and balance
* (this way even no-arg will show the balances, without "-a 1000")
* -a  --from-account  - withdraw from account balance
* -d  --from-deposit   - withdraw from relayhub deposit
e.g.
```
> gsn relayer-withdraw --from-account 1
...
RelayHub balance 2.3 eth
account balance:  0.5 eth
Failed to withdraw: account balance too low
```

(I also think it would help to add an easy way to run the tx on a fork, e.g. adding optional "-n")

drortirosh (approved):

PR #722, opened 2021-12-14
Prevent calls to eth_feeHistory if not supported by node

forshtat:


packages/cli/src/CommandsLogic.ts
```diff
@@ -471,7 +471,8 @@ export class CommandsLogic {
         .contract(json)
         .deploy(constructorArgs)
       const estimatedGasCost = await sendMethod.estimateGas()
-      const maxCost = new BN(options.gasPrice).mul(new BN(options.gas))
+      console.log('options', JSON.stringify(options, null, 2))
```
shahafn, line 474: Please remove this log

drortirosh (approved):

PR #724, opened 2021-12-15
use estimateGas instead of fixed cost

drortirosh:
perform gas estimation when doing value-transfer, instead of assuming
fixed-cost (which is different on rollups)

NOTE: should perform such estimates for other transactions, not only
value transfers.


forshtat (changes requested):

drortirosh:
@forshtat  - looks ok. can you approve (I can't - I created it..)

forshtat (approved):

PR #731, opened 2021-12-20
Implement solidity code changes necessary to support batching

forshtat:


packages/contracts/src/GatewayForwarder.sol
```diff
@@ -0,0 +1,28 @@
...
+        ForwardRequest calldata req,
+        bytes32 domainSeparator,
+        bytes32 requestTypeHash,
+        bytes calldata suffixData,
+        bytes calldata sig)
+    internal
+    override
+    view
+    {
+        if (msg.sender != trustedRelayHub) {
```
drortirosh, line 24: this check is counter-intuitive. please add a comment (with trustedRelayHub, we assume the relayHub verifies the signature)

packages/contracts/src/RelayHub.sol
```diff
@@ -71,6 +72,11 @@ contract RelayHub is IRelayHub, Ownable {
         relayRegistrar = _relayRegistrar;
     }
 
+    function setBatchGateway(address _batchGateway) external onlyOwner {
+        require(batchGateway == address(0), "batchGateway already set");
```
drortirosh, line 76: setting a batch gateway changes the security model of the RelayHub. it is not something that should be changeable, even with governance decision.
It should be settable in the constructor only, just like penalizer and stake manager.

packages/contracts/src/RelayHub.sol
```diff
@@ -176,8 +183,12 @@ contract RelayHub is IRelayHub, Ownable {
     {
         RelayCallData memory vars;
         vars.initialGasLeft = aggregateGasleft();
+        vars.relayRequestId = GsnUtils.getRelayRequestID(relayRequest, signature);
         require(!isDeprecated(), "hub deprecated");
         vars.functionSelector = relayRequest.request.data.length>=4 ? MinLibBytes.readBytes4(relayRequest.request.data, 0) : bytes4(0);
+        if (msg.sender == batchGateway){
+        require(signature.length == 0, "batch gateway signature not zero");
```
drortirosh, line 190: indents.

packages/contracts/src/RelayHub.sol
```diff
@@ -31,6 +31,7 @@ contract RelayHub is IRelayHub, Ownable {
 
     IStakeManager public immutable override stakeManager;
     address public immutable override penalizer;
+    address public override batchGateway;
```
drortirosh, line 34: should be immutable. just like all other code-related items (sm, penalizer)

packages/contracts/src/RelayHub.sol
```diff
@@ -176,16 +180,21 @@ contract RelayHub is IRelayHub, Ownable {
...
-        require(vars.relayManager != address(0), "Unknown relay worker");
-        require(relayRequest.relayData.relayWorker == msg.sender, "Not a right worker");
-        require(
-            isRelayManagerStaked(vars.relayManager),
-            "relay manager not staked"
-        );
+        if (msg.sender == batchGateway){
+            require(signature.length == 0, "batch gateway signature not zero");
+        } else {
+            require(msg.sender == tx.origin, "relay worker must be EOA");
```
drortirosh, line 189: should add `require(signature.length!=0)`. 
Otherwise, anyone can send a request using Gatewayforwarder, on behalf of another sender

packages/contracts/src/RelayHub.sol
```diff
@@ -176,16 +180,21 @@ contract RelayHub is IRelayHub, Ownable {
...
-        require(msg.sender == tx.origin, "relay worker must be EOA");
-        vars.relayManager = workerToManager[msg.sender];
-        require(vars.relayManager != address(0), "Unknown relay worker");
-        require(relayRequest.relayData.relayWorker == msg.sender, "Not a right worker");
-        require(
-            isRelayManagerStaked(vars.relayManager),
-            "relay manager not staked"
-        );
+        if (msg.sender == batchGateway){
+            require(signature.length == 0, "batch gateway signature not zero");
```
drortirosh, line 187: remove this zero-length sig check: you prevent batching of ECDSA requests. the required check is below (non-empty signature UNLESS its a batch gateway)

packages/contracts/src/utils/RelayHubValidator.sol
```diff
@@ -26,7 +26,8 @@ library RelayHubValidator {
             dynamicParamSize(approvalData) +
             dynamicParamSize(relayRequest.request.data) +
             dynamicParamSize(relayRequest.relayData.paymasterData);
-        require(signature.length <= 65, "invalid signature length");
+        // zero-length signature is allowed in a batch relay transaction
+        require(signature.length <= 65 || signature.length == 0, "invalid signature length");
```
drortirosh, line 30: not needed: zero is a special case of <= 65...

packages/dev/test/RelayHub.test.ts
```diff
@@ -772,6 +775,56 @@ contract('RelayHub', function ([_, relayOwner, relayManager, relayWorker, sender
...
+          const batchGateway = other
+
+          let gatewayForwarder: GatewayForwarderInstance
+          let relayHubInstance: RelayHubInstance
+          let recipientContract: TestRecipientInstance
+          let relayRequest: RelayRequest
+
+          beforeEach(async function () {
+            relayRequest = cloneRelayRequest(sharedRelayRequestData)
+            gatewayForwarder = await GatewayForwarder.new()
```
drortirosh, line 788: any special reason to construct all contracts before each test?

packages/dev/test/RelayHub.test.ts
```diff
@@ -772,6 +775,56 @@ contract('RelayHub', function ([_, relayOwner, relayManager, relayWorker, sender
...
+              from: senderAddress,
+              value: ether('1')
+            })
+            relayRequest.request.to = recipientContract.address
+            relayRequest.request.data = recipientContract.contract.methods.emitMessageNoParams().encodeABI()
+            relayRequest.relayData.paymaster = paymasterContract.address
+            relayRequest.relayData.forwarder = gatewayForwarder.address
+          })
+
+          it('should reject relayCall with non-empty signature coming from the BatchGateway', async function () {
```
drortirosh, line 804: wrong test: 
should reject empty signature coming from non-batch-gateway (also commented on the relayhub on this)

packages/contracts/src/GatewayForwarder.sol
forshtat: Fixed

packages/contracts/src/RelayHub.sol
forshtat: Fixed.

packages/contracts/src/RelayHub.sol
forshtat: Wider issue - Gateway must verify signature if it is set even coming from TrustedGateway, as otherwise RelayHub will have to verify signatures. Fixed (I think)

packages/contracts/src/RelayHub.sol
forshtat: Fixed.

drortirosh: no, it isn't.
its: `address public override batchGateway`
which is not immutable...

packages/contracts/src/RelayHub.sol
drortirosh: ^^

packages/dev/test/RelayHub.test.ts
drortirosh: (that is "before" is enough)

packages/contracts/src/RelayHub.sol
forshtat: Fixed

packages/contracts/src/RelayHub.sol
drortirosh: where did you fix that?

packages/contracts/src/utils/RelayHubValidator.sol
drortirosh: remove zero-check. it is covered by signature.length <=

packages/dev/test/RelayHub.test.ts
forshtat: There is a weird order the framework calls 'before/beforeEach' in a nested test. I could not figure it out but the inner 'beforeEach' is run after the outer 'beforeEach', which is something I do need here.

packages/dev/test/RelayHub.test.ts
forshtat: Fixed.

packages/contracts/src/RelayHub.sol
forshtat: Fixed.

packages/contracts/src/RelayHub.sol
forshtat: The new code looks like this:

```
        if (msg.sender != batchGateway){
            require(signature.length != 0, "missing signature or bad gateway");
            ...
            ...
```

so only BatchGateway can send a 0-length signature transactions.

Also in GatewayForwarder, the check is now

```
        if (msg.sender != trustedRelayHub || sig.length != 0) {
            super._verifySig(req, domainSeparator, requestTypeHash, suffixData, sig);
        }
```

so if the signature is set it will be verified.

My assumption is that it means there is no way to sneak a transaction that is not signed by the user through the GSN.

Am I correct?

packages/contracts/src/RelayHub.sol
forshtat: Now fixed.

packages/contracts/src/RelayHub.sol
```diff
@@ -176,16 +180,20 @@ contract RelayHub is IRelayHub, Ownable {
...
-        require(msg.sender == tx.origin, "relay worker must be EOA");
-        vars.relayManager = workerToManager[msg.sender];
-        require(vars.relayManager != address(0), "Unknown relay worker");
-        require(relayRequest.relayData.relayWorker == msg.sender, "Not a right worker");
-        require(
-            isRelayManagerStaked(vars.relayManager),
-            "relay manager not staked"
-        );
+        if (msg.sender != batchGateway){
+            require(signature.length != 0, "missing signature or bad gateway");
```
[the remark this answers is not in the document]
forshtat: You can, theoretically, use a gateway that is not trusted by this hub and this will cause a 'missing signature' error. Do you prefer it this way?

packages/contracts/src/RelayHub.sol
```diff
@@ -176,16 +180,20 @@ contract RelayHub is IRelayHub, Ownable {
...
+            require(msg.sender == tx.origin, "relay worker must be EOA");
+            vars.relayManager = workerToManager[msg.sender];
+            require(vars.relayManager != address(0), "Unknown relay worker");
+            require(relayRequest.relayData.relayWorker == msg.sender, "Not a right worker");
+            require(
+                isRelayManagerStaked(vars.relayManager),
```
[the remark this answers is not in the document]
forshtat: That would mean using a 'tx.origin' which is not good. Or it can be done for all transactions in the batch by a Gateway. What do you think?

packages/contracts/src/utils/RelayHubValidator.sol
forshtat: Fixed.

packages/contracts/src/RelayHub.sol
drortirosh: not sure what you mean by the assumption. anyone can create a transaction and call the relayHub (or gateway). our code should be resilient against such "rogue" transactions
we need to make sure that if it is sent directly to relayHub, it should fail (since the signature is not from user) and if it is sent via a gateway, it would fail, since the gateway will validate it (since it will fail the batch signature)

packages/contracts/src/RelayHub.sol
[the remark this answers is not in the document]
drortirosh: regardless of tx.origin: if a transaction is signed to pay a given relayManager, it has to be staked.

packages/dev/test/RelayHub.test.ts
drortirosh: I don't think its weird: it should always run the "before" in order from outer to inner, and the "after" in the inner-to-outer order.
do you need anything else?

packages/contracts/src/RelayHub.sol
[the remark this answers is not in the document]
forshtat: Fixed. Do note that it is not a very restricting check as there is no way in the RelayHub to validate the relay worker actually sending the transaction, or to penalize the relay in any way.

packages/contracts/src/RelayHub.sol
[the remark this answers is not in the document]
drortirosh: the call comes either from a trusted gateway, or from a relayer.
it means that this signature check should be moved down - report missing signature length only after verfiying it came from a relayer.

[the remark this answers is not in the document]
forshtat: I don't see a significant difference in where the check is. It will come to the same result. The transaction does not come from a valid Worker => it will be rejected. 
`"relay worker must be EOA"`, `"Not a right worker"`, `"Unknown relay worker"`, `"relay manager not staked"` will all trigger if the transaction is coming from an incorrect BatchGateway.

So I added this `or bad gateway` to the first thrown exception. 
@shahafn what do you think?

packages/contracts/src/RelayHub.sol
forshtat: My "assumption" is that `with the current code in this PR` there is no way to sneak a transaction that is not signed by the user through the GSN.

drortirosh (approved):

PR #732, opened 2021-12-21
OG-670 dev fee

shahafn:


packages/common/src/Environments.ts
```diff
@@ -27,13 +27,17 @@ const defaultPenalizerConfiguration: PenalizerConfiguration = {
...
-  postOverhead: 19073,
+  gasOverhead: 61970,
+  postOverhead: 19037,
   gasReserve: 100000,
   maxWorkerCount: 10,
   minimumStake: 1e18.toString(),
   minimumUnstakeDelay: 1000,
-  maximumRecipientDeposit: 2e18.toString()
+  maximumRecipientDeposit: 2e18.toString(),
+  devInfo: {
```
forshtat, line 37: Why does it have to be an object? We tend to prefer flat configuration types.

packages/contracts/src/RelayHub.sol
```diff
@@ -42,6 +42,7 @@ contract RelayHub is IRelayHub, Ownable {
     }
 
     function setConfiguration(RelayHubConfig memory _config) public override onlyOwner {
+        require(_config.devInfo.fee < 100, "dev fee too high");
```
forshtat, line 45: We may consider having more than 2 decimal points for percentage. I would go as far as, say, supporting 1/100'000th as a dev fee.

packages/contracts/src/interfaces/IRelayHub.sol
```diff
@@ -22,6 +22,14 @@ interface IRelayHub {
         uint256 minimumUnstakeDelay;
         // Minimum stake a relay can have. An attack on the network will never cost less than half this value.
         uint256 minimumStake;
+        // Developers information
+        DevInfo devInfo;
```
forshtat, line 26: Same here - not sure extra 2-field struct is justified.

forshtat (changes requested):

forshtat:
I think this code is lacking a 'dev tip' mode where the Hub does not enforce the dev fee but it is a part of RelayRequest.

packages/common/src/Environments.ts
shahafn: It doesn't have to be, it just seems reasonable to pack them to a struct...

packages/contracts/src/RelayHub.sol
shahafn: If that's the case, we may as well consider that for relayer fe

shahafn:
@forshtat we've discussed that specifically and decided against it.. not sure why you bring it up here.

packages/contracts/src/RelayHub.sol
forshtat: I guess we should have. What do you think?

packages/common/src/Environments.ts
forshtat: Well, I'd prefer to keep it flat.

packages/contracts/src/RelayHub.sol
shahafn: Not sure honestly, since the trade off is clear between granularity and convenience. It really depends on how much we really need this precision when more relayers/clients are joining the pool.

packages/contracts/src/interfaces/IRelayHub.sol
```diff
@@ -22,6 +22,14 @@ interface IRelayHub {
...
+    }
+
+    struct DevInfo {
+        address addr;
+        // 0 < fee < 100, as percentage of total charge from paymaster to relayer
+        uint256 fee;
```
drortirosh, line 32: 1. I dislike inner struct too.
2. name it "pct", since its percent, like normal fee.

packages/dev/test/RelayHubGasCalculations.test.ts
```diff
@@ -464,114 +472,127 @@ contract('RelayHub gas calculations', function ([_, relayOwner, relayWorker, rel
...
+            [0, 20].forEach(devFee => {
+              // avoid duplicate coverage checks. they do the same, and take a lot of time:
+              if (requestedFee !== 0 && messageLength !== 0 && process.env.MODE === 'coverage') return
+              // 50k tests take more than 10 seconds to complete so will run once for sanity
+              if (messageLength === 50000 && requestedFee !== 10) return
+              it(
+                `should compensate relay with requested fee of ${requestedFee.toString()}%, dev fee of ${devFee.toString()}% and ${messageLength.toString()} calldata size`,
+                async function () {
+                  let gasOverhead = parseInt(defaultEnvironment.relayHubConfiguration.gasOverhead.toString())
+                  if (devFee !== 0) {
```
drortirosh, line 488: huh? read doesn't matter for zero/nonzero

shahafn: It also includes a 'write'

packages/contracts/src/interfaces/IRelayHub.sol
shahafn: 1. Ok ok, I will flatten it.
2. Ok

packages/dev/test/RelayHubGasCalculations.test.ts
drortirosh: hard-coding "2800" this way means it going to cause headache when moved to different networks (e.g. arbitrum)

packages/common/src/Environments.ts
```diff
@@ -27,13 +27,15 @@ const defaultPenalizerConfiguration: PenalizerConfiguration = {
 }
 
 const defaultRelayHubConfiguration: RelayHubConfiguration = {
-  gasOverhead: 54771,
-  postOverhead: 19073,
+  gasOverhead: 61976,
```
drortirosh, line 30: notice that devFee just added 7200 gas to the request.
(can be somewhat improved, by arranging fields into same-slot. e.g.  devFee and devAddress. need to be added when entire config is optimized)

packages/dev/test/RelayHubGasCalculations.test.ts
drortirosh: ?

shahafn: Fixed

packages/common/src/Environments.ts
```diff
@@ -72,7 +74,8 @@ const ganacheLocal: Environment = {
   mintxgascost: 21000,
   gtxdatanonzero: 16,
   gtxdatazero: 4,
-  getGasPriceFactor: 1
+  getGasPriceFactor: 1,
+  nonZeroDevOverhead: 7847
```
drortirosh, line 78: 1. I hate the idea of more gas estimation params. Should be based on estimateGas. 
2. But even if we do add it, we need a way to calculate it.

drortirosh (changes requested):

packages/common/src/Environments.ts
shahafn: The way to calculate it is not different than the current way we do it, through tests, logging the diff, then fixing the gasOverhead based on it.

packages/contracts/src/RelayHub.sol
```diff
@@ -381,6 +386,15 @@ contract RelayHub is IRelayHub, Ownable {
         }
     }
 
+    function calculateDevCharge(uint256 charge) public view returns (uint256){
+        if (config.devFee == 0){ // save some gas in case of zero dev charge
+            return 0;
+        }
+        unchecked {
+        return charge.mul(config.devFee).div(100);
```
drortirosh, line 394: `unchecked` only works on your current function, not inner functions.
it will only take effect if you do `charge*config.devFee/100`

shahafn: Fixed

packages/common/src/Environments.ts
```diff
@@ -13,6 +13,7 @@ export interface Environment {
   readonly gtxdatazero: number
   readonly dataOnChainHandlingGasCostPerByte: number
   readonly getGasPriceFactor: number
+  readonly nonZeroDevOverhead: number
```
forshtat, line 16: First, I assume this is the gas cost of transferring the dev fee? So it is only added when the fee is "not zero"? 
I would not name it like that. We have "gtxdatanonzero" and it is related to a calldata byte and this comes directly from the Yellow Paper.

How about "devFeeTransferringOverhead"?


Second, I don't see it being used outside of tests. Maybe we shouldn't add an Environment variable that is not used in code?

packages/contracts/src/RelayHub.sol
```diff
@@ -381,6 +386,15 @@ contract RelayHub is IRelayHub, Ownable {
         }
     }
 
+    function calculateDevCharge(uint256 charge) public view returns (uint256){
+        if (config.devFee == 0){ // save some gas in case of zero dev charge
+            return 0;
+        }
+        unchecked {
```
forshtat, line 393: Is it only for gas optimisations? If so, how much gas do we save doing it?

forshtat (changes requested):

packages/common/src/Environments.ts
shahafn: It should be used to set new gasOverhead once we define dev fee.

packages/contracts/src/RelayHub.sol
shahafn: It saves a few hundreds, and it's safe.

packages/common/src/Environments.ts
forshtat: 1. I asked two questions. 
2. I don't understand your answer.

shahafn: 1. I think the name is fine
2. Read again

forshtat (approved):

forshtat (approved):

PR #733, opened 2022-01-03
OG-681: Add 'solpp' module dependency and JS command script

forshtat:


packages/contracts/src/RelayHub.sol
```diff
@@ -174,6 +176,9 @@ contract RelayHub is IRelayHub, Ownable {
     override
     returns (bool paymasterAccepted, bytes memory returnValue)
     {
+        // #if ENABLE_CONSOLE_LOG
```
drortirosh, line 179: better find a way to use a macro rather than "ifdef" wrapping.

drortirosh (commented):
also, not sure where do you use it ?

packages/contracts/src/RelayHub.sol
drortirosh: I still think this ifdef wrapping makes using logging cumbersome and far less intuitive

packages/dev/package.json
```diff
@@ -13,14 +13,12 @@
   "scripts": {
     "tsc": "tsc",
     "watch-tsc": "tsc -w",
-    "truffle-compile": "truffle compile",
     "lint": "eslint -f unix .",
     "lint-fix": "yarn lint --fix",
-    "test": "lerna run tsc && truffle --compile-all --network npmtest test",
+    "solpp": "yarn --cwd=\"../..\" solpp",
+    "test": "yarn solpp && truffle --compile-all --network npmtest test",
```
drortirosh, line 19: just to make sure: how do you test & compile incrementally ?

packages/contracts/src/RelayHub.sol
drortirosh: solpp can have "macro expansion", e.g.
$$(consoleLog("hello", myvar))
the point is that it will make our source code unusable without solpp - which is bad.
but the alternative of having 3 lines for each comment is also not that great...

packages/dev/package.json
forshtat: Well, the created/modified timestamps on files are carried over, so truffle takes care of compiling incrementally. It is just important to run a correct yarn script so that `solpp` is invoked.

packages/contracts/src/RelayHub.sol
forshtat: but you can do multiline stuff like
```
//#ifdef
uint a = 7;
uint b = contractB.getB();
console.log(a, b)
//#endif
```

If you want you can create a macro that would expand, but I don't think we should ban `ifdef` altogether.
This particular log is more of an example than an actual important log.

drortirosh: That's exactly what I don't want us to have: creeping conditional code. 
I wish the compiler could verify that all methods are view/pure, but at the minimum, NOT allowing extra code except actual log line is a good thing

packages/dev/package.json
drortirosh: So you must you a wrapper script and not use "truffle test" directly

packages/contracts/src/RelayHub.sol
forshtat: Well,  I strongly disagree.

packages/dev/test/RelayHubGasCalculations.test.ts
```diff
@@ -32,6 +35,12 @@ const TestPaymasterVariableGasLimits = artifacts.require('TestPaymasterVariableG
 const TestPaymasterConfigurableMisbehavior = artifacts.require('TestPaymasterConfigurableMisbehavior')
 const RelayRegistrar = artifacts.require('RelayRegistrar')
 
+const contractOrig = contract
+if (process.env.GAS_CALCULATIONS == null) {
```
drortirosh, line 39: this is ugly. a much better way is 
```
before(function() { if (...) this.skip() } )
```
and maybe it also work at the "contract" level.

packages/dev/test/StoredTransaction.test.ts
```diff
@@ -33,7 +32,7 @@ contract('StoredTransaction', function (accounts) {
       data: '0x12345678',
       nonce: 111,
       txId: '0x437a03ff976c96807aad1bc895f1535292e4ddf798fd8232600d89912961fc57',
-      value: bufferToHex(toBuffer(1e18))
+      value: `0x0${1e18.toString(16)}`
```
drortirosh, line 35: ugly. we use `import { toHex } from 'web3-utils'` all over..

scripts/solpp.js
```diff
@@ -0,0 +1,51 @@
...
+const recursiveSolidityPreprocess = async function (dirPath) {
+  const files = fs.readdirSync(dirPath)
+
+  for (const file of files) {
+    if (fs.statSync(dirPath + '/' + file).isDirectory()) {
+      await recursiveSolidityPreprocess(dirPath + '/' + file)
+    } else {
+      const orig = path.join(dirPath, '/', file)
+      const dest = orig.replace(contractsFolder, outAbiFolder)
+      console.log(dest)
```
drortirosh, line 42: why dump all filenames? at most, dump a total of files processed, and whether logs enabled/disabled

drortirosh (commented):
I'm still not sure it adds enough benefit. you must (manually) run solpp before test/build.

It will surely add some code verification problems. it modifies each source, by moving the "SPDX" comment. not sure why.
as a result, no file stays the same, even if it doesn't contain //ifdef

packages/dev/test/RelayHubGasCalculations.test.ts
forshtat: In my experience `this.skip()` works inconsistently. If this is important I can try, but I suggest ignoring the ugliness of this, as `if (process.env.GAS_CALCULATIONS == null) this.skip()` isn't much better.
Better way in my opinion would be to keep tests in separate folders, but this shouldn't be a part of this PR.

scripts/solpp.js
forshtat: Fixed.

packages/dev/test/StoredTransaction.test.ts
forshtat: Unfortunately, `toHex` won't work here.
The objects are later compared using `assert.deepEqual`, and transactions are created using buffer.toHex, which pads the value with a leading zero. This won't happen for any normal int-to-hex conversion, so I had to do a manual change. 
Note that this issue is not related to this PR per se. It is a result of a fresh `yarn install`.

forshtat:
> I'm still not sure it adds enough benefit. you must (manually) run solpp before test/build.
> 
> It will surely add some code verification problems. it modifies each source, by moving the "SPDX" comment. not sure why. as a result, no file stays the same, even if it doesn't contain //ifdef

SPDX moving is really weird :-)

It should not create too much problems, though, as there is still the same old json intermediary file generated by out truffle patch that we can upload to the etherscan.

packages/dev/test/StoredTransaction.test.ts
shahafn: Why did you remove the imports in the first place?

patches/source-map-support+0.5.16.patch
```diff
@@ -1,13 +0,0 @@
-diff --git a/node_modules/source-map-support/source-map-support.js b/node_modules/source-map-support/source-map-support.js
-index 1561cee..1f753ff 100644
---- a/node_modules/source-map-support/source-map-support.js
-+++ b/node_modules/source-map-support/source-map-support.js
```
shahafn, line 4: Why did you remove the patch?

packages/dev/package.json
forshtat: > So you must you a wrapper script and not use "truffle test" directly

Absolutely correct.

drortirosh (approved):
approved, though I still strongly believe this the the wrong solution...

PR #735, opened 2022-01-07
OG-679: FOR v2.2.x: Fix V and add support for Ledger and Trezor hardware wallets

forshtat:


drortirosh (approved):

drortirosh:
when merging in, please change the description "to support signing with Ledger"

PR #737, opened 2022-01-07
OG-679: FOR master: Fix V and add support for Ledger and Trezor hardware wallets

forshtat:


drortirosh (approved):
its OK, but when squashing it, add "for ledger support" to the git commit.
(I'm not sure whether it is needed for trezor,)

PR #738, opened 2022-01-07
OG-656: Consider a transaction 'pending' for 15 blocks after mining

forshtat:
There is an inherent race condition between server sending a transaction
and observing the result in emitted event logs.

Up until now, the code considered the transaction to stop being
'pending' the moment it is mined. However, it appears to take more time
to get the RPC node to provide a satisfying response to
'getTransactionReceipt' and 'getPastEvents', causing the server to send
multiple identical transactions.

### DIFF
--- a/packages/relay/src/RegistrationManager.ts
+++ b/packages/relay/src/RegistrationManager.ts
@@ -137,11 +137,13 @@ export class RegistrationManager {
       switch (eventData.event) {
         case RelayServerRegistered:
           if (this.lastMinedRegisterTransaction == null || isSecondEventLater(this.lastMinedRegisterTransaction, eventData)) {
+            this.logger.warn('New lastMinedRegisterTransaction: ' + JSON.stringify(eventData))
             this.lastMinedRegisterTransaction = eventData
           }
           break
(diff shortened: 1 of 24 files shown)


drortirosh:
I'm not sure this is the right solution to the problem (repeated submission of admin TXs). but even if it is, adding 15*12sec = 3 minutes delay is an exaggeration: AFAIK, there is 1-2 block slippage between nodes. no need to wait 15.

packages/relay/src/RegistrationManager.ts
```diff
@@ -137,11 +137,13 @@ export class RegistrationManager {
       switch (eventData.event) {
         case RelayServerRegistered:
           if (this.lastMinedRegisterTransaction == null || isSecondEventLater(this.lastMinedRegisterTransaction, eventData)) {
+            this.logger.warn('New lastMinedRegisterTransaction: ' + JSON.stringify(eventData))
```
drortirosh, line 140: do we need those ? even if we do, they are debug, not warnings.

forshtat:
Will only include in 'master' branch.

PR #739, opened 2022-01-09
OG-674: Add exponential backoff to '_getPastEventsPaginated' method

forshtat:


packages/common/src/ContractInteractor.ts
```diff
@@ -574,11 +578,38 @@ export class ContractInteractor {
...
+        console.error('_getPastEventsPaginated failed', error)
+        if (!this.getPastEventsExponentialBackoff) {
+          throw error
+        }
+        // exponential backoff up to 1000 sec = 16 minutes
+        await this.sleep(delay)
+        delay = Math.min(delay * STEP, MAXIMUM_DELAY)
+      }
+    }
+    throw new Error(`_getPastEventsPaginated failed after ${MAX_RETRY_COUNT} attempts`)
```
drortirosh, line 605: you lost the error here. do it inside the loop, so you can throw the nth error

packages/common/src/ContractInteractor.ts
```diff
@@ -574,11 +578,38 @@ export class ContractInteractor {
...
+    const INITIAL_DELAY = 1000
+    const MAXIMUM_DELAY = 500000
+    const STEP = 2
+    let delay = INITIAL_DELAY
+    for (let i = 0; i < MAX_RETRY_COUNT; i++) {
+      try {
+        return await this._getPastEventsPaginatedInternal(contract, names, extraTopics, options)
+      } catch (error) {
+        console.error('_getPastEventsPaginated failed', error)
+        if (!this.getPastEventsExponentialBackoff) {
```
drortirosh, line 597: if (!this.getPastEventsExponentialBackoff || i> MAX_RETRY_COUNT ) throw

forshtat:
On a closer look, this is not solvable as our default behaviour is to try to split the getPastEvents into smaller pages, and it is not easy to figure out the reason for the error, so it is better to keep the old approach and fail the relay server process if 'getPastEvents' is stuck in failure.

PR #740, opened 2022-01-10
OG-664 Client website-supplied configuration

shahafn:


packages/common/src/ConfigResponse.ts
```diff
@@ -0,0 +1,59 @@
...
+}
+
+export interface ConfigEntry {
+  name: string
+  gsnConfig: Partial<GSNConfig>
+}
+
+/**
+ * @field methodSuffix - allows use of versioned methods, i.e. 'eth_signTypedData_v4'. Should be '_v4' for Metamask
+ * @field jsonStringifyRequest - should be 'true' for Metamask, false for ganache
```
forshtat, line 26: As you are modifying it, let's either have all comments up here or all comments down where the variable is located, but not both.

packages/common/src/ConfigResponse.ts
```diff
@@ -0,0 +1,59 @@
...
+  userId?: string
+  applicationId?: string
+}
+
+export interface ConfigResponse {
+  networks: Record<number|string, ConfigEntry>
+}
+
+export interface ConfigEntry {
+  name: string
```
forshtat, line 20: Name of what?

packages/common/src/ConfigResponse.ts
```diff
@@ -0,0 +1,59 @@
...
+
+export interface LoggerConfiguration {
+  logLevel: NpmLogLevel
+  loggerUrl?: string
+  userId?: string
+  applicationId?: string
+}
+
+export interface ConfigResponse {
+  networks: Record<number|string, ConfigEntry>
```
forshtat, line 16: What do you need a 'Record' for?

packages/common/src/HttpClient.ts
```diff
@@ -44,4 +46,11 @@ export class HttpClient {
     this.logger.info(`auditTransaction response: ${JSON.stringify(auditResponse)}`)
     return auditResponse
   }
+
+  async getNetworkConfiguration (): Promise<ConfigResponse> {
+    const gsnUrl = `https://raw.githubusercontent.com/opengsn/gsn-networks/${gsnRuntimeVersion}/client-config.json`
```
forshtat, line 51: Why not request a flat configuration JSON for a specific network? Why get all networks' configurations? Most Dapps know very well where they are deployed and don't need to support any other network. I would say the only exception is CTF.

packages/provider/src/RelayClient.ts
```diff
@@ -469,12 +470,47 @@ export class RelayClient {
...
+        if (errorAsBoolean(e)) {
+          reject(e)
+        } else if (errorAsBoolean(r.error)) {
+          reject(r.error)
+        } else {
+          chainId = parseInt(r.result)
+          resolve(r.result)
+        }
+      })
+    })
```
forshtat, line 492: Please extract to `getEthChainIdResponse()` or something like that

packages/provider/src/RelayClient.ts
```diff
@@ -469,12 +470,47 @@ export class RelayClient {
...
+    const useGsnDocsConfig = config.useGsnDocsConfig != null ? config.useGsnDocsConfig : defaultGsnConfig.useGsnDocsConfig
```
forshtat, line 493: `config.useGsnDocsConfig ?? defaultGsnConfig.useGsnDocsConfig` ?

packages/common/src/HttpClient.ts
forshtat, line 51: Also, very very very strongly opposed to a hard-coded URL in code. If something must be configurable in this world, that's a URL.

packages/common/src/ConfigResponse.ts
forshtat, line 16: And why `number|string`? Do you mean the response would look like
```
"1" : { "bla": "bla bla" },
"mainnet": { "bla": "bla bla" },
"42": { "bla": "bla" },
"rinkeby": { "bla": "bla" }
``` 
???

packages/provider/src/RelayClient.ts
```diff
@@ -469,12 +470,47 @@ export class RelayClient {
...
     return {
       ...defaultGsnConfig,
+      ...configFromServer,
       ...removeNullValues(config)
     }
   }
 
+  async _resolveConfigurationFromServer (chainId: number): Promise<Partial<GSNConfig>> {
+    const httpClient = new HttpClient(new HttpWrapper(), this.logger)
+    const jsonConfig = await httpClient.getNetworkConfiguration()
```
forshtat, line 507: I am concerned that if the GSN will not be able to initialise without centralised GSN "servers" responding to a request, very quickly all integrations will do `useGsnDocsConfig: false` and the feature will be dead. 
Is there a reason we wouldn't `try/catch` and proceed without the server configuration?

forshtat (changes requested):
Was this idea approved by @yoavw ? I think I was not present when this was scheduled for implementation.

packages/dev/test/provider/RelayClient.test.ts
```diff
@@ -744,4 +745,65 @@ contract('RelayClient', function (accounts) {
...
+    describe('_resolveConfigurationFromServer()', function () {
+      let supportedNetworks: number[]
+      let jsonConfig: ConfigResponse
+      before('get all supported networks', async function () {
+        jsonConfig = await relayClient.dependencies.httpClient.getNetworkConfiguration()
+        supportedNetworks = Object.keys(jsonConfig.networks).map(k => parseInt(k))
+      })
+      it('should get configuration from opengsn for all supported networks', async function () {
+        for (const network of supportedNetworks) {
+          console.log(`Network ${jsonConfig.networks[network].name}`)
```
forshtat, line 800: Also, I suggest not logging big objects in tests

packages/dev/test/provider/RelayClient.test.ts
```diff
@@ -744,4 +745,65 @@ contract('RelayClient', function (accounts) {
...
+          const config = await relayClient._resolveConfigurationFromServer(network)
+          console.log('config', config)
+          const GSNConfigKeys = Object.keys(defaultGsnConfig)
+          Object.keys(config).forEach(key => assert.isTrue(GSNConfigKeys.includes(key), `key ${key} not found in GSConfig`))
```
forshtat, line 804: This line tests a file on our 'server', not this code per se...

packages/common/src/ConfigResponse.ts
shahafn: Why not?

packages/common/src/ConfigResponse.ts
shahafn: network e.g. xdai

packages/common/src/ConfigResponse.ts
shahafn: I only moved that part from GSNConfigurator.ts

packages/common/src/ConfigResponse.ts
forshtat: Will it work with [key: number]: ConfigEntry

packages/common/src/ConfigResponse.ts
shahafn: fixed

packages/provider/src/RelayClient.ts
shahafn: fixed

packages/common/src/ConfigResponse.ts
shahafn: No

packages/provider/src/RelayClient.ts
shahafn: fixed

packages/dev/test/provider/RelayClient.test.ts
shahafn: This is not a big log, just the network name

packages/common/src/ContractInteractor.ts
```diff
@@ -1129,3 +1129,22 @@ export function getRawTxOptions (chainId: number, networkId: number, chain?: str
       }, 'london')
   }
 }
+
+export async function getChainId (provider: Web3ProviderBaseInterface): Promise<number> {
```
drortirosh, line 1133: we're doing it too many times: better create `async sendAsync(provider,method, ...params)`, which performs callback handling and use that.

packages/dev/test/provider/RelayClient.test.ts
```diff
@@ -744,4 +745,67 @@ contract('RelayClient', function (accounts) {
...
+        provider: relayClient.getUnderlyingProvider(),
+        config: {}
+      })
+      assert.equal(resolvedConfig.methodSuffix, defaultGsnConfig.methodSuffix)
+      sinon.restore()
+    })
+    it('should not use website configuration if useGsnDocsConfig is false', async function () {
+      const spy = sinon.spy(relayClient, '_resolveConfigurationFromServer')
+      const config: Partial<GSNConfig> = {
+        useGsnDocsConfig: false
```
drortirosh, line 781: I'm reluctant of leaving the client with silently configuring from a server.
I prefer to add to all our documentation and samples
`{ useDefaultConfig: 'https://config.opengsn.org' }`

drortirosh (commented):
general: missing in this PR the "file format" of the configuration

packages/provider/src/RelayClient.ts
```diff
@@ -468,12 +468,34 @@ export class RelayClient {
     provider,
     config = {}
   }: GSNUnresolvedConstructorInput): Promise<GSNConfig> {
+    let configFromServer = {}
+    const chainId = await getChainId(provider)
+    const useGsnDocsConfig = config.useGsnDocsConfig ?? defaultGsnConfig.useGsnDocsConfig
+    if (useGsnDocsConfig) {
+      this.logger.debug('Reading config from docs')
```
drortirosh, line 475: its not "docs"

packages/common/src/ConfigResponse.ts
```diff
@@ -0,0 +1,57 @@
...
+  paymasterAddress?: Address
+  clientId: IntString
+  auditorsCount: number
+  requestValidSeconds: number
+  maxViewableGasLimit: IntString
+  environment: Environment
+  maxApprovalDataLength: number
+  maxPaymasterDataLength: number
+  gsnUrl: string
+  useGsnDocsConfig: boolean
```
drortirosh, line 56: its useOpenGsnConfig - nothign to do with "docs"

packages/common/src/ConfigResponse.ts
```diff
@@ -0,0 +1,57 @@
...
+  maxRelayNonceGap: number
+  paymasterAddress?: Address
+  clientId: IntString
+  auditorsCount: number
+  requestValidSeconds: number
+  maxViewableGasLimit: IntString
+  environment: Environment
+  maxApprovalDataLength: number
+  maxPaymasterDataLength: number
+  gsnUrl: string
```
drortirosh, line 55: openGsnConfigUrl
(name should resemble the "useXXX" , as they are twin items.

packages/common/src/ConfigResponse.ts
forshtat: @drortirosh I advise against using any "branding" in config. Our brand is subject to change.

packages/common/src/ContractInteractor.ts
```diff
@@ -1174,3 +1174,22 @@ export function getRawTxOptions (chainId: number, networkId: number, chain?: str
...
```
forshtat, line 1178: This particular one is not necessary. This is done to avoid calling `new Web3(provider).eth.getChainId()`, which is not costly

forshtat, line 1178: Also, this global function has no business being declared in `ContractInteractor.ts`

forshtat (changes requested):

drortirosh (approved):

forshtat (approved):

PR #741, opened 2022-01-10
OG-656: FOR MASTER: Consider a transaction 'pending' for N blocks after mining

forshtat:
There is an inherent race condition between server sending a transaction
and observing the result in emitted event logs.

Up until now, the code considered the transaction to stop being
'pending' the moment it is mined. However, it appears to take more time
to get the RPC node to provide a satisfying response to
'getTransactionReceipt' and 'getPastEvents', causing the server to send
multiple identical transactions.

Make 'recentActionAvoidRepeatDistanceBlocks' a parameter

### DIFF
--- a/packages/relay/src/ServerConfigParams.ts
+++ b/packages/relay/src/ServerConfigParams.ts
@@ -85,6 +85,8 @@ export interface ServerConfigParams {
   pastEventsQueryMaxPageSize: number
 
   environmentName?: string
+  // number of blocks the server will not repeat a ServerAction for regardless of blockchain state to avoid duplicates
+  recentActionAvoidRepeatDistanceBlocks: number
 }
 
 export interface ServerDependencies {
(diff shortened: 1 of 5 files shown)


packages/relay/src/ServerConfigParams.ts
```diff
@@ -150,7 +152,8 @@ export const serverDefaultConfiguration: ServerConfigParams = {
   requestMinValidBlocks: 3000, // roughly 12 hours (half client's default of 6000 blocks
   runPaymasterReputations: true,
   coldRestartLogsFromBlock: 1,
-  pastEventsQueryMaxPageSize: Number.MAX_SAFE_INTEGER
+  pastEventsQueryMaxPageSize: Number.MAX_SAFE_INTEGER,
+  recentActionAvoidRepeatDistanceBlocks: 10
```
shahafn, line 156: As I suggested over the phone, can we use `confirmationsNeeded`? They seem to match, instead of having another config param for it.

drortirosh: seems to be ok (I'm all for less config param), except that the REASON for them is different: confirmations is used to protect against uncles/forks, and this param against network instabilities (which in turn, and the source for forks, so it might be ok..)

shahafn: The reason for them is quite similar actually, finality issues when getting an answer from the node. You separate them because you know that there is a difference on the rpc node for the reason of the inconsistent answer, but from the relayer PoV, it waits this amount of blocks as confirmations to further the chance of finality. It doesn't matter whether it's because of a node out of sync with other nodes on the network, or a uncles/forks. The reason for the server is the same.

shahafn (approved):

PR #742, opened 2022-01-12
OG-703 add 'gsn send-request' command to master

forshtat:


packages/cli/src/CommandsLogic.ts
```diff
@@ -545,4 +545,10 @@ export class CommandsLogic {
       }
     }
   }
+
+  async getGasPrice (): Promise<string> {
+    const gasPrice = await this.contractInteractor.getGasPrice()
+    console.log(`Using network gas price of ${gasPrice}`)
```
drortirosh, line 551: display in gwei ?

packages/cli/src/commands/gsn-send-request.ts
```diff
@@ -0,0 +1,144 @@
...
+      logger
+    }
+    const input: GSNUnresolvedConstructorInput = {
+      provider,
+      config,
+      overrideDependencies
+    }
+    const relayProvider = await RelayProvider.newProvider(input).init()
+    if (privateKey != null) {
+      relayProvider.relayClient.dependencies.accountManager.addAccount(privateKey)
```
drortirosh, line 80: `relayProvider.addAccount`

packages/cli/src/commands/gsn-send-request.ts
```diff
@@ -0,0 +1,144 @@
...
+  )
+  if (commander.abiFile == null || !fs.existsSync(commander.abiFile)) {
+    const file: string = commander.abiFile
+    throw new Error(`--abiFile: ABI file ${file} does not exist`)
+  }
+  const abiJson = JSON.parse(fs.readFileSync(commander.abiFile, 'utf8'))
+  if (commander.to == null) {
+    throw new Error('--to: target address is missing')
+  }
+  const web3Contract = logic.contract(abiJson, commander.to)
```
drortirosh, line 110: you require the abi file to be ABI only: can't pass "artifact" file, which contains `{ abi: [] }` inside it
I think we should support both: either an array of items ( abiJson[0].type ) or artifact file where `.abi` is the array.

packages/cli/src/commands/gsn-send-request.ts
```diff
@@ -0,0 +1,144 @@
...
+  const method = web3Contract.methods[methodName]
+  if (method == null) {
+    throw new Error(`Method (${methodName}) is not found on contract`)
+  }
+  const methodParams = commander.methodParams
+
+  const receipt = await method(...methodParams).send({
+    from,
+    gas: 100000,
+    forceGasPrice: 8000000000
```
drortirosh, line 132: why? we have a "-g" parameter for that

packages/cli/src/CommandsLogic.ts
forshtat: Fixed.

packages/cli/src/commands/gsn-send-request.ts
forshtat: Fixed.

packages/cli/src/commands/gsn-send-request.ts
forshtat: Fixed.

shahafn (approved):

PR #745, opened 2022-01-14
OG-700 OG-701 OG-702 Fixing boosting

shahafn:
- Checking nonce is possible to boost (no gap with onchain nonce)
- If needed, rebroadcasting all previous transactions regardless of
boosting gas fees

### DIFF
--- a/packages/dev/test/NetworkSimultaion.test.ts
+++ b/packages/dev/test/NetworkSimultaion.test.ts
@@ -15,6 +15,7 @@ import { evmMine, evmMineMany, revert, snapshot } from './TestUtils'
 import { signedTransactionToHash } from '@opengsn/common/dist/Utils'
 import { GSNContractsDeployment } from '@opengsn/common/dist/GSNContractsDeployment'
 import { defaultEnvironment } from '@opengsn/common'
+import sinon from 'sinon'
 
 contract('Network Simulation for Relay Server', function (accounts) {
   const pendingTransactionTimeoutBlocks = 5
@@ -89,7 +90,7 @@ contract('Network Simulation for Relay Server', function (accounts) {
 
     before(async function () {
       await env.relayServer.txStoreManager.clearAll()
-      await sendMultipleRelayedTransactions()
+      await sendMultipleRelayedTransactions(stuckTransactionsCount)
     })
 
     describe('first time boosting stuck transactions', function () {
@@ -135,15 +136,7 @@ contract('Network Simulation for Relay Server', function (accounts) {
     describe('repeated boosting', function () {
       const expectedGasPriceAfterSecondBoost = toBN(expectedGasPriceAfterBoost).muln(12).divn(10).toString()
 
-      before('boosting transaction', async function () {
-        await env.relayServer.txStoreManager.clearAll()
-        await env.relayServer.transactionManager._initNonces()
-        await sendMultipleRelayedTransactions()
-        await evmMineMany(pendingTransactionTimeoutBlocks)
-        const latestBlock = await env.web3.eth.getBlock('latest')
-        const allBoostedTransactions = await env.relayServer._boostStuckPendingTransactions(latestBlock.number)
-        assert.equal(allBoostedTransactions.size, stuckTransactionsCount - 1)
-      })
+      before('boosting transaction', assertBoostAndRebroadcast)
 
       it('should not resend the transaction if not enough blocks passed since it was boosted', async function () {
         await evmMineMany(pendingTransactionTimeoutBlocks - 1)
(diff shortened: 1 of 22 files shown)


packages/dev/test/NetworkSimultaion.test.ts
```diff
@@ -135,15 +136,7 @@ contract('Network Simulation for Relay Server', function (accounts) {
...
-      before('boosting transaction', async function () {
-        await env.relayServer.txStoreManager.clearAll()
-        await env.relayServer.transactionManager._initNonces()
-        await sendMultipleRelayedTransactions()
-        await evmMineMany(pendingTransactionTimeoutBlocks)
-        const latestBlock = await env.web3.eth.getBlock('latest')
-        const allBoostedTransactions = await env.relayServer._boostStuckPendingTransactions(latestBlock.number)
-        assert.equal(allBoostedTransactions.size, stuckTransactionsCount - 1)
-      })
+      before('boosting transaction', assertBoostAndRebroadcast)
```
forshtat, line 139: What's going on here? How come you need to extract a function that is both a 'before' clause for this test AND a test on itself within this same describe? This is at least unconventional.

packages/dev/test/NetworkSimultaion.test.ts
```diff
@@ -161,7 +154,48 @@ contract('Network Simulation for Relay Server', function (accounts) {
...
+    describe('boosting & rebroadcasting all transactions', function () {
+      let id: string
+      before(async () => {
+        id = (await snapshot()).result
+      })
+      after(async () => {
+        await revert(id)
+      })
+      it('should boost underpriced transactions and only rebroadcast fairly priced transactions', assertBoostAndRebroadcast)
+      it('should not boost any transactions if latest on-chain nonce is lower than oldest pending tx nonce', async function () {
```
forshtat, line 166: Description is unclear. I assume it says something more in line with "should throw exception if there is a gap between last saved and last mined transaction that we don't know hot to fill, so there is no way to re-use this worker ever again", right?

packages/dev/test/NetworkSimultaion.test.ts
```diff
@@ -161,7 +154,48 @@ contract('Network Simulation for Relay Server', function (accounts) {
...
+        await env.relayServer.transactionManager._initNonces()
+        await sendMultipleRelayedTransactions(stuckTransactionsCount)
+
+        const storedTxs = await env.relayServer.txStoreManager.getAll()
+        const latestNonce = await env.web3.eth.getTransactionCount(storedTxs[0].from)
+        assert.equal(storedTxs[0].nonce, latestNonce)
+        const spy = sinon.spy(env.relayServer.logger, 'debug')
+        const message = `${storedTxs[0].from} : awaiting transaction with ID: ${storedTxs[0].txId} to be mined. creationBlockNumber: ${storedTxs[0].creationBlockNumber} nonce: ${storedTxs[0].nonce}`
+        await env.relayServer._boostStuckPendingTransactions(await env.web3.eth.getBlockNumber())
+        spy.calledWith(message)
```
forshtat, line 190: This line does not look like a valid assertion. I believe `calledWith` returns a value, and it is your job to write an assertion (isTrue, isFalse, throws etc.)

packages/dev/test/NetworkSimultaion.test.ts
```diff
@@ -161,7 +154,48 @@ contract('Network Simulation for Relay Server', function (accounts) {
...
+      it('should boost underpriced transactions and only rebroadcast fairly priced transactions', assertBoostAndRebroadcast)
+      it('should not boost any transactions if latest on-chain nonce is lower than oldest pending tx nonce', async function () {
+        const storedTxs = await env.relayServer.txStoreManager.getAll()
+        const latestNonce = await env.web3.eth.getTransactionCount(storedTxs[0].from)
+        assert.equal(storedTxs[0].nonce, latestNonce)
+        await env.relayServer.txStoreManager.removeTxsUntilNonce(storedTxs[0].from, latestNonce)
+        await evmMineMany(pendingTransactionTimeoutBlocks)
+        try {
+          await env.relayServer._boostStuckPendingTransactions(await env.web3.eth.getBlockNumber())
+          assert.fail()
```
forshtat, line 174: I used to write tests like that before I discovered the syntax we use now:
```await expect(promise).to.be.eventually.rejectedWith('Failed to sign relayed transaction for 0x4cfb3f70bf6a80397c2e634e5bdd85bc0bb189ee')```

File: packages/dev/test/NetworkSimultaion.test.ts [hunk 5]
```diff
@@ -161,7 +154,48 @@ contract('Network Simulation for Relay Server', function (accounts) {
...
+        await env.relayServer._boostStuckPendingTransactions(await env.web3.eth.getBlockNumber())
+        spy.calledWith(message)
+        await evmMineMany(pendingTransactionTimeoutBlocks - 1)
+        await env.relayServer._boostStuckPendingTransactions(await env.web3.eth.getBlockNumber())
+        spy.calledWith(message)
+        sinon.restore()
+      })
+    })
+
+    async function sendMultipleRelayedTransactions (stuckTransactionsCount: number): Promise<void> {
```
forshtat, line 198: You are shadowing local `stuckTransactionsCount`, please add an `_` or something

packages/relay/src/TransactionManager.ts
```diff
@@ -375,6 +378,11 @@ data                     | ${transaction.data}
       return boostedTransactions
     }
 
+    // Sanity-check: the oldest tx.nonce immediately follows 'latest' nonce
+    if (nonce !== oldestPendingTx.nonce) {
```
forshtat, line 382: I think this is very dangerous as there are a lot of ways this can break. For one, I am not even sure if `oldestPendingTx` will be guaranteed to be `pending` here, I don't see what prevents it from being "unconfirmed", therefore being counted towards a nonce.

forshtat (changes requested):

packages/relay/src/TransactionManager.ts
drortirosh: @forshtat - I think its ok. what else would you expect ? 
By definition, if last mined TX is "n", then we must boost (or re-submit) TX "n+1"

packages/relay/src/TransactionManager.ts
```diff
@@ -385,11 +393,15 @@ data                     | ${transaction.data}
...
-      boostedTransactions.set(transaction.txId, boostedTransactionDetails)
-      this.logger.debug(`Replaced transaction: nonce: ${transaction.nonce} sender: ${signer} | ${transaction.txId} => ${boostedTransactionDetails.transactionHash}`)
+    for (const transaction of sortedTxs) {
+      // The tx is underpriced, boost it
+      if (transaction.maxFeePerGas < newMaxFee || transaction.maxPriorityFeePerGas < newMaxPriorityFee) {
+        const boostedTransactionDetails = await this.resendTransaction(transaction, currentBlockHeight, newMaxFee, newMaxPriorityFee, isMaxGasPriceReached)
+        boostedTransactions.set(transaction.txId, boostedTransactionDetails)
+        this.logger.debug(`Replaced transaction: nonce: ${transaction.nonce} sender: ${signer} | ${transaction.txId} => ${boostedTransactionDetails.transactionHash}`)
+      } else { // The tx is ok, just rebroadcast it
+        await this.resendTransaction(transaction, currentBlockHeight, transaction.maxFeePerGas, transaction.maxPriorityFeePerGas, transaction.maxFeePerGas > parseInt(this.config.maxGasPrice))
```
drortirosh, line 403: do you count on the return value for anything? you didn't add resent TX there  (for all practical reasons, it was "boosted" )

packages/dev/test/NetworkSimultaion.test.ts
shahafn: That's because when you test something in a waterfall, a test is also a prerequisite state change for another test. It was like that before, I just created a function and added some asserts..

packages/dev/test/NetworkSimultaion.test.ts
shahafn: Yes, but I didn't want to right a book in the description

packages/relay/src/TransactionManager.ts
shahafn: Boosted in this context also means txId change, so there's no need to return it I believe, since it's not a new transaction per se, but just a rebroadcast of the same one we have in storage.

packages/dev/test/NetworkSimultaion.test.ts
shahafn: I don't like that syntax... chai annoys me

shahafn: fixed, though I don't like chai in general..

packages/dev/test/NetworkSimultaion.test.ts
shahafn: Fixed

drortirosh (approved):

packages/dev/test/NetworkSimultaion.test.ts
forshtat: At least write "should throw an exception", please :-)

packages/dev/test/NetworkSimultaion.test.ts
forshtat: > Fixed, though I don't like chai in general..

Nobody does, but what's the alternative?

packages/relay/src/TransactionManager.ts
drortirosh: depends.
afaik, you don't use the returned txs for anything, except in test, where it is used to validate the relayer method did something - and return nothing means the method did nothing.
if this is the case, then re-broadcasting is still a "something" to report.
if tests don't use these return values anymore (e.g. test for side-effect using spies), then the ret values should be removed

forshtat (approved):

PR #750, opened 2022-01-18
OG-552: FOR master: relayer throttling should not block trusted paymasters

forshtat:


packages/relay/src/RelayServer.ts
```diff
@@ -239,6 +236,9 @@ export class RelayServer extends EventEmitter {
   }
 
   async validatePaymasterReputation (paymaster: Address, currentBlockNumber: number): Promise<void> {
+    if (this._isTrustedPaymaster(paymaster)) {
```
shahafn, line 239: Please add a test for that :)

forshtat: There is no test for `validatePaymasterReputation`, so this addition is to an untested method, therefore testing it is a separate task in my opinion.

shahafn (approved):

PR #751, opened 2022-01-18
OG-552: FOR 2.2.x: relayer throttling should not block trusted paymasters

forshtat:


packages/dev/test/ReputationFlow.test.ts
```diff
@@ -11,7 +11,7 @@ const TestRecipient = artifacts.require('TestRecipient')
 const Forwarder = artifacts.require('Forwarder')
 const RelayHub = artifacts.require('RelayHub')
 
-contract('ReputationFlow', function (accounts) {
+contract.skip('ReputationFlow', function (accounts) {
```
drortirosh, line 14: why skip the reputation tests?
I think the proper thing to do do is to to add a test for the new scenario.

forshtat:
New versions of GSN v.2.x are not likely to be published.

PR #755, opened 2022-01-29
OG-411: Add ERC-165: Standard Interface Detection

forshtat:
Also, create TestRelayHub that does not enforce ERC-165 interface for
Paymasters so we can test gas calculations and relay manager balance
withdrawal logic the same way we used to.


drortirosh:
I think this PR is important - but not implmemented at the right place.

we have strict interfaces of **RelayHub** and **Forwarder**, and a paymaster should be protected when using those.
That is:

- add support for ERC-165 to IForwarder, IRelayHub
- call this method from Paymaster's setRelayHub, setForwarder

this way, a paymaster CANNOT transfer funds to an invalid relayhub (that is, the check during depositFor becomes redundant)

drortirosh:
in addition, the above change makes the versionXXX strings redundant: we currently don't test them at all.

packages/paymasters/test/PermitERC20UniswapV3Paymaster.test.ts
```diff
@@ -45,7 +45,7 @@ const IQuoter = artifacts.require('IQuoter')
 // as we are using forked mainnet, we will need to impersonate an account with a lot of DAI & UNI
 const MAJOR_DAI_AND_UNI_HOLDER = '0x47ac0fb4f2d84898e4d9e7b4dab3c24507a6d503'
 
-const GAS_USED_BY_POST = 197490
+const GAS_USED_BY_POST = 204766
```
drortirosh, line 48: what chnaged that caused an increase of 5000 in post ?

drortirosh (approved):

drortirosh (approved):

PR #756, opened 2022-01-29
OG-397: Add RelayHub, StakeManager, RelayRegistrar creation blocks

forshtat:


packages/contracts/src/interfaces/IRelayRegistrar.sol
```diff
@@ -34,6 +34,11 @@ interface IRelayRegistrar {
      */
     function registerRelayServer(uint256 baseRelayFee, uint256 pctRelayFee, string calldata url) external;
 
+    /**
+     * @return the block number in which the contract has been deployed.
+     */
+    function creationBlock() external view returns (uint256);
```
drortirosh, line 40: why not name it getCreationBlock ? 
I have a PR to change all other view items into a getter/setter model, so why add now more code to change later?

forshtat: Well, creation block cannot have a setter, but I can rename it if you want. Do you mean that I should make a variable private and only expose it through a method?

drortirosh: making it a getter will make the underlying field separate, and then it can be immutable - which means its cost is reduced to zero..

drortirosh (approved):

PR #762, opened 2022-02-04
OG-352:  make solidity comments natspec compliant and informative

forshtat:
…ration


packages/contracts/contract.hbs
```diff
@@ -0,0 +1,76 @@
...
+{{#each members}}
+
+{{type}} {{name}}
+
+{{/each}}
+
+---
+{{/each}}
+
+## Enums
```
drortirosh, line 54: please wrap enums, modifiers with "if": on most cases, these are empty structures.

packages/contracts/contract.hbs
```diff
@@ -0,0 +1,76 @@
+## `{{name}}`
```
drortirosh, line 1: don't they have a standard template ? that looks pretty generic.

drortirosh: or is it copied from OZ documentation itself?

packages/contracts/src/RelayHub.sol
```diff
@@ -89,59 +97,72 @@ contract RelayHub is IRelayHub, Ownable, ERC165 {
         setConfiguration(_config);
     }
 
+    /// @inheritdoc IRelayHub
```
drortirosh, line 100: very strange. don't they inherit by default ? I suggest asking OZ to add it at least as an option, and avoid splashing all our source with this lines.
comments are useful, but not when they needlessly modify (and uglify) the source code.

(that is, require it only in case of diamonds)

packages/contracts/src/RelayHub.sol
```diff
@@ -89,59 +97,72 @@ contract RelayHub is IRelayHub, Ownable, ERC165 {
...
     }
 
+    /// @inheritdoc IERC165
     function supportsInterface(bytes4 interfaceId) public view virtual override(IERC165, ERC165) returns (bool) {
         return interfaceId == type(IRelayHub).interfaceId ||
             interfaceId == type(Ownable).interfaceId ||
             super.supportsInterface(interfaceId);
     }
 
+    // TODO: make registrar immutable
```
drortirosh, line 152: todo: should inherit Registrar and avoid this setter at all (relevant to this commit just the same..)

packages/contracts/src/forwarder/IForwarder.sol
```diff
@@ -4,8 +4,22 @@ pragma abicoder v2;
...
 
+/**
+ * @title The Forwarder Interface
+ * @notice The contracts implementing this interface take a role of authorization, authentication and replay protection
+ * for contracts that choose to trust a `Forwarder`, instead of relying on a mechanism built into the Ethereum protocol.
+ *
+ * @notice if the `Forwarder` contract decides that an incoming `ForwardRequest` is valid, it must append 20 bytes that
+ * represent the caller to the `data` field of the request and send this new data to the target address (the `to` field)
+ *
+ * :warning: **Warning** :warning: The Forwarder can have a full control over a `Recipient` contract.
```
drortirosh, line 15: Recipient contracts should only trust forwarders that passed through security audit, otherwise they are susceptible to identity theft.

packages/contracts/contract.hbs
forshtat: It is copied, they don't present enums in the default template, I think.

packages/contracts/src/RelayHub.sol
forshtat: When I checked the parent natspec is not carried over by default. I don't mind it that much as it is just a oneliner comment.

packages/contracts/contract.hbs
forshtat: Fixed.

packages/contracts/src/forwarder/IForwarder.sol
forshtat: Fixed.

packages/contracts/src/RelayHub.sol
forshtat: Is currently working on it in a separate PR

packages/contracts/src/RelayHub.sol
drortirosh: It is not a single-line, but 60-some lines, all over the code -  and a requirement to keep adding them in the future.

according to docgen creator, they are about to add it by default:
https://github.com/OpenZeppelin/solidity-docgen/issues/356#issuecomment-1032059249

So I don't see a reason to add these lines all over the place, and later remove it - especially if we're not currently publishing these docs on our doc site.

packages/contracts/src/RelayHub.sol
```diff
@@ -382,6 +407,10 @@ contract RelayHub is IRelayHub, Ownable, ERC165 {
         bool rejectOnRecipientRevert;
     }
 
+    /**
+     * @notice This method can only by called by this `RelayHub`.
```
drortirosh, line 411: its a "@dev" (that is, this method should be "developer-only" function, or internal. but it has to be "external", so at least not document it as a real external function)

packages/contracts/src/RelayHub.sol
```diff
@@ -513,21 +545,25 @@ contract RelayHub is IRelayHub, Ownable, ERC165 {
...
         deprecationTime = _deprecationTime;
         emit HubDeprecated(deprecationTime);
     }
 
+    /// @inheritdoc IRelayHub
     function isDeprecated() public override view returns (bool) {
         return block.timestamp >= deprecationTime;
     }
 
+    /// @notice Prevents any address other than the `Penalizer` from calling this method.
```
drortirosh, line 560: its @dev, but less critical (since its not "external" in any way)

packages/contracts/src/interfaces/IPaymaster.sol
```diff
@@ -40,7 +26,33 @@ interface IPaymaster is IERC165 {
     }
 
     /**
-     * Return the Gas Limits and msg.data max size constants used by the Paymaster.
+     * @notice Return the Gas Limits for Paymaster's functions and maximum msg.data length values for this Paymaster.
```
drortirosh, line 29: There is a problem with deep documentation in the source: you're putting on-chain security assumptions that belong to our doc.

PR #764, opened 2022-02-08
OG-728: Simplify RelayRegistrar and move string operations off-chain

forshtat:
* Move string manipulation code to JavaScript
* No uncompressed struct needed, only expose one struct
* Do not store RelayHub address, accept it as function parameter
* Remove setter for RelayRegsitrar from the RelayHub
* Significantly improve code coverage for RelayRegistrar


packages/common/src/Utils.ts
```diff
@@ -314,3 +329,24 @@ export function formatTokenAmount (balance: BN, tokenDecimals: BN, tokenSymbol:
   }
   return `${fromWei(shiftedBalance)} ${tokenSymbol}`
 }
+
+export function splitRelayUrlForRegistrar (url: string, partsCount: number = 3): string[] {
+  const maxLength = 32 * partsCount
+  if (url.length > maxLength) {
+    throw new Error(`The URL does not fit to the RelayRegistrar. Please shorten it to less than ${maxLength} characters`)
+  }
+  const parts = url.match(/.{1,32}/g) ?? []
```
drortirosh, line 338: I think this does the entire split:
```
Buffer.from(url).toString('hex').match(/.{2,64}/g).map(x=>'0x'+x.padEnd(64,'0'))
```

packages/contracts/src/interfaces/IRelayRegistrar.sol
```diff
@@ -56,20 +62,27 @@ interface IRelayRegistrar is IERC165 {
      * @return `true` if this `RelayRegistrar` keeps registrations on-chain in storage in addition to emitting events.
      * `false` if only events are emitted as part of Relay Server registration on this `RelayRegistrar`.
      */
-    function isUsingStorageRegistry() external returns (bool);
+    function getIsUsingStorageRegistry() external returns (bool);
```
drortirosh, line 65: no need to rename here: the standard spec for "getter" is a method that starts with "get" or "is"

packages/contracts/src/utils/RelayRegistrar.sol
```diff
@@ -9,153 +9,124 @@ import "../interfaces/IRelayHub.sol";
...
-        RelayInfo memory info;
-        RelayStorageInfo storage storageInfo = values[relayManager];
-        require(storageInfo.lastBlockNumber != 0, "relayManager not found");
-        info.lastBlockNumber = storageInfo.lastBlockNumber;
-        info.stakeBlockNumber = storageInfo.stakeBlockNumber;
-        info.baseRelayFee = storageInfo.baseRelayFee;
-        info.pctRelayFee = storageInfo.pctRelayFee;
-        info.relayManager = relayManager;
-        info.url = packString(storageInfo.urlParts);
+    function getRelayInfo(address relayHub, address relayManager) public view override returns (RelayInfo memory) {
```
drortirosh, line 94: change return value to "RelayInfo memory info", and thus copying `info` into return value

forshtat: I am not sure this is how it works. Defining a variable name places it in the returndata memory? How do you know that?

packages/dev/test/utils/RelayRegistrar.test.ts
```diff
@@ -1,80 +1,178 @@
...
+  afterEach(async function () {
+    await revert(id)
   })
 
-  it('#splitString, packString', async () => {
-    expect(await reg.splitString('1')).to.eql(['0x31'.padEnd(66, '0'), HashZero, HashZero])
-    expect(await reg.splitString('1'.repeat(32))).to.eql(['0x' + '31'.repeat(32), HashZero, HashZero])
-    expect(await reg.splitString('1'.repeat(33))).to.eql(['0x' + '31'.repeat(32), '0x31'.padEnd(66, '0'), HashZero])
+  context('#registerRelayServer()', function () {
+    it('should fail to register if the Relay Manager does not get approved by the RelayHub', async function () {
```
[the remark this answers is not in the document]
forshtat: **does** not **get** approved

packages/common/src/Utils.ts
forshtat: This does not fill in the array of 3, therefore it is not much better.

drortirosh: Buffer.from(url).toString('hex').padEnd(96*2,'0') .

packages/common/src/Utils.ts
```diff
@@ -314,3 +329,24 @@ export function formatTokenAmount (balance: BN, tokenDecimals: BN, tokenSymbol:
   }
   return `${fromWei(shiftedBalance)} ${tokenSymbol}`
 }
+
+export function splitRelayUrlForRegistrar (url: string, partsCount: number = 3): string[] {
+  const maxLength = 32 * partsCount
+  if (url.length > maxLength) {
+    throw new Error(`The URL does not fit to the RelayRegistrar. Please shorten it to less than ${maxLength} characters`)
```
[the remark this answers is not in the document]
drortirosh: it is an error to display to its user.
your "user" is the developer of GSN, which knows what "RelayRegistar" is.
the actual user is a devop - the owner of a relayer server

drortirosh (approved):

PR #767, opened 2022-02-10
OG-731: native weth

drortirosh:


packages/cli/src/CommandsLogic.ts
```diff
@@ -307,24 +320,27 @@ export class CommandsLogic {
...
           stake.toString() === '0' ? '' : ` (already has ${currentStakeFormatted})`)
 
-        if (options.mintToken) {
-          console.log('Minting 100 tokens')
-          const mintTx = await this.contractInteractor.erc20Token.mint(100e18.toString(), {
+        const stakingTokenContract = await this.contractInteractor._createERC20(stakingToken)
+        const tokenBalance = await stakingTokenContract.balanceOf(options.from)
+        if (tokenBalance.lt(stakeValue) && isDefaultToken) {
+          console.log(`Wrapping ${formatTokenAmount(stakeValue, 18, 'eth')}`)
+          // default token is wrapped eth, so deposit eth to make then into tokens.
```
forshtat, line 327: This is a wild assumption. At least try to view-call this before sending a transaction with value to this address. Also, obviously, this behaviour cannot be a silent default. Use a flag and default it to `false`

packages/cli/src/CommandsLogic.ts
```diff
@@ -307,24 +320,27 @@ export class CommandsLogic {
...
+          const depositTx = await stakingTokenContract.deposit({ from: options.from, value: stakeValue }) as any
+          transactions.push(depositTx.transactionHash)
+        }
+
+        const currentAllowance = await stakingTokenContract.allowance(options.from, stakeManager.address)
+        if (currentAllowance.lt(stakeValue)) {
+          console.log(`Approving ${stakeValueFormatted} to StakeManager`)
```
forshtat, line 334: Please log here current allowance as well, this may be helpful.

packages/cli/src/CommandsLogic.ts
```diff
@@ -505,20 +521,23 @@ export class CommandsLogic {
     }
     await registerForwarderForGsn(fInstance, options)
 
-    let ttInstance: Contract | undefined
-    if (deployOptions.deployTestToken ?? false) {
-      ttInstance = await this.getContractInstance(TestToken, {}, undefined, { ...options }, deployOptions.skipConfirmation)
-      console.log('Setting minimum stake of 1 TestToken on Hub')
-      await rInstance.methods.setMinimumStakes([ttInstance.options.address], [1e18.toString()]).send({ ...options })
+    let stakingTokenAddress = deployOptions.stakeManagerAddress
+    if (stakingTokenAddress == null) {
```
forshtat, line 525: I don't think it is wise to deploy a contract by default. I think the flag should stay. This differentiates running this command for tests or to actually deploy the contracts. We still didn't migrate to hardhat so let's maintain this command reasonable.

packages/cli/src/CommandsLogic.ts
```diff
@@ -505,20 +521,23 @@ export class CommandsLogic {
...
+      const ttInstance = await this.getContractInstance(WrappedEthToken, {}, undefined, { ...options }, deployOptions.skipConfirmation)
+      stakingTokenAddress = ttInstance.options.address
     }
+    const stakingContact = await this.contractInteractor._createERC20(stakingTokenAddress)
+
+    console.log(`Setting minimum stake of ${deployOptions.minimumTokenStake} of ${await stakingContact.symbol()}`)
```
forshtat, line 531: Please don't use `await` inside a string formatter tags.

packages/cli/src/commands/gsn-deploy.ts
```diff
@@ -21,10 +21,11 @@ gsnCommander(['n', 'f', 'm', 'g', 'l'])
   .option('--relayRegistrar <address>', 'relayRegistrar')
   .option('--environmentName <string>', `name of one of the GSN supported environments: (${Object.keys(EnvironmentsKeys).toString()}; default: ethereumMainnet)`, EnvironmentsKeys.ethereumMainnet)
   .option('--burnAddress <string>', 'address to transfer burned stake tokens into', constants.BURN_ADDRESS)
+  .option('--stakingToken <string>', 'default staking token to use. (wrapped eth)')
```
forshtat, line 24: What does `default` and `(wrapped eth)` mean in this line?

packages/cli/src/commands/gsn-deploy.ts
```diff
@@ -21,10 +21,11 @@ gsnCommander(['n', 'f', 'm', 'g', 'l'])
...
+  .option('--minimumTokenStake <number>', 'minimum staking eth value (in wrapped eth staking token)', '1')
```
forshtat, line 25: This can be denominated in any token passed to `stakingToken`, why do you write `wrapped eth` explicitly?

packages/contracts/src/RelayHub.sol
```diff
@@ -70,6 +70,7 @@ contract RelayHub is IRelayHub, Ownable, ERC165 {
         require(token.length == minimumStake.length, "setMinimumStakes: wrong length");
         for (uint256 i = 0; i < token.length; i++) {
             minimumStakePerToken[token[i]] = minimumStake[i];
+            emit StakingToken(address(token[i]), minimumStake[i]);
```
forshtat, line 73: `MinimumStakePerTokenChanged`

packages/contracts/src/interfaces/IERC20Token.sol
```diff
@@ -8,10 +8,11 @@ import "@openzeppelin/contracts/token/ERC20/extensions/IERC20Metadata.sol";
 
 /**
  * @notice Extended ERC-20 token interface used internally in OpenGSN modules.
- * Renamed to avoid conflict with OZ namespace. Includes IERC20, ERC20Metadata and 'mint(uint256)'.
- * @notice Interface of the ERC20 standard as defined in the EIP.
+ * @dev Renamed to avoid conflict with OZ namespace. Includes IERC20, ERC20Metadata.
```
forshtat, line 11: Use 'notice' - dev is not presented in the generated .md files. Our .md files are for devs anyway.

forshtat (changes requested):

packages/cli/src/CommandsLogic.ts
```diff
@@ -238,8 +238,21 @@ export class CommandsLogic {
       const stakeManager = await this.contractInteractor._createStakeManager(stakeManagerAddress)
       const { stake, unstakeDelay, owner, token } = (await stakeManager.getStakeInfo(relayAddress))[0]
 
-      if (!(isSameAddress(token, options.token) || isSameAddress(token, constants.ZERO_ADDRESS))) {
-        throw new Error(`Cannot use specified token for staking. Specified token address: ${options.token} Currently used token: ${token}`)
+      var isDefaultToken = false
+      let stakingToken = options.token
+      if (stakingToken === constants.ZERO_ADDRESS) {
```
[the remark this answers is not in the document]
drortirosh: `--token` means "use a token instead of ETH". why add yet another parameter for that purpose?

packages/cli/src/commands/gsn-deploy.ts
drortirosh: this is denominated in the stakingToken units (which are supposed to be WETH..)
so it is just "minimum staking value"

forshtat (approved):

PR #772, opened 2022-02-16
OG-746 - force jira reference in github commits to master

drortirosh:
some more info?

### DIFF
--- a/.circleci/checkGit
+++ b/.circleci/checkGit
@@ -4,10 +4,13 @@ branch=`git branch | sed -ne 's/[*] *//p'`
 
 title=`git log --format="%h %ad, %an, %s"|head -1 `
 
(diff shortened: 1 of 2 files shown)


drortirosh:
it is more tricky than that: we don't want all commits to require "OG-xx" numbers: pr-specific branches can work with many such small commits.
its only  when squashing to master that we require a jira reference.. but if we run the check only on commit to master, the circleCI checks runs AFTER it gets merged, so we end up with failed master commit.

So the options are:
- use git hooks (via github webhooks) - requires integration with external service
- lighter check: don't check current commit, but all commits since the branch from master. note that even this is not a fail-safe check, since actual commit message is created by manual "squash", that might remove those references - but still, at least SOME commits in the branch did have such reference.

PR #774, opened 2022-02-18
OG-84: Do not request client signature if dry-run view call to 'relayCall' failed

forshtat:


packages/contracts/src/RelayHub.sol
```diff
@@ -287,14 +289,19 @@ contract RelayHub is IRelayHub, Ownable, ERC165 {
...
             require(signature.length != 0, "missing signature or bad gateway");
             require(msg.sender == tx.origin, "relay worker must be EOA");
             require(msg.sender == relayRequest.relayData.relayWorker, "Not a right worker");
         }
-        vars.relayManager = workerToManager[relayRequest.relayData.relayWorker];
-        require(vars.relayManager != address(0), "Unknown relay worker");
-        verifyRelayManagerStaked(vars.relayManager);
+
+        // solhint-disable-next-line avoid-tx-origin
+        if (tx.origin != DRY_RUN_ADDRESS) {
```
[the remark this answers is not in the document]
forshtat: As there is an entire `verifyRelayManagerStaked` that has 6 requires in it, I don't think it is feasible. Also, it is not safe. Writing solidity code and trying to keep gas usage unchanged manually while keeping branching logic is, IMO, unpractical. We will do the 'estimate gas' from a real tx.origin after we have a relay worker selected.

packages/contracts/src/forwarder/Forwarder.sol
```diff
@@ -146,6 +148,10 @@ contract Forwarder is IForwarder, ERC165 {
     virtual
     view
     {
+        // solhint-disable-next-line avoid-tx-origin
```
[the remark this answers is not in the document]
forshtat: Agree here. Fixed.

packages/common/src/Constants.ts
```diff
@@ -12,6 +12,7 @@ export const constants = {
   ZERO_ADDRESS: '0x0000000000000000000000000000000000000000',
   // OpenZeppelin's ERC-20 implementation bans transfer to zero address
   BURN_ADDRESS: '0xFFfFfFffFFfffFFfFFfFFFFFffFFFffffFfFFFfF',
+  DRY_RUN_ADDRESS: '0xdeaDDeADDEaDdeaDdEAddEADDEAdDeadDEADDEaD',
```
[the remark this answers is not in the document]
forshtat: Switched to 0x777...777, if that is ok

packages/contracts/src/forwarder/Forwarder.sol
```diff
@@ -152,7 +154,8 @@ contract Forwarder is IForwarder, ERC165 {
                 "\x19\x01", domainSeparator,
                 keccak256(_getEncoded(req, requestTypeHash, suffixData))
             ));
-        require(digest.recover(sig) == req.from, "FWD: signature mismatch");
+        // solhint-disable-next-line avoid-tx-origin
+        require(tx.origin == DRY_RUN_ADDRESS || digest.recover(sig) == req.from, "FWD: signature mismatch");
```
drortirosh, line 158: please reorder them, adding `|| tx.origin at the end`
You're preventing any possible estimation of gas, even of the internal calls.

packages/common/src/ContractInteractor.ts
```diff
@@ -458,7 +467,7 @@ export class ContractInteractor {
           method: 'eth_call',
           params: [
             {
-              from: relayCallABIData.relayRequest.relayData.relayWorker,
+              from,
```
drortirosh, line 470: some networks fail the CALL request if caller doesn't have enough to cover the gas price...
which means that in dry-run, we must have `gasPrice:0`
This, however, will alter our gas-price calculations
(BTW: this is, by itself, a good reason to have the ZERO_ADDRESS as a "DRY_RUN_ADDRESS", since zero address has balance on all networks...

packages/provider/src/RelayClient.ts
```diff
@@ -537,6 +546,61 @@ export class RelayClient {
...
+    const relayCallABI: RelayCallABI = {
+      relayRequest,
+      signature: '0x',
+      approvalData: '0x',
+      maxAcceptanceBudget: this.config.maxViewableGasLimit
+    }
+    return await this._verifyViewCallSuccessful(dryRunRelayInfo, relayCallABI, true)
+  }
+
+  async _verifyViewCallSuccessful (
```
drortirosh, line 581: the idea was to have dry-run INSTEAD of normal validation, not in addition to it.
(and its OK it is not 100% - there is always another full verification on the relayer)

forshtat: Well, I don't know if it should be instead though. It makes sense to me to check with what you can, prepare real request and then check with that as well. These are just milliseconds if everything is working correctly.

packages/contracts/src/forwarder/Forwarder.sol
forshtat: `digest.recover(sig)` will revert if `sig` is not a valid ecdsa signature, and I don't want to put a stub byte array in `sig` to reorder things. Can you please explain how reordering things will help with gas estimation? I guess you can estimate all you want as long as `tx.origin != DRY_RUN_ADDRESS`, can't you?

drortirosh: the `tx.origin== DRY_RUN_ADDRESS` check is very cheap - about 10 gas - and placing it first means it "short-circuit" and doesn't calculate the rest of the line - which means dry-run will cost ~3000 less.
Putting that check at the end means the gas "fluctuation" with and without dry-run is roughly 10 gas.
(in normal run, the sig is checked and the "`tx.origin==DRU_RUN_ADDRESS`" check is skipped.
in dry-run, the signature is checked (and fail, and the same cost), and then the dry-run check is executed

packages/contracts/src/RelayHub.sol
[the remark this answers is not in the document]
drortirosh: why not? we're going to add ~20 gas to each of those statements when done in "dry-run". so the overall cost of "dry-run" will be about 100 more than normal run. this is neglible, and adding this to gas estimation doesn't change.
with your "short-circuiting" the tests, the cost will become lower than real cost by few thousands, which makes the estimation during the dry-run useless.

packages/provider/src/RelayClient.ts
drortirosh: I think check after dry-run is completely redundant. 
I agree to leave it now to reduce the complexity of this PR - it can be removed in a future PR
I still think we should rely on the gas estimation in the dry-run: with the above "post-check" of txorigin, it add little to the total gas cost.

packages/contracts/src/RelayHub.sol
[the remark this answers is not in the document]
forshtat: I really see no point doing estimate gas with a dry-run. Where would we want to do it?

packages/contracts/src/forwarder/Forwarder.sol
forshtat: I really fail to understand where does your interest in gas consumption in dry run come from. I think it is completely irrelevant.

packages/provider/src/RelayClient.ts
forshtat: I could not disagree any more. I see no point in removing the actual real check. There is a ton of checks that are short-circuted in a dry-run mode. I don't like the idea of not doing the actual check very very much.

packages/common/src/ContractInteractor.ts
forshtat: I can make the dry-run origin address configurable. I assume some rollups won't have any Ether on address(0). What do you think?

packages/provider/src/RelayClient.ts
```diff
@@ -567,7 +584,8 @@ export class RelayClient {
         version: ''
       }
     }
-    const relayRequest = await this._prepareRelayRequest(gsnTransactionDetails, dryRunRelayInfo)
+    // TODO: clone?
+    this.fillRelayInfo(relayRequest, dryRunRelayInfo)
```
drortirosh, line 588: why do you need to fill from a dummy relayInfo ? (and the callDatacost is also irrelevant, since it doesn't cover all fields.)

packages/common/src/types/Aliases.ts
```diff
@@ -18,6 +18,11 @@ export type SemVerString = string
  * TODO: make ping filtering sane!
  */
 export type PingFilter = (pingResponse: PingResponse, gsnTransactionDetails: GsnTransactionDetails) => void
+
+/**
```
drortirosh, line 22: comment is misplaced: its not a definition of the type, but of the usage, for asyncPaymasterData callback

packages/dev/test/Flows.test.ts
```diff
@@ -228,7 +228,7 @@ options.forEach(params => {
           await approvalPaymaster.setRelayHub(rhub.address)
           await approvalPaymaster.setTrustedForwarder(await sr.getTrustedForwarder())
           await rhub.depositFor(approvalPaymaster.address, { value: (1e18).toString() })
-          relayClientConfig = { ...relayClientConfig, ...{ paymasterAddress: approvalPaymaster.address } }
+          relayClientConfig = { ...relayClientConfig, ...{ paymasterAddress: approvalPaymaster.address }, performDryRunViewRelayCall: false }
```
drortirosh, line 231: why override defaults?
we should override defaults only if we're explicitly testing a rare case, and Flows is all about testing the "usual" cases.

packages/dev/test/ServerTestEnvironment.ts
```diff
@@ -277,8 +277,11 @@ export class ServerTestEnvironment {
       const mergedTransactionDetail = Object.assign({}, gsnTransactionDetails, overrideDetails)
       // do not 'return await' here as it will defer executing the 'finally' block and enable re-stubbing
       // (will crash on 'let x = [createRelayHttpRequest(), createRelayHttpRequest()]')
-      // eslint-disable-next-line @typescript-eslint/return-await
-      return this.relayClient._prepareRelayHttpRequest(relayInfo, mergedTransactionDetail)
+      // eslint-disable-next-line @typescript-eslint/return-await,@typescript-eslint/promise-function-async
```
drortirosh, line 280: is this delibarate? you return (and actually, call "finally" too) before the async function is executed.
why not do "return  __await__ this.relayClient._prepareRelayRequest" ?

packages/dev/test/provider/RelayClient.test.ts
```diff
@@ -156,6 +157,7 @@ contract('RelayClient', function (accounts) {
     const loggerConfiguration: LoggerConfiguration = { logLevel: 'debug' }
     gsnConfig = {
       loggerConfiguration,
+      performDryRunViewRelayCall: false,
```
drortirosh, line 160: so you run all tests with dryRun = false, and only a single test with dryRun=true ..
Since we want dry-run to be the default mode, it seems that it need to be reversed.

packages/dev/test/Flows.test.ts
drortirosh: mistake. its only for a single test...

drortirosh (approved):

PR #776, opened 2022-02-23
OG-329: Log 'gsnRuntimeVersion' on RelayClient, RelayProvider creation

forshtat:


packages/provider/src/RelayClient.ts
```diff
@@ -95,6 +95,7 @@ export class RelayClient {
     this.rawConstructorInput = rawConstructorInput
     this.logger = rawConstructorInput.overrideDependencies?.logger ??
       createClientLogger(rawConstructorInput.config?.loggerConfiguration ?? defaultLoggerConfiguration)
+    this.logger.info(`Created new RelayClient ver.${gsnRuntimeVersion}`)
```
drortirosh, line 98: any special reason to report it twice ? we always have both RelayClient and RelayProvider (and RelayClient is basically an internal class...)

drortirosh (dismissed):
approved, but I really don't think we need to report it twice..

drortirosh (approved):

PR #777, opened 2022-02-23
OG-725: Extract webpack test to a separate CircleCI flow

forshtat:


drortirosh:
its ok, except that it doesn't really saves anything: it takes 43 seconds because it recompiles the contracts. the actual test takes 5 seconds (which is a lot for a test, but we need to invest in larger gains, such as extracting paymasters, which have 1-2 minutes overhead.

drortirosh (approved):

PR #783, opened 2022-03-18
OG-764: Add withdraw() to multiple targets functionality

forshtat:


packages/contracts/src/RelayHub.sol
```diff
@@ -192,11 +192,29 @@ contract RelayHub is IRelayHub, Ownable, ERC165 {
...
+            // #if ENABLE_CONSOLE_LOG
+            console.log("withdrawMultiple: balances", balances[account]);
+            console.log("withdrawMultiple: account", account);
+            console.log("withdrawMultiple: dest[i]", dest[i]);
+            console.log("withdrawMultiple: amount[i]", amount[i]);
+            // #endif
+            uint256 balance = balances[account];
+            require(balance >= amount[i], "insufficient funds");
+            balances[account] = balance.sub(amount[i]);
+            dest[i].transfer(amount[i]);
```
drortirosh, line 215: don't use "transfer" 
https://consensys.net/diligence/blog/2019/09/stop-using-soliditys-transfer-now/

packages/contracts/src/RelayHub.sol
```diff
@@ -192,11 +192,29 @@ contract RelayHub is IRelayHub, Ownable, ERC165 {
     function withdraw(uint256 amount, address payable dest) public override {
         address payable account = payable(msg.sender);
         require(balances[account] >= amount, "insufficient funds");
```
drortirosh, line 194: redundant check. will be done within withdrawMultiple

packages/contracts/src/RelayHub.sol
```diff
@@ -192,11 +192,29 @@ contract RelayHub is IRelayHub, Ownable, ERC165 {
...
+        uint256[] memory amounts = new uint256[](1);
+        address payable[] memory destinations = new address payable[](1);
+        amounts[0] = amount;
+        destinations[0] = dest;
+        withdrawMultiple(amounts, destinations);
```
drortirosh, line 199: I understand your argument regarding withdraw[single]. you're right it is better to leave it as a helper.

packages/contracts/src/RelayHub.sol
```diff
@@ -192,11 +192,29 @@ contract RelayHub is IRelayHub, Ownable, ERC165 {
...
 
-        balances[account] = balances[account].sub(amount);
-        dest.transfer(amount);
-
-        emit Withdrawn(account, dest, amount);
+    /// @inheritdoc IRelayHub
+    function withdrawMultiple(uint256[] memory amount, address payable[] memory dest) public override {
+        address payable account = payable(msg.sender);
+        for (uint256 i = 0; i < amount.length; i++) {
+            // #if ENABLE_CONSOLE_LOG
```
drortirosh, line 206: I still think that adding these logs does't help a bit, only add to clutter the logs. sending out 4 log lines adds even more. can use `log( "string with %s", a,b,c)` (though it is limited only to 3 dynamic params)

packages/contracts/src/interfaces/IRelayHub.sol
```diff
@@ -160,6 +160,13 @@ interface IRelayHub is IERC165 {
      */
     function withdraw(uint256 amount, address payable dest) external;
 
+    /**
+     * @notice Withdraws from an account's balance, sending funds to multiple provided addresses.
+     * Relay Managers call this to retrieve their revenue, and `Paymasters` can also use it to reduce their funding.
+     * Emits a `Withdrawn` event for each destination.
+     */
+    function withdrawMultiple(uint256[] memory amount, address payable[] memory dest) external;
```
drortirosh, line 168: I think the right param order is "dest first, amount after" (yes, its a mistake in the old "withdraw" function too)
notice that the `Withdrawn(dest,amount)` event.

packages/contracts/src/RelayHub.sol
drortirosh: and while you're at it, add "ifdef" around the "import console.log" too.

packages/contracts/src/RelayHub.sol
forshtat: We did use ```.transfer()``` here for years, and also in forwarder, should it be changed?

packages/contracts/src/RelayHub.sol
forshtat: Fixed.

packages/contracts/src/RelayHub.sol
forshtat: Thank you. If you know how to create dynamically-sized in-memory array without declaring it please tell me. I could not explicitly convert ```uint[1]``` to ``` uint[]```.

packages/contracts/src/interfaces/IRelayHub.sol
forshtat: Fixed.

packages/contracts/src/RelayHub.sol
forshtat: Fixed.

packages/dev/test/RelayHub.test.ts
```diff
@@ -162,6 +164,39 @@ contract('RelayHub', function ([paymasterOwner, relayOwner, relayManager, relayW
       })
     })
 
+    it('accounts with deposits can withdraw to multiple destinations', async function () {
```
drortirosh, line 167: nitpic: instead of complex withdraw, could simply withdraw "1 wei" to a random address, and 2 wei to another random address. no need for pre/post balance diff...

drortirosh (approved):

PR #784, opened 2022-03-25
OG-744: lookup transaction relayed events by RelayRequestID

forshtat:


packages/provider/src/RelayProvider.ts
```diff
@@ -131,7 +141,33 @@ export class RelayProvider implements HttpProvider, Web3ProviderBaseInterface {
...
+    const relayRequestID = payload.params[0] as string
+    const submissionDetails = this.submittedRelayRequests.get(relayRequestID)
+    if (submissionDetails == null) {
+      this._ethGetTransactionReceiptWithTransactionHash(payload, callback)
+      return
+    }
+    try {
+      const result = await this._createTransactionReceiptForRelayRequestID(relayRequestID, submissionDetails)
+      const rpcResponse = {
+        id,
```
drortirosh, line 160: why do you translate id to int ? I think its better to return id as it is (and also jsonrpc) from the original request.

forshtat: Something broke down when I didn't. I can try again. I didn't do it out of pure passion for integers if I recall correctly.

packages/provider/src/RelayProvider.ts
```diff
@@ -131,7 +141,33 @@ export class RelayProvider implements HttpProvider, Web3ProviderBaseInterface {
     })
   }
 
-  _ethSendTransaction (payload: JsonRpcPayload, callback: JsonRpcCallback): void {
+  /**
+   * The ID can be either a RelayRequestID which requires event-based lookup or Transaction Hash that goes through
+   * @param payload
+   * @param callback
+   */
+  async _ethGetTransactionReceipt (payload: JsonRpcPayload, callback: JsonRpcCallback): Promise<void> {
```
drortirosh, line 149: ugly merge. I think all internal funcs should be async and not callback (there are good "promisify" (and vice-versa) tools
this way, you can also easily call in parallel the getByHash and getByRequestId

drortirosh (approved):

PR #786, opened 2022-03-27
OG-432: Not calling .init on a Relay Provider causes an exception

forshtat:


packages/provider/src/RelayProvider.ts
```diff
@@ -62,6 +62,7 @@ export class RelayProvider implements HttpProvider, Web3ProviderBaseInterface {
     this.origProvider = this.relayClient.getUnderlyingProvider() as HttpProvider
     this.host = this.origProvider.host
     this.connected = this.origProvider.connected
+    this.logger = this.relayClient.logger
```
drortirosh, line 65: this prevent hooking, since the provider's logger might not be initialized.
my suggestion was:
in the provider define the logger as a "property"
```
get logger() { return this.relayClient.logger ?? console }
```

and now `provider.logger`  can be used as it is (the above sample falls back to the `console.log`) before it is initialized.

PR #788, opened 2022-04-04
OG-767: Allow withdrawing abandoned funds from the RelayHub as a dev fee

shahafn:
This reverts commit 5637d6d701eeb6ac65572d5379886ecc4aef7673.

### DIFF
--- a/packages/contracts/src/RelayHub.sol
+++ b/packages/contracts/src/RelayHub.sol
@@ -86,8 +86,6 @@ contract RelayHub is IRelayHub, Ownable, ERC165 {
 
     mapping(address => uint256) internal balances;
 
-    mapping(address => uint256) internal markedAsAbandoned;
-
     uint256 internal immutable creationBlock;
     uint256 internal deprecationTime = type(uint256).max;
 
@@ -104,11 +102,6 @@ contract RelayHub is IRelayHub, Ownable, ERC165 {
         batchGateway = _batchGateway;
         relayRegistrar = _relayRegistrar;
         setConfiguration(_config);
-    }
-
-    /// @inheritdoc IRelayHub
-    function getMarkedAsAbandoned(address relayManager) external override virtual view returns (uint256){
-        return markedAsAbandoned[relayManager];
     }
 
     /// @inheritdoc IRelayHub
@@ -597,25 +590,16 @@ contract RelayHub is IRelayHub, Ownable, ERC165 {
         stakeManager.penalizeRelayManager(relayManager, beneficiary, stakeInfo.stake);
     }
 
-    function markRelayAbandoned(address relayManager) external override onlyOwner {
+    function verifyRelayAbandoned(address relayManager) public view override {
         (IStakeManager.StakeInfo memory stakeInfo,) = stakeManager.getStakeInfo(relayManager);
-        require(stakeInfo.stake > 0, "relay manager not staked");
-        require(markedAsAbandoned[relayManager] == 0, "relay manager already abandoned");
-        markedAsAbandoned[relayManager] = block.timestamp;
+        require(stakeInfo.abandonedTime != 0 && stakeInfo.abandonedTime + config.abandonedRelayEscheatmentDelay < block.timestamp, "relay manager not abandoned yet");
     }
 
     function escheatAbandonedRelayBalance(address relayManager) external override onlyOwner {
-        require(markedAsAbandoned[relayManager] != 0 && markedAsAbandoned[relayManager] + config.abandonedRelayEscheatmentDelay < block.timestamp, "relay manager not abandoned yet");
+        verifyRelayAbandoned(relayManager);
         uint256 balance = balances[relayManager];
         balances[relayManager] = 0;
         balances[config.devAddress] = balances[config.devAddress].add(balance);
-    }
-
-    function revokeAbandonedStatus(address relayManager) external override {
-        (IStakeManager.StakeInfo memory stakeInfo,) = stakeManager.getStakeInfo(relayManager);
-        require(stakeInfo.owner == msg.sender, "only owner can revoke abandoned");
-        require(markedAsAbandoned[relayManager] != 0, "relay manager not abandoned");
-        markedAsAbandoned[relayManager] = 0;
     }
 
     /// @inheritdoc IRelayHub
--- a/packages/contracts/src/RelayHub.sol
+++ b/packages/contracts/src/RelayHub.sol
@@ -157,9 +157,11 @@ contract RelayHub is IRelayHub, Ownable, ERC165 {
     }
 
     /// @inheritdoc IRelayHub
-    function verifyCanRegister(address relayManager) external view override {
+    function onRelayServerRegistered(address relayManager) external override {
+        require(msg.sender == relayRegistrar, "caller is not relay registrar");
         verifyRelayManagerStaked(relayManager);
         require(workerCount[relayManager] > 0, "no relay workers");
+        stakeManager.updateRelayKeepaliveTime(relayManager);
     }
 
     /// @inheritdoc IRelayHub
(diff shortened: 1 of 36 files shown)


packages/contracts/src/StakeManager.sol
```diff
@@ -179,4 +212,42 @@ contract StakeManager is IStakeManager, Ownable {
...
+        info.withdrawTime = 0;
+        info.token.safeTransfer(devAddress, amount);
+        emit AbandonedRelayManagerStakeEscheated(relayManager, msg.sender, info.token, amount);
+    }
+
+    /// @inheritdoc IStakeManager
+    function updateRelayKeepaliveTime(address relayManager) external override {
+        StakeInfo storage info = stakes[relayManager];
+        bool isHubAuthorized = authorizedHubs[relayManager][msg.sender].removalTime == type(uint256).max;
+        bool isRelayOwner = info.owner == msg.sender;
```
drortirosh, line 247: this method is. called from relayHub, so msg.sender is the hub, and thus its not an owner

packages/contracts/src/interfaces/IRelayRegistrar.sol
```diff
@@ -36,11 +36,20 @@ interface IRelayRegistrar is IERC165 {
...
+        address indexed relayHub,
         uint256 baseRelayFee,
         uint256 pctRelayFee,
         bytes32[3] relayUrl
     );
 
+    /**
+     * @notice Emitted when a relay server and its details are removed from the RelayRegistrar.
+     */
+    event RelayServerRemoved(
```
drortirosh, line 48: unused?

packages/contracts/src/interfaces/IStakeManager.sol
```diff
@@ -73,6 +73,30 @@ interface IStakeManager {
         address indexed burnAddress
     );
 
+    /// @notice Emitted when a `burnAddress` is changed.
+    event DevAddressSet(
+        address indexed burnAddress
+    );
+
+    /// @notice
+    event RelayAbandoned(
```
drortirosh, line 82: "RelayServer" or "Relayer".

we use "Relay" (a verb) to refer to a "server" (noun)

packages/contracts/src/interfaces/IStakeManager.sol
```diff
@@ -179,6 +240,17 @@ interface IStakeManager {
      */
     function getBurnAddress() external view returns (address);
 
+    /**
+     * @notice Change the address that will receive the 'abandoned' stake.
```
drortirosh, line 244: ok. the doc about "abondoned" relays is on "devAddress".
maybe add a 
struct AbndonedRelay {
  devAddress
  chewingDelay
  abandonedDelay
}

(for one thing, it adds a single method to retrieve those otherwise unused params. for another, it shows the relation between them...)

packages/dev/test/AbandonedRelayFlow.test.ts
```diff
@@ -0,0 +1,160 @@
...
+    await stakeManager.authorizeHubByOwner(relayManager, relayHubInstance.address, { from: relayOwner })
+    await relayHubInstance.addRelayWorkers([relayWorker], { from: relayManager })
+    await relayRegistrar.registerRelayServer(relayHubInstance.address, baseRelayFee, pctRelayFee, splitRelayUrlForRegistrar(url), { from: relayManager })
+  })
+
+  it('should not allow to mark a relay server with a recent keepalive transaction as abandoned', async function () {
+    await expectRevert(stakeManager.markRelayAbandoned(relayManager), 'relay manager was alive recently')
+  })
+
+  it('should allow contract owner to set relay as abandoned on the StakeManager', async function () {
```
drortirosh, line 63: only "**stakemanager owner**" should

packages/contracts/src/interfaces/IRelayRegistrar.sol
```diff
@@ -36,11 +36,20 @@ interface IRelayRegistrar is IERC165 {
      */
     event RelayServerRegistered(
         address indexed relayManager,
+        address indexed relayHub,
```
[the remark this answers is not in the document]
forshtat: The opposite. All events start with a relay manager address, and we look up all events by relay manager address. When I tried to have RelayHub first, I would have to do `getPastEvents([undefined, relayManager])` in a dozen of places.

packages/contracts/src/StakeManager.sol
forshtat: There is an "OR" relationship for these two checks - `isHubAuthorized OR isRelayOwner`

drortirosh (commented):
also missing: the relayer doesn't handle the markAbandoned

packages/contracts/src/RelayHub.sol
```diff
@@ -588,6 +592,20 @@ contract RelayHub is IRelayHub, Ownable, ERC165 {
...
+    function isRelayEscheatable(address relayManager) public view override returns (bool){
+        return stakeManager.isRelayEscheatable(relayManager);
+    }
+
+    /// @inheritdoc IRelayHub
+    function escheatAbandonedRelayBalance(address relayManager) external override onlyOwner {
+        require(stakeManager.isRelayEscheatable(relayManager), "relay server not escheatable yet");
+        uint256 balance = balances[relayManager];
+        balances[relayManager] = 0;
+        balances[config.devAddress] = balances[config.devAddress].add(balance);
```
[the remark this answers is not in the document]
forshtat: But we have a lot of 'add' in our code, so I guess removing SafeMath should be a separate issue:
https://opengsn.atlassian.net/browse/OG-771

I will keep using `.add` here for consistency.

packages/contracts/src/interfaces/IRelayRegistrar.sol
forshtat: Removed.

packages/contracts/src/interfaces/IStakeManager.sol
```diff
@@ -73,6 +73,30 @@ interface IStakeManager {
         address indexed burnAddress
     );
 
+    /// @notice Emitted when a `burnAddress` is changed.
+    event DevAddressSet(
```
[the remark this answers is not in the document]
forshtat: Fixed.

packages/contracts/src/interfaces/IStakeManager.sol
forshtat: In general I would be opposed to creating structs of three fields just to show they are related. It is generally much easier to work with basic types than structures in solidity and Web3.js, but if you really think it will be better I will join them into a struct.

forshtat:
> also missing: the relayer doesn't handle the markAbandoned

JavaScript server changes tend to take a lot of time, so I would take it to a separate issue

packages/contracts/src/interfaces/IStakeManager.sol
```diff
@@ -73,13 +73,13 @@ interface IStakeManager {
         address indexed burnAddress
     );
 
-    /// @notice Emitted when a `burnAddress` is changed.
+    /// @notice Emitted when a `devAddress` is changed.
     event DevAddressSet(
         address indexed burnAddress
```
drortirosh, line 78: still burnt?

packages/contracts/src/interfaces/IStakeManager.sol
```diff
@@ -73,6 +73,30 @@ interface IStakeManager {
...
+    );
+
+    /// @notice Emitted if Relay Server is inactive for an `abandonmentDelay` and contract owner initiates its removal.
+    event RelayServerAbandoned(
+        address indexed relayManager,
+        uint256 abandonedTime
+    );
+
+    /// @notice
+    event RelayKeepalive(
```
drortirosh, line 88: This is also RelayerServer
actually, "Relayer" is shorter (thinking also about all the method below, such as relayAbandoned, etc)

packages/contracts/src/interfaces/IStakeManager.sol
drortirosh: a struct (especial of "storage") is just a scope of several items (an alternative to a `structname_FieldName` uglifying)
It also has the benefit of having single method to set, and single method to retrieve.
btw: I think "OrphanRelayer" is a better name than "Abandoned", but its just a suggestion.

packages/dev/test/AbandonedRelayFlow.test.ts
forshtat: Fixed.

packages/contracts/src/interfaces/IStakeManager.sol
forshtat: Sorry. Double-fixed.

packages/contracts/src/interfaces/IStakeManager.sol
forshtat: I prefer RelayServer. Fixed.

packages/contracts/src/interfaces/IStakeManager.sol
forshtat: Created a struct. 
Orphan is not a good name and does not convey what the intention is.

drortirosh (approved):

PR #792, opened 2022-04-05
Og 671 hardhat deployment script

drortirosh:


packages/dev/deploy/deploy.ts
```diff
@@ -0,0 +1,178 @@
...
+    const arbos = await ArbSys.arbOSVersion()
+
+    console.log('== Running on', chalk.yellowBright('Arbitrum'), 'arbOSVersion=', arbos)
+    isArbitrum = true
+  } catch (e) {
+  }
+
+  const chainId = parseInt(await getChainId())
+
+  const envname = Object.keys(environments).find(env => environments[env as EnvironmentsKeys].chainId === chainId)
```
forshtat, line 63: I guess this means that we need a chainId => environment mapping in our javascript code, don't you think so?

packages/dev/deploy/deploy.ts
```diff
@@ -0,0 +1,178 @@
...
+  console.log('loading env ( based on chainId', chainId, ')', envname ?? 'DefaultEnvironment')
```
forshtat, line 64: I would not recommend this. Someone who is trying to deploy the GSN on his own BobChain will get a deployment with a ganache parameters. Better to stop deployment I think, and have a parameter to skip this lookup for ganache.

packages/dev/deploy/deploy.ts
```diff
@@ -0,0 +1,178 @@
...
+    })
+    stakingTokenAddress = TestWEth.address
+    stakingTokenValue = stakingTokenValue ?? '0.1'
+  } else {
+    if (stakingTokenValue == null) {
+      console.error('must specify TOKEN_STAKE for staking token')
+      process.exit(1)
+    }
+    const token = new ethers.Contract(stakingTokenAddress,
+      ['function symbol() view returns (string)'], ethers.provider)
```
forshtat, line 86: I am not familiar with this syntax, what does this `symbol` mean here? You don't need the ABI to declare contract functions? This is nice.

packages/dev/deploy/deploy.ts
```diff
@@ -0,0 +1,178 @@
...
+    // verify token address is a valid token...
+    const symbol = await token.symbol().catch((e: any) => null)
+    if (symbol == null) throw new Error(`invalid token: ${stakingTokenAddress}`)
+    console.log('using token', symbol)
```
forshtat, line 90: `using token AAA at address 0xAAA`

packages/dev/deploy/deploy.ts
```diff
@@ -0,0 +1,178 @@
...
+  const hubConfig = env.relayHubConfiguration
+  let relayHub: DeployResult
+  let hubContractName: string
+  if (isArbitrum) {
+    console.log(`Using ${chalk.yellow('Arbitrum')} relayhub`)
+    hubContractName = 'ArbRelayHub'
+    relayHub = await deploy(hubContractName, {
+      from: deployer,
+      args: [
+        '0x0000000000000000000000000000000000000064', // ArbSys
```
forshtat, line 128: Please declare a constant in GSN commons

packages/dev/hardhat.config.ts
```diff
@@ -0,0 +1,83 @@
...
+
+function getInfuraNetwork (name: string): NetworkUserConfig {
+  return getNetwork(infuraUrl(name))
+}
+
+const CONTRACTS_LINK = 'contracts-link'
+
+if (!fs.existsSync(path.join(CONTRACTS_LINK, 'RelayHub.sol'))) {
+  console.log('== creating symlink', chalk.yellow(CONTRACTS_LINK), 'for contracts')
+  fs.symlinkSync('../contracts/solpp', CONTRACTS_LINK)
```
forshtat, line 39: Is this really necessary? I don't like this touching the file system from the config file. It can be executed by the engine dozens of times.

packages/dev/hardhat.config.ts
```diff
@@ -0,0 +1,83 @@
...
+
+    goerli: getInfuraNetwork('goarli'),
+    kovan: getInfuraNetwork('kovan'),
+    mainnet: getInfuraNetwork('mainnet')
+  },
+
+  etherscan: { apiKey: process.env.ETHERSCAN_API_KEY }
+}
+
+// support url-based network: either start with "http", or a single work (infura network name)
```
forshtat, line 72: "or a single word" ? also, we already have such a code in CLI package. Can we reuse it?

forshtat (changes requested):

PR #794, opened 2022-04-07
Stuff

forshtat:


packages/dev/deploy/deploy.ts
```diff
@@ -25,86 +25,93 @@ interface DeploymentConfig {
...
+}
 
-  let env: Environment
+function getMergedEnvironment (chainId: number): Environment {
   try {
-    const envname = Object.keys(environments).find(env => environments[env as EnvironmentsKeys].chainId === chainId)
-    console.log('loading env ( based on chainId', chainId, ')', envname ?? 'DefaultEnvironment')
-    const defaultEnv = envname != null ? environments[envname as EnvironmentsKeys] : defaultEnvironment
+    const env = getEnvironment(chainId)
+    if (env == null) {
```
drortirosh, line 77: this will fail for any environment not already defined in our code.
the idea that for "goerli/rinkeby/fuji" (or any other "simple" network that doesn't require heavy fiddling with params) we only need to define a deployment in deployment-config.ts, not in the "Environments"

PR #795, opened 2022-04-08
OG-669: separate Paymasters build task

forshtat:


.circleci/config.yml
```diff
@@ -52,6 +52,19 @@ jobs: # a collection of steps
       - store_test_results: # special step to upload test results for display in Test Summary
           path: /tmp/test-dev-results.log
 
+  test_paymasters:
```
drortirosh, line 55: that's half of the work.
- PMs should be removed from the the "tsc", "compile" tasks
- a separate "pm-compile"  task should be executed just before the "test-paymasters"
(pms take more time to compile than the rest, since they compile everything again AND external dependencies. so removing them drops the solidity/tsc time by half)

PR #802, opened 2022-04-13
OG-637: prevent repeated reverted transactions by relayer

forshtat:


packages/relay/src/RegistrationManager.ts
```diff
@@ -369,7 +369,7 @@ export class RegistrationManager {
   async addRelayWorker (currentBlockNumber: number, currentBlockTimestamp: number): Promise<PrefixedHexString> {
     // register on chain
     const addRelayWorkerMethod = await this.contractInteractor.getAddRelayWorkersMethod([this.workerAddress])
-    const gasLimit = await this.transactionManager.attemptEstimateGas('AddRelayWorkers', addRelayWorkerMethod, this.managerAddress)
+    const gasLimit = await this.transactionManager.estimateGas('AddRelayWorkers', addRelayWorkerMethod, this.managerAddress)
```
drortirosh, line 372: its ok, but redundant .
`sendTransaction` itself does `estimateGas()` in case it isn't given a `gasLimit`.
So what is the reason for explicitly estimating gas before calling it?

packages/relay/src/RegistrationManager.ts
```diff
@@ -412,12 +412,19 @@ export class RegistrationManager {
...
     const addWorkersPending = await this.txStoreManager.isActionPendingOrRecentlyMined(ServerAction.ADD_WORKER, currentBlockNumber, this.config.recentActionAvoidRepeatDistanceBlocks)
+    let skipGasEstimationForRegisterRelay = false
     if (!(workersAdded || addWorkersPending)) {
       const txHash = await this.addRelayWorker(currentBlockNumber, currentBlockTimestamp)
       transactions = transactions.concat(txHash)
+      skipGasEstimationForRegisterRelay = true
     }
     const registerMethod = await this.contractInteractor.getRegisterRelayMethod(this.hubAddress, this.config.baseRelayFee, this.config.pctRelayFee, this.config.url)
-    const gasLimit = await this.transactionManager.attemptEstimateGas('RegisterRelay', registerMethod, this.managerAddress)
+    let gasLimit: number
```
drortirosh, line 422: same is more relevant here.
just gasLeft=value if you want explicit value, or leave it as null to estimate..

packages/relay/src/RegistrationManager.ts
forshtat: Right. Fiexd.

packages/relay/src/RegistrationManager.ts
forshtat: Fixed.

drortirosh (approved):

packages/relay/src/RegistrationManager.ts
```diff
@@ -412,12 +410,17 @@ export class RegistrationManager {
     // add worker only if not already added
     const workersAdded = await this._isWorkerValid()
     const addWorkersPending = await this.txStoreManager.isActionPendingOrRecentlyMined(ServerAction.ADD_WORKER, currentBlockNumber, this.config.recentActionAvoidRepeatDistanceBlocks)
+    let skipGasEstimationForRegisterRelay = false
```
drortirosh, line 413: nitpic: you actually no longer need this boolean: just move the gasLimit initialization here...

packages/relay/src/TransactionManager.ts
```diff
@@ -160,7 +149,7 @@ data                     | ${transaction.data}
       gasLimit = await this.contractInteractor.estimateGas({
         from: txDetails.signer,
         to: txDetails.destination,
-        data: txDetails.method,
+        data: encodedCall,
```
drortirosh, line 152: good catch..

PR #803, opened 2022-04-13
OG-773: Make Server read "relayRegistrationMaximumAge" from the Relay…

forshtat:
…Registrar

### DIFF
--- a/packages/contracts/src/utils/RelayRegistrar.sol
+++ b/packages/contracts/src/utils/RelayRegistrar.sol
@@ -32,13 +32,21 @@ contract RelayRegistrar is IRelayRegistrar, ERC165 {
 
     uint256 private immutable creationBlock;
 
-    constructor() {
+    uint256 private relayRegistrationMaxAge;
+
+    constructor(uint256 _relayRegistrationMaxAge) {
         creationBlock = block.number;
+        relayRegistrationMaxAge = _relayRegistrationMaxAge;
     }
 
     /// @inheritdoc IRelayRegistrar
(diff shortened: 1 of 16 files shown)


packages/contracts/src/utils/RelayRegistrar.sol
```diff
@@ -32,15 +32,23 @@ contract RelayRegistrar is IRelayRegistrar, ERC165 {
 
     uint256 private immutable creationBlock;
 
-    constructor() {
+    uint256 private relayRegistrationMaxAge;
```
drortirosh, line 35: there is no setter. this means it is not configurable, and set hard in stone.
Since a RelayHub can't change a Registrar, it means we need to re-deploy a relayHub in order to change this value.

packages/contracts/src/utils/RelayRegistrar.sol
```diff
@@ -97,6 +111,21 @@ contract RelayRegistrar is IRelayRegistrar, ERC165 {
...
+    )
+    public
+    view
+    override
+    returns (
+        RelayInfo[] memory info
+    ) {
+        uint256 blockTimestamp = block.timestamp;
+        uint256 oldestBlockTimestamp = blockTimestamp >= relayRegistrationMaxAge ? blockTimestamp - relayRegistrationMaxAge : 0;
+        return readRelayInfosInRange(relayHub, 0, oldestBlockTimestamp, 100);
```
drortirosh, line 124: please make a constant, and set to a real large value (that is still good gas-wise, e.g. 1000)

PR #805, opened 2022-04-18
Og 784 fix provider for ethers

drortirosh:


packages/provider/src/RelayProvider.ts
```diff
@@ -213,6 +281,35 @@ export class RelayProvider implements HttpProvider, Web3ProviderBaseInterface {
...
 
+  /**
+   * convert relayRequestId (which is a "synthethic" transaction ID) into the actual transaction Id.
+   * This is done by parsing RelayHub event, and can only be done after mining.
+   * @param relayRequestId
+   * @return transactionId or marker:
+   * If the transaction is already mined, return a real transactionId
+   * If requestId doesn't appear in the cache, return TX_NOTSENTHERE
+   * If the transaction is no longer valid, return TX_NOTFOUND
+   * If the transaction can still be mined, returns TX_FUTURE
```
forshtat, line 292: What is the point of `TX_FUTURE`?

drortirosh: I duplicated the logic of extracting transactionId from requestId that was inside _createTransactionReceiptForRelayRequestID .
currently, it is used only by the new _ethGetTransactionByHash, but it was built so createXX can also use it.
(that difference doesn't matter for getTxByHash)

PR #807, opened 2022-05-09
fix "bug somewhere"

DenkevichOleg:
Fix "Due to a bug somewhere chainId and netId need to be identical." here https://docs.opengsn.org/javascript-client/tutorial.html#manual-tests

becouse https://github.com/trufflesuite/truffle/issues/3961 this issue already closed


drortirosh:
this is unrelated. 
the client code doesn't use hdwallet-provider, and the mentioned bug fix affect only "truffle migrate".

PR #809, opened 2022-05-11
OG-979: relayer-withdrawal custom target

drortirosh:
added optional --target
if not set, withdraw to owner address

### DIFF
--- a/packages/cli/src/CommandsLogic.ts
+++ b/packages/cli/src/CommandsLogic.ts
@@ -53,6 +53,7 @@ export interface WithdrawOptions {
   config: ServerConfigParams
   broadcast: boolean
   gasPrice?: BN
+  withdrawTarget?: string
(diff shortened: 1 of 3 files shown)


packages/cli/src/CommandsLogic.ts
```diff
@@ -456,7 +462,7 @@ export class CommandsLogic {
         if (balance.lt(options.withdrawAmount)) {
           throw new Error('Relay manager hub balance lower than withdrawal amount')
         }
-        const method = relayHub.contract.methods.withdraw(owner, options.withdrawAmount)
+        const method = relayHub.contract.methods.withdraw(options.withdrawAmount, withdrawTarget)
```
shahafn, line 465: You swapped target and amount..

forshtat (approved):

PR #810, opened 2022-05-12
OG-692 server penalization protection

shahafn:
Validate that the server doesn't make itself penalizable by reusing
workers on multiple hubs


packages/relay/src/ServerConfigParams.ts
```diff
@@ -21,6 +21,8 @@ export enum LoggingProviderMode {
   CHATTY
 }
 
+export const HUB_FILE = '/hub.lock'
```
drortirosh, line 24: nitpic: better put the "slash" when joining the string. otherwise, it looks like the file is at the root..

drortirosh (commented):
the real check is missing: 
check that worker's address is "clean" at registration: it shouldn't have neither balance nor transactionCount.
That check is something should not be override-able: There is absolutely no reason to addWorker if that address generated transaction.

shahafn:
> the real check is missing: check that worker's address is "clean" at registration: it shouldn't have neither balance nor transactionCount. That check is something should not be override-able: There is absolutely no reason to addWorker if that address generated transaction.

Right, but that check is insufficient to achieve protection prom penalization, so while adding it is OK, it's not enough. What I added covers that case as well. I can add another layer of protection right when adding the worker address, but again, it doesn't cover all cases. Should I?

packages/relay/src/ServerConfigParams.ts
shahafn: Ok

packages/relay/src/ServerConfigParams.ts
```diff
@@ -369,6 +371,20 @@ export function validateBalanceParams (config: ServerConfigParams): void {
   }
 }
 
+export async function validateHub (hubFile: string, hubAddress: string, managerAddress: string, workerAddress: string, contractInteractor: ContractInteractor): Promise<void> {
+  const managerNonce = await contractInteractor.getTransactionCount(managerAddress, 'pending')
```
drortirosh, line 375: we don't care about manager: it is perfectly ok to have the same manager (with nonce!=0) for multiple relayhubs (actually, it is even expected, as we want to reuse the stake)

packages/relay/src/ServerConfigParams.ts
```diff
@@ -369,6 +371,20 @@ export function validateBalanceParams (config: ServerConfigParams): void {
...
+  const workerNonce = await contractInteractor.getTransactionCount(workerAddress, 'pending')
```
drortirosh, line 376: when verifying that worker is "pristine", verify its BALANCE is also zero: the end of registration always put some eth into a worker, and that (along with transactionCount) is a great marker that it should not be re-used in another relayhub

packages/relay/src/ServerConfigParams.ts
shahafn: This check doesn't prevent the user from using the same manager with a different hub, it's being used to know if the hub lock file should be overwritten silently, or the user has to manually remove it since there has been activity on a different hub, and we currently do not have any way to know if that makes the relayer penalizable.

packages/relay/src/ServerConfigParams.ts
```diff
@@ -368,6 +370,22 @@ export function validateBalanceParams (config: ServerConfigParams): void {
...
+  if (!fs.existsSync(hubFile) || (managerNonce === 0 && workerNonce === 0)) {
+    fs.writeFileSync(hubFile, hubAddress)
+    console.log(`RelayHub lock file ${hubFile} created.\n`)
+  } else {
+    console.log(`Using existing lock file ${hubFile}.\n`)
+    const lockedHub = fs.readFileSync(hubFile, 'utf8')
+    if (lockedHub !== hubAddress) {
```
forshtat, line 382: Please use `isSameAddress` to avoid uppercase/lowercase. Also, there is no reason not to print both addresses as part of an error message, in case the error is in having an extra whitespace character in config or something.

packages/relay/src/ServerConfigParams.ts
```diff
@@ -368,6 +370,22 @@ export function validateBalanceParams (config: ServerConfigParams): void {
   }
 }
 
+export async function validateHub (hubFile: string, hubAddress: string, managerAddress: string, workerAddress: string, contractInteractor: ContractInteractor): Promise<void> {
+  const managerNonce = await contractInteractor.getTransactionCount(managerAddress, 'pending')
+  const workerNonce = await contractInteractor.getTransactionCount(workerAddress, 'pending')
+  if (!fs.existsSync(hubFile) || (managerNonce === 0 && workerNonce === 0)) {
```
forshtat, line 376: Actually, what is the added value of having `managerNonce === 0 && workerNonce === 0` here? This is a very-very edge case if you ask me. Let's maybe have a simple, binary rule:
IF (FILE_EXISTS) {
 HUB_ADDRESS_MUST_MATCH()
} ELSE {
 CREATE_FILE()
}

packages/relay/src/ServerConfigParams.ts
shahafn: Whitespace is not an issue when parsing the config. I can use `isSameAddress`, though it seems unnecessary to me... why would you add it lowercase, then change it to the exact same address but checksummed or vice versa?

packages/relay/src/ServerConfigParams.ts
shahafn: It might be the case where the dev got the hub wrong then fixed it, and if that happens after funding the manager, that would be annoying I guess. What's the downside?

packages/relay/src/ServerConfigParams.ts
drortirosh: I don't think so: 
calling addWorker() if the worker is not "pristine" is an error, regardless of how you got there. 
your code relies on the "relayhub" lock file. if the admin removes it, then its OK to create a penalizeable worker. 
I say that even if the admin remove the lock file, it is STILL forbidden to reuse this worker.

drortirosh (commented):
I miss in this PR the documentation of the flow (of initial setup, of correct relayhub update, of broken relayhub update)

note that regardless of the lock file, calling addWorker on a "tainted" worker account is ALWAYS forbidden.

packages/relay/src/ServerConfigParams.ts
forshtat: > Whitespace is not an issue when parsing the config. I can use `isSameAddress`, though it seems unnecessary to me... why would you add it lowercase, then change it to the exact same address but checksummed or vice versa?

Basically, because you can... I don't know, I prefer comparing addresses taking that into account.

packages/relay/src/ServerConfigParams.ts
forshtat: > It might be the case where the dev got the hub wrong then fixed it, and if that happens after funding the manager, that would be annoying I guess. What's the downside?

It is super edge case and handling it explicitly is unnecessary. I vote to remove this scenario. 
The downside is increased complexity. The upside is not even there as far as I can understand.

drortirosh: manager funding doesn't really matter: you can always withdraw, both deposit and account balance.
its the worker we need to be careful about.
- the simple scenario to protect: avoid re-using a worker on another hub.
   - This is simply protected by never addWorker if it has balance/nonce: it was probably used in another relay (since our server only send balance after addWorker)
- the complex scenario: worker was added to a hub, but before doing anything with it the owner switched hub.
   -   this is complex, since there is no "footprint" on the worker, so @shahaf tried to resolve it by remembering (off-chain) the old hub.
   - while I agree we need shahaf's method, its not good to rely on it, so I want to use the above simple protection.
   - the needed change: transfer funds to a worker immediately after addWorker succeeds. This way, there is almost no time a worker is added but with no on-chain footprint. This is not bullet-proof (hence the need of the offchain mechanism) - but it is very close

packages/relay/src/ServerConfigParams.ts
shahafn: I would agree if that caught all of these errors, but that's not the case, the worker might still be penalizeable. So I'm not sure about adding an extra check to cover only some errors, if the admin really wants to lose his money and remove internal files AND change the hub address in the config. That's really taking an extra step to sabotage yourself. 
@forshtat ?

forshtat (approved):

PR #812, opened 2022-05-12
OG-788: Blacklist for Relay Servers in client configuration

forshtat:


packages/common/src/ConfigResponse.ts
```diff
@@ -21,8 +21,14 @@ export interface ConfigEntry {
   gsnConfig: Partial<GSNConfig>
 }
 
+export interface BlacklistedRelayServerInfo {
```
drortirosh, line 24: I think it should be a `string[]`, not some compex structure, just like the preferred relays

forshtat: Don't you think we should differentiate between manager address and host? Just drop everything in one string array?

forshtat: Fixed.

packages/provider/src/KnownRelaysManager.ts
```diff
@@ -80,7 +80,16 @@ export class KnownRelaysManager {
 
     this.logger.info(`fetchRelaysAdded: found ${relayInfos.length} relays`)
 
-    const filteredRelayInfos = relayInfos.filter(this.relayFilter)
+    const blacklistFilteredRelayInfos = relayInfos.filter((info: RelayRegisteredEventInfo) => {
+      let host = ''
+      try {
+        host = new URL(info.relayUrl).host.toLowerCase()
+      } catch (e) {}
+      const isHostBlacklisted = this.config.blacklistedRelays.find(relay => host.includes(relay.toLowerCase())) != null
```
drortirosh, line 88: better match against relayUrl, and not just the "host" part.
that is, allow filtering by full url, including the "https" prefix
(e.g. copying a URL from the preferred relay into the blacklistedRelays should work)

not sure if the blacklist-by-address is useful, but it causes no harm either.

drortirosh:
```js
 const isHostBlacklisted = this.config.blacklistedRelays.find(relay => info.relayUrl.toLowerCase().includes(relay.toLowerCase())) != null
```

PR #818, opened 2022-05-17
OG-274: when sending a "reverted" transaction, a user is not informed.

forshtat:


packages/common/src/ContractInteractor.ts
```diff
@@ -535,24 +536,26 @@ export class ContractInteractor {
       this.logger.debug('relayCall res=' + res)
 
       // @ts-ignore
-      const decoded = abi.decodeParameters(['bool', 'bytes'], res)
+      const decoded = abi.decodeParameters(['bool', 'uint256', 'bytes'], res)
```
drortirosh, line 539: technically, an enum is `uint8`...

drortirosh (commented):
The PR does do what its title says:

- when sending reverted code, a client WILL be informed if it does "view" call (or estimateGas). no need for change.
- the only case it will "miss" a revert is if the view call (from:user, to:recipient) succeeds, but fails when sent through GSN (e.g. a call that checks tx.origin)
- still even in that case, this PR doesn't expose the error to the user - the revert is exposed as internal RelayClient api, not to the provider.
- I don't see a single test to check the claim you have in the title. You only change internal API.

drortirosh (commented):
Do you think its needed? 
While I do agree that returning the value in view mode is a good feature (not requiring real transaction), the reasoning is wrong: everything can (and should) be checked off-chain, using estimateGas.

forshtat:
> The PR does do what its title says:
> 
> * when sending reverted code, a client WILL be informed if it does "view" call (or estimateGas). no need for change.
> * the only case it will "miss" a revert is if the view call (from:user, to:recipient) succeeds, but fails when sent through GSN (e.g. a call that checks tx.origin)
> * still even in that case, this PR doesn't expose the error to the user - the revert is exposed as internal RelayClient api, not to the provider.
> * I don't see a single test to check the claim you have in the title. You only change internal API.

1. The client does not have to do a view call in our system. It will if the gas is not specified, but this is not the same as this transaction going through GSN. It is right that it should work the same 99% of the time, but I wanted to make sure we do not rely on that. Other than direct `tx.origin`/`msg.sender` being used somewhere, there may be additional state changes happening in the Paymaster, for example, or have some weird gas-related checks that may be broken with "view"/"estimateGas" calls. This change allows the DRY-RUN execution to present the recipient error to the caller.
2. The PR does expose the error. It is thrown as error for DRY-RUN execution.
3. Please explain how did you come to a conclusion that the test called `should report the recipient revert reason without requesting client signature` does not check the claim I made in this PR.

packages/common/src/ContractInteractor.ts
forshtat: In the decoded structure it will not matter.

packages/common/src/Constants.ts
```diff
@@ -42,3 +42,13 @@ export const erc165Interfaces = {
   relayHub: getERC165InterfaceID(relayHubAbi as any),
   stakeManager: getERC165InterfaceID(stakeManagerAbi as any)
 }
+
+export const RelayCallStatusCodes = {
```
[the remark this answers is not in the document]
forshtat: `BN` is what is returned to you by web3.js contracts. I can change it if it matters to you, but I think this is ok to use BN for on-chain values.

forshtat:
> Do you think its needed? While I do agree that returning the value in view mode is a good feature (not requiring real transaction), the reasoning is wrong: everything can (and should) be checked off-chain, using estimateGas.

I strongly believe this is very important that we do this.

drortirosh (approved):

drortirosh (approved):

drortirosh (approved):

PR #819, opened 2022-05-17
OG-805: deployer: extract "applyConfig" from actual "deploy"

drortirosh:


packages/deployer/src/deployUtils.ts
```diff
@@ -151,3 +160,99 @@ export async function printRelayInfo (hre: HardhatRuntimeEnvironment): Promise<v
   console.log(chalk.white('Relayer register:'))
   console.log(chalk.grey(`gsn relayer-register -m $SECRET --network ${network.url} --relayUrl https://${hre.hardhatArguments.network}.v3.opengsn.org/v3 --token ${stakingTokenAddress} --stake ${stakingTokenValue} --wrap`))
 }
+
+export async function getEnv (hre: HardhatRuntimeEnvironment): Promise<{ deployer: string, deployments: DeploymentsExtension, env: Environment }> {
```
forshtat, line 164: We have `getEnvironment` global function. Please rename to `getDeploymentEnvironment`.

packages/deployer/src/applyConfig.ts
```diff
@@ -0,0 +1,8 @@
+import { applyDeploymentConfig } from './deployUtils'
```
forshtat, line 1: Don't you want to export these scripts as CLI commands? Why nort?

packages/deployer/src/deployUtils.ts
```diff
@@ -151,3 +160,99 @@ export async function printRelayInfo (hre: HardhatRuntimeEnvironment): Promise<v
...
+
+  return {
+    deployer,
+    deployments,
+    env
+  }
+}
+
+// check if token's minimum stake is correct. return null if no need to change.
+async function updateTokenStakeOrNull (hub: Contract, tokenAddr: string, configMinimumStake: string): Promise<{ token: string, minimumStake: string } | null> {
```
forshtat, line 180: This is not actor, this is getter. `getTokenMinimumStake`?

packages/deployer/src/applyConfig.ts
drortirosh: this is an ethers/hardhat script, like deploy. 
it also depends on the "deployments" created by deploy under this folder.

forshtat (approved):

PR #820, opened 2022-05-23
OG-68: Specifying preferred relay that is not registered should log error and continue

forshtat:


packages/provider/src/RelaySelectionManager.ts
```diff
@@ -76,8 +76,8 @@ export class RelaySelectionManager {
             relayInfo
           }
         } else {
-          // TODO: do not throw! The preferred relay may be removed since.
-          throw new Error('Could not find register event for the winning preferred relay')
+          this.logger.error('Could not find registration info in the RelayRegistrar for the selected preferred relay')
```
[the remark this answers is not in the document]
forshtat: Well, I am not re-writing the old code that was based on wrong assumptions now. I think this is a valid question but is only relevant for a re-write. As far as I can tell this won't make an RPC call, but will just find an entry in the known relays list.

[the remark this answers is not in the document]
forshtat: This code is only executed for "preferred" relays because the object does not have the details and need to find them. Normal relay will have all details and not get to this branch in the first place.

packages/provider/src/RelaySelectionManager.ts
```diff
@@ -76,8 +76,8 @@ export class RelaySelectionManager {
...
+          return undefined
```
drortirosh, line 80: missing a test to show the outcome of this `return undefined`

forshtat: > missing a test to show the outcome of this `return undefined`

It is a bit hard to observe anything happening here other than adding a `sinon.spy` to the logger and then check the error is logged. This code is explicitly about ignoring an error.
Do you want a spy test or do you have a better idea?

drortirosh: The method didn't return `undefined` before, and now it does.
I'm not sure how it behaves, and thus I think it should be tested somehow.

You can do a manual test, e.g. by doing `yarn link` into ctf and test it.

forshtat: > The method didn't return `undefined` before, and now it does. I'm not sure how it behaves, and thus I think it should be tested somehow.
> 
> You can do a manual test, e.g. by doing `yarn link` into ctf and test it.

This method is, however, explicitly marked as `Promise<RelayInfo | undefined>`, with `undefined` meaning "no more [preferred] relays", as handled in `selectNextRelay`

packages/provider/src/RelaySelectionManager.ts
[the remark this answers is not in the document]
shahafn: It's unclear in the code why this part handles preferred relays only

drortirosh (approved):

PR #825, opened 2022-05-25
OG-810 unexpected paymaster balance too low

drortirosh:


forshtat:
Note that this probably should not be merged. It is legitimate for the protocol to demand Paymaster to have enough balance to cover `maxViewableGasLimit`, which is also client-configurable.

PR #823, opened 2022-05-25
OG-801 getBlock() fix

shahafn:

PR #828, opened 2022-05-27
OG-812 Using RelayRegistrar in server

shahafn:
Also removing coldRestartLogsFromBlock param

### DIFF
--- a/packages/common/src/ContractInteractor.ts
+++ b/packages/common/src/ContractInteractor.ts
@@ -1155,7 +1155,7 @@ calculateTransactionMaxPossibleGas: result: ${result}
     return await this.relayRegistrar.getRelayRegistrationMaxAge()
   }
 
-  async getRelayInfo (relayHubAddress: string, relayManagerAddress: string): Promise<{
+  async getRelayInfo (relayManagerAddress: string): Promise<{
     lastSeenBlockNumber: BN
     lastSeenTimestamp: BN
     firstSeenBlockNumber: BN
@@ -1165,7 +1165,7 @@ calculateTransactionMaxPossibleGas: result: ${result}
     urlParts: string[]
     relayManager: string
   }> {
-    return await this.relayRegistrar.getRelayInfo(relayHubAddress, relayManagerAddress)
+    return await this.relayRegistrar.getRelayInfo(this.relayHubInstance.address, relayManagerAddress)
   }
 
   /**
--- a/packages/common/src/ContractInteractor.ts
+++ b/packages/common/src/ContractInteractor.ts
@@ -54,13 +54,10 @@ import Common from '@ethereumjs/common'
 import { GSNContractsDeployment } from './GSNContractsDeployment'
 import {
   ActiveManagerEvents,
-  HubUnauthorized,
   RelayRegisteredEventInfo,
   RelayServerRegistered,
   RelayWorkersAdded,
-  StakeInfo,
-  StakePenalized,
-  StakeUnlocked
+  StakeInfo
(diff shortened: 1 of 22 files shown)


packages/common/src/ContractInteractor.ts
```diff
@@ -1158,6 +1158,19 @@ calculateTransactionMaxPossibleGas: result: ${result}
     return await this.relayRegistrar.getRelayRegistrationMaxAge()
   }
 
+  async getRelayInfo (relayHubAddress: string, relayManagerAddress: string): Promise<{
```
drortirosh, line 1161: why would you need the params? don't you have them in the ContractInteractor ?

packages/dev/test/RelayServerRequestsProfiling.test.ts
```diff
@@ -11,7 +11,7 @@ import { defaultEnvironment } from '@opengsn/common'
 
 contract('RelayServerRequestsProfiling', function (accounts) {
   const refreshStateTimeoutBlocks = 2
-  const callsPerStateRefresh = 13
+  const callsPerStateRefresh = 14
```
drortirosh, line 14: we talked about decreasing this method count. why it keep rising ?

packages/relay/src/RelayServer.ts
```diff
@@ -780,31 +786,14 @@ latestBlock timestamp   | ${latestBlock.timestamp}
...
-    }
-  }
-
   async _queryLatestActiveEvent (): Promise<EventData | undefined> {
-    const events: EventData[] = await this.contractInteractor.getPastEventsForHub([address2topic(this.managerAddress)], {
-      fromBlock: this.config.coldRestartLogsFromBlock
+    const hubEvents: EventData[] = await this.contractInteractor.getPastEventsForHub([address2topic(this.managerAddress)], {
+      fromBlock: this.lastScannedBlock + 1
+    })
+    const registrarEvents: EventData[] = await this.contractInteractor.getPastEventsForRegistrar([address2topic(this.managerAddress)], {
```
drortirosh, line 793: why would you care about "registrar" events?

shahafn: Because that was a bug, since it includes registration events, which registrationManager in the server still uses.

packages/dev/test/RelayServerRequestsProfiling.test.ts
shahafn: Because we now read two different values from registrar, relayRegistrationMaxAge (which is configurable so can't be cached indefinitely) and relayInfo. I decreased it from 14 to 13 just a few commits  ago at #823 so it technically didn't increase for a long time.

packages/common/src/ContractInteractor.ts
drortirosh: ?

packages/dev/test/RelayServerRequestsProfiling.test.ts
drortirosh: the point is that we have an event that happens once every 180 days (or at a rate we know, which is supposed to be very large), but we check it on every block...

packages/common/src/ContractInteractor.ts
shahafn: We only have hub in ContractInteractor, not manager address.

drortirosh: right the manager is per-relay. the relayHubAddress is redundant.

drortirosh (approved):

PR #829, opened 2022-05-28
OG-252: Improve client assurance when using maxNonceGap

forshtat:
1. Store raw signed serialized transactions in Relay Server database
2. Return signed transactions inside requested nonce gap to the client
3. Clients validate all transactions inside nonce gap to be well priced
and not penalizeable
4. Return detailed response of what is wrong with each transaction
instead of throwing on the first encountered problem


packages/provider/src/RelayedTransactionValidator.ts
```diff
@@ -23,77 +63,130 @@ export class RelayedTransactionValidator {
...
-      }
-
-      this.logger.info('validateRelayResponse - valid transaction response')
-      return true
+    if (transaction instanceof Transaction) {
+      isLegacyGasPriceValid = transaction.gasPrice.gte(toBN(request.relayRequest.relayData.maxFeePerGas))
+      // throw new Error(`Relay Server signed gas price too low (${transaction.gasPrice}). Requested transaction with gas price at least ${request.relayRequest.relayData.maxPriorityFeePerGas}`)
+    } else if (transaction instanceof FeeMarketEIP1559Transaction) {
+      isFeeMarket1559Transaction = true
+      isMaxPriorityFeePerGasValid = transaction.maxPriorityFeePerGas.gte(toBN(request.relayRequest.relayData.maxPriorityFeePerGas))
```
drortirosh, line 161: sshould compare to latest ping result's maxPriorityFee.
(if relayer returned 5, and I decided on prioFee of 10, I can't expect anyone else to have that prio, only what the server's minimum is)

packages/relay/src/TxStoreManager.ts
```diff
@@ -78,35 +78,31 @@ export class TxStoreManager {
     return await this.txstore.asyncFindOne({ txId: txId.toLowerCase() })
   }
 
-  async getTxsUntilNonce (signer: PrefixedHexString, nonce: number): Promise<StoredTransaction[]> {
-    return await this.txstore.asyncFind({
+  async getTxsInNonceRange (signer: PrefixedHexString, fromNonce: number, toNonce: number): Promise<StoredTransaction[]> {
```
drortirosh, line 81: toNonce can be optional: setting $gte to "undefined" is supposed to mean not checking it.

drortirosh (dismissed):

drortirosh (approved):

PR #830, opened 2022-05-29
OG-815 Calling RelayProvider.send(...) without gas fees params

shahafn:


packages/cli/src/GsnTestEnvironment.ts
```diff
@@ -186,7 +186,6 @@ class GsnTestEnvironmentClass {
       reputationManager
     }
     const relayServerParams: Partial<ServerConfigParams> = {
-      coldRestartLogsFromBlock: 1,
```
drortirosh, line 189: unrelated. this is only because you make a PR out of another PR's branch. 
I suggest to avoid it, as it most cases, you'll need to update code and not assume both PRs will get in, and thus it will also make reviewing it easier.

shahafn: Please review the former PR first. After that, I will rebase it and it will be fine.

drortirosh: I understand. but PRs that are unrelated are better be based on master, for easier review. they will need "update to master" anyway..

drortirosh (approved):

PR #836, opened 2022-06-07
OG-830 run-with-gsn

drortirosh:


packages/cli/src/commands/gsn-start.ts
```diff
@@ -1,19 +1,35 @@
...
-    showDeployment(env.contractsDeployment, 'GSN started')
 
-    console.log(`Relay is active, URL = ${env.relayUrl} . Press Ctrl-C to abort`)
+    if (commander.run != null) {
+      console.log('running command: ', commander.run)
+      try {
+        execSync(commander.run, { stdio: 'inherit' })
+        process.exit(0)
+      } catch (e: any) {
+        process.exit(1)
```
forshtat, line 27: won't this consume the exception instead of printing a stack trace?

forshtat (approved):

packages/cli/src/commands/gsn-start.ts
drortirosh: yes, it does. but the exception is "execSync failed", or something like that. 
The external command itself inherits both stdio and stdin, so you'd get as a a response something like this
`/bin/sh: asdasd: command not found`
or whatever error the command emits.
If the command you're running is completely silent, then yes, it would not display any error (e.g. `run-with-gsn "exit 1"` just dumps the "gsn start" output, and exits with error code 1)

PR #837, opened 2022-06-07
OG-814: Make ERC-1271 Forwarder standard for v3

forshtat:


forshtat:
Few issues with ERC-1271:
1. It does not define the signature type and we are reliant on ERC-721 signatures. Their reference implementation uses ERC-191 signed data format. This Forwarder will not be compatible with it.
2. It allows 'from' to be a contract, breaking an assumption that value can be sent back to signer. This PR bans message value for ERC-1271 signers only.
3. We call the signer's `isValidSignature` without limiting its gas. Need to think if this can hurt us.

packages/contracts/src/forwarder/Forwarder.sol
```diff
@@ -64,7 +67,6 @@ contract Forwarder is IForwarder, ERC165 {
         _verifySig(req, domainSeparator, requestTypeHash, suffixData, sig);
     }
 
-    /// @inheritdoc IForwarder
```
forshtat, line 67: undo

PR #842, opened 2022-06-14
OG-633: Add Server Config dumping method to Relay Server REST API

forshtat:


packages/relay/src/RelayServer.ts
```diff
@@ -159,6 +159,10 @@ export class RelayServer extends EventEmitter {
     return statsResponse
   }
 
+  configDumpHandler (): ServerConfigParams {
+    return this.config
```
drortirosh, line 163: some config params are private (e.g. nodeUrl might contain private server key)

drortirosh, line 163: need to filter out some params (e.g. nodeUrl, which contains private API key)

PR #847, opened 2022-06-25
OG-527: export GSN types better

forshtat:


packages/provider/src/GSNConfigurator.ts
```diff
@@ -53,13 +55,6 @@ export const defaultGsnConfig: GSNConfig = {
   performDryRunViewRelayCall: true
 }
 
-export interface LoggerConfiguration {
```
drortirosh, line 56: Not strictly related to this refactor. ..
Though I agree its not really needed.

drortirosh (approved):

PR #849, opened 2022-06-26
(broken branch) OG-819 version from package

drortirosh:


forshtat:
I think something like this did not work for us in WebPack or some other browser environment and that is the reason we have originally removed it and went with a hardcoded string. 
Please confirm that it works in a browser before we can merge this. Thanks.

drortirosh:
you're right, but we had "indentation problem" back then.
we need that the source file Version.ts and its compiled version (Version.js) are the same-depth  to reference package.json.
they used to be in different depths. right now, they are named "./src/Version.ts" and "./dist/Version.js", so its OK.

anyway, I validated it by compiling ctf-react with it (using local verdaccio), and it properly logs the right version .

drortirosh:
wrong branch name.

drortirosh:
wrong branch name. recreated as #851

PR #856, opened 2022-07-04
OG-858: RelayServer output print both Config and StakeManger owner address

forshtat:


packages/relay/src/RegistrationManager.ts
```diff
@@ -559,7 +559,8 @@ Hub authorized | ${boolString(this.isHubAuthorized)}
 Stake locked   | ${boolString(this.isStakeLocked)}
 Manager        | ${this.managerAddress}
 Worker         | ${this.workerAddress}
-Owner          | ${this.ownerAddress ?? chalk.red('unknown')}
+Stake Owner    | ${this.ownerAddress ?? chalk.red('not set')}
```
drortirosh, line 562: empty stakeOwner is "yellow" (its a transient state not error during setup, and once a balance is put in, it will be fixed automatically)

drortirosh (approved):

PR #857, opened 2022-07-04
OG-859: relayer doesn't complete registration without restart

forshtat:
This is not a good solution but this will force the same re-check that
is done in the 'init' to be done if the server wants to register itself
but is prevented from doing so by stake or balance checks not satisfied.


packages/relay/src/RegistrationManager.ts
```diff
@@ -390,6 +390,8 @@ export class RegistrationManager {
       stakeOnHubStatus.isStaked
     if (!allPrerequisitesOk) {
       this.logger.debug('will not actually attempt registration - prerequisites not satisfied')
+      await this.refreshStake()
```
drortirosh, line 393: amazing. a 2-line bug fix... without any validation test...

drortirosh (approved):

PR #858, opened 2022-07-04
OG-743: Remove all residual event-based relay lookup code

forshtat:


packages/relay/src/RegistrationManager.ts
```diff
@@ -228,6 +202,7 @@ export class RegistrationManager {
           break
       }
     }
+    await this.refreshRegistrarRelayInfo()
```
drortirosh, line 205: why is it called each time? better to trigger it only on event (alas, the events handled here are from the relayhub, not registrar...)

forshtat: The event that would be a good trigger for it is actually 'RelayRegistered' that is emitted on a Registrar contract. We don't query these events yet, so adding this trigger will add exactly 1 RPC call as well, and a maximum of 2 if the event actually happened.

drortirosh (approved):

PR #860, opened 2022-07-07
OG-368: Logger should be extracted to a separate package

forshtat:


packages/provider/src/RelayClient.ts
```diff
@@ -110,8 +110,7 @@ export class RelayClient {
       throw new Error('Sorry, but the constructor parameters of the RelayClient class have changed. See "GSNUnresolvedConstructorInput" interface for details.')
     }
     this.rawConstructorInput = rawConstructorInput
-    this.logger = rawConstructorInput.overrideDependencies?.logger ??
-      createClientLogger(rawConstructorInput.config?.loggerConfiguration ?? defaultLoggerConfiguration)
+    this.logger = rawConstructorInput.overrideDependencies?.logger ?? new LoggerEmpty()
```
drortirosh, line 113: client is better off to use `console` rather than `new LoggerEmpty()`. this way, you get debugger tool to get into the source line  where "console.log" was used

PR #861, opened 2022-07-12
OG-778: relayer should call authorizeByManager to update relayHub

forshtat:


packages/relay/src/RegistrationManager.ts
```diff
@@ -321,22 +332,41 @@ export class RegistrationManager {
...
+  async refreshStake (currentBlockNumber: number, currentBlockHash: string, currentBlockTimestamp: number): Promise<PrefixedHexString[]> {
+    const transactionHashes: string[] = []
     if (this.stakeRequired == null) {
       throw new Error('not initialized')
     }
     const stakeInfo = await this.contractInteractor.getStakeInfo(this.managerAddress)
     const stakedOnHubStatus = await this.contractInteractor.isRelayManagerStakedOnHub(this.managerAddress)
     if (stakedOnHubStatus.isStaked) {
       this.isHubAuthorized = true
+    } else if (stakedOnHubStatus.errorMessage === 'this hub is not authorized by SM') {
```
drortirosh, line 344: are we sure we receive the revert string on all node types ? (esp. geth, but not only)
usually, we use revert strings in tests and logs. here we have actual functionality based on it.

drortirosh (dismissed):

drortirosh (approved):

PR #867, opened 2022-07-14
OG-820: New token paymaster

shahafn:


packages/paymasters/src/PermitPaymasterUtils.ts
```diff
@@ -30,15 +32,25 @@ export const UNISWAP_V3_QUOTER_CONTRACT_ADDRESS = '0xb27308f9F90D607463bb33eA1Be
 export const SWAP_ROUTER_CONTRACT_ADDRESS = '0xE592427A0AEce92De3Edee1F18E0157C05861564'
 export const GSN_FORWARDER_CONTRACT_ADDRESS = '0xAa3E82b4c4093b4bA13Cb5714382C99ADBf750cA'
 export const UNISWAP_V3_DAI_WETH_POOL_CONTRACT_ADDRESS = '0xC2e9F25Be6257c210d7Adf0D4Cd6E3E881ba25f8'
+export const UNISWAP_V3_DAI_WETH_2_POOL_CONTRACT_ADDRESS = '0x60594a405d53811d3BC4766596EFD80fd545A270'
```
drortirosh, line 35: it is very bad to hard-code contract addresses (and so many of them) into the code. this works for exactly one network..

shahafn: These are used for tests only

package.json
```diff
@@ -42,11 +42,11 @@
   },
   "scripts": {
     "test": "yarn preprocess && yarn lerna-run-test-with-hardhat",
-    "ganache": "ganache-cli --chainId 1337 --hardfork 'istanbul' --gasLimit 120000000 --defaultBalanceEther 1000 --deterministic --keepAliveTimeout 2147483647",
-    "ganache-fork-mainnet": "ganache-cli --fork https://mainnet.infura.io/v3/f40be2b1a3914db682491dc62a19ad43 --port 8544 --chainId 1 --l 10000000  --unlock 0x47ac0fb4f2d84898e4d9e7b4dab3c24507a6d503",
+    "hardhat-fork-mainnet":  "hardhat node --port 8544 --fork https://mainnet.infura.io/v3/f40be2b1a3914db682491dc62a19ad43",
```
drortirosh, line 45: don't you need CHAINID=1 here?

packages/common/src/ConfigResponse.ts
```diff
@@ -105,6 +106,16 @@ export interface GSNConfig {
    */
   paymasterAddress?: Address
 
+  /**
```
drortirosh, line 109: I really dislike the fact we need to add here token-specific values.
it means our extension mechanism to add paymasters is broken. 
see the suggestion here: https://hackmd.io/vBD0nJKZRda3Ns-5brQ4qg

packages/paymasters/contracts/PermitERC20UniswapV3Paymaster.sol
```diff
@@ -15,71 +16,67 @@ import "@opengsn/contracts/src/utils/GsnUtils.sol";
...
 
+    using SafeERC20 for IERC20;
+
     event Received(address indexed sender, uint256 eth);
-    event TokensCharged(uint256 gasUseWithoutPost, uint256 gasJustPost, uint256 ethActualCharge);
+    event FundingNeeded();
+    event TokensCharged(uint256 gasUseWithoutPost, uint256 gasJustPost, uint256 tokenActualCharge, uint256 ethActualCharge);
 
-    IChainlinkOracle public priceFeed;
+    mapping(IERC20Metadata => IChainlinkOracle) public priceFeeds;
```
drortirosh, line 33: its bad to have multiple mappings with same key.
seems better to have a single mapping to a struct (better both in code readability and performance.)
along the way, it allows packing values together: (mapping address=>uint24) takes full uint256 for each member, but struct { uint24, address, } are packed into a single storage cell (and the mapping key is indexed once)

packages/paymasters/contracts/PermitERC20UniswapV3Paymaster.sol
```diff
@@ -91,25 +88,82 @@ contract PermitERC20UniswapV3Paymaster is BasePaymaster, ERC2771Recipient {
...
+        IERC20Metadata[] memory _tokens,
+        IChainlinkOracle[] memory _priceFeeds,
+        string[] memory _permitMethodSignatures,
+        uint24[] memory _poolFees) public onlyOwner {
+        tokens = _tokens;
+        // allow uniswap to transfer from paymaster balance
+        for (uint256 i = 0; i < tokens.length; i++) {
+            IERC20Metadata token = tokens[i];
+            token.approve(address(uniswap), type(uint256).max);
+            priceDivisors[token] = 10 ** uint256(_priceFeeds[i].decimals() + token.decimals());
```
drortirosh, line 121: sum of decimals? I thought it should be diff (minus)

packages/paymasters/contracts/PermitERC20UniswapV3Paymaster.sol
```diff
@@ -120,22 +174,29 @@ contract PermitERC20UniswapV3Paymaster is BasePaymaster, ERC2771Recipient {
...
+                "wrong \"permit\" method sig");
             // execute permit method for this token
             // solhint-disable-next-line avoid-low-level-calls
-            (bool success, bytes memory ret) = address(token).call(relayRequest.relayData.paymasterData);
+            (bool success, bytes memory ret) = address(token).call(relayRequest.relayData.paymasterData[:relayRequest.relayData.paymasterData.length - 20]);
             require(success, string(abi.encodePacked("permit call reverted:", string(ret))));
         }
-        uint256 tokenPreCharge = _calculatePreCharge(relayRequest, maxPossibleGas);
+
+        uint256 priceQuote = uint256(priceFeeds[token].latestAnswer());
```
drortirosh, line 194: why bother with "priceDivisor" (both here and in postOp) ?
Seems like we can align "priceQuote" to 18 decimals here, and then the "convert" functions don't need to handle the divisor anymore.

packages/paymasters/contracts/PermitERC20UniswapV3Paymaster.sol
```diff
@@ -120,22 +174,29 @@ contract PermitERC20UniswapV3Paymaster is BasePaymaster, ERC2771Recipient {
...
+            require(relayRequest.relayData.paymasterData.length >= 24, "must contain \"permit\" and token");
             require(
-                permitMethodSignature == GsnUtils.getMethodSig(relayRequest.relayData.paymasterData),
-                "paymasterData: wrong method sig");
+                permitMethodSignatures[token] == GsnUtils.getMethodSig(relayRequest.relayData.paymasterData),
+                "wrong \"permit\" method sig");
             // execute permit method for this token
             // solhint-disable-next-line avoid-low-level-calls
-            (bool success, bytes memory ret) = address(token).call(relayRequest.relayData.paymasterData);
+            (bool success, bytes memory ret) = address(token).call(relayRequest.relayData.paymasterData[:relayRequest.relayData.paymasterData.length - 20]);
```
drortirosh, line 190: your paymasterData format is very strange:
either "permit" followed by token, 
or only "token".
better have fixed offsets (e.g. token at offset 0, permit starts at offset 20)

packages/paymasters/contracts/PermitERC20UniswapV3Paymaster.sol
```diff
@@ -147,33 +208,110 @@ contract PermitERC20UniswapV3Paymaster is BasePaymaster, ERC2771Recipient {
...
+            return;
+        }
+        uint256 depositAmount = targetHubBalance - hubBalance + ethActualCharge;
+        _refillHubDeposit(depositAmount);
+    }
+
+    function _refillHubDeposit(uint256 depositAmount) private {
+        uint256 balance = address(this).balance;
+        uint256 amountSwapped = 0;
+        if (balance < depositAmount) {
```
drortirosh, line 234: the "if" is not needed: we enter the loop when we now the balance is too low, and the "for" performs exactly this if check..

packages/paymasters/contracts/PermitERC20UniswapV3Paymaster.sol
```diff
@@ -147,33 +208,110 @@ contract PermitERC20UniswapV3Paymaster is BasePaymaster, ERC2771Recipient {
...
+        return charge * (100 + paymasterFee) / 100;
     }
 
     // as this Paymaster already has a permission from a user to operate the tokens on behalf of the gasless account,
     // it makes this same Paymaster a great recipient of a transaction if its only action is a pure token transfer
-    function transferToken(address target, uint256 value) external {
+    function transferToken(IERC20Metadata token, address target, uint256 value) external {
         require(msg.sender == getTrustedForwarder(), "must be a meta-tx");
-        require(token.transferFrom(_msgSender(), target, value), "transferToken failed");
+        IERC20(token).safeTransferFrom(_msgSender(), target, value);
```
drortirosh, line 282: afaik, this casting is not needed: IERC20Metadata inherits from IERC20

packages/common/src/ConfigResponse.ts
shahafn: Again, this is because we don't currently have a way to update clients code if we add/remove a token. This is not the case for all paymasters as most are written by the dapp devs who can also update the client code. This is exactly the mechanism that we added to dynamically update the client config.

packages/paymasters/contracts/PermitERC20UniswapV3Paymaster.sol
shahafn: Yes, I added them one by one and forgot about it. I'll pack them to a struct

packages/paymasters/contracts/PermitERC20UniswapV3Paymaster.sol
shahafn: decimals is the exponent, not of 10**exponent, so a sum it is.

packages/paymasters/contracts/PermitERC20UniswapV3Paymaster.sol
shahafn: Because for some tokens it's not 18 and I wanted to keep it compatible

packages/paymasters/contracts/PermitERC20UniswapV3Paymaster.sol
shahafn: You baiscally tell me to swap their position. I can do that, but what do we gain from it?

packages/paymasters/src/WrapContract.ts
```diff
@@ -0,0 +1,34 @@
...
+export async function wrapContract (
+  contract: Contract,
+  config: Partial<TokenPaymasterConfig>,
+  overrideDependencies?: Partial<GSNDependencies>
+): Promise<Contract> {
+  const signer = await wrapSigner(contract.signer, config, overrideDependencies)
+  return contract.connect(signer)
+}
+
+export async function wrapSigner (
```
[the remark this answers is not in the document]
shahafn: It leaves me little choice with our packaging the way they are. Besides, this paymaster is a little different by design.

packages/paymasters/contracts/PermitERC20UniswapV3Paymaster.sol
```diff
@@ -91,25 +96,88 @@ contract PermitERC20UniswapV3Paymaster is BasePaymaster, ERC2771Recipient {
...
+        uint24[] memory _poolFees) public onlyOwner {
+        tokens = _tokens;
+        // allow uniswap to transfer from paymaster balance
+        for (uint256 i = 0; i < tokens.length; i++) {
+            TokenSwapData memory data;
+            IERC20Metadata token = tokens[i];
+            token.approve(address(uniswap), type(uint256).max);
+            data.priceDivisor = 10 ** uint256(_priceFeeds[i].decimals() + token.decimals());
+            data.priceFeed = _priceFeeds[i];
+            data.permitMethodSignature = bytes4(keccak256(bytes(_permitMethodSignatures[i])));
```
forshtat, line 132: I would vote for accepting `bytes4`. I don't see any benefit in keccak-ing on-chain.

packages/paymasters/contracts/PermitERC20UniswapV3Paymaster.sol
```diff
@@ -91,25 +96,88 @@ contract PermitERC20UniswapV3Paymaster is BasePaymaster, ERC2771Recipient {
...
+    function _weiToToken(uint256 amount, uint256 divisor, uint256 quote) internal pure returns (uint256) {
+        return amount * divisor / quote / 1e18;
+    }
     // solhint-disable-next-line no-empty-blocks
     function _verifyPaymasterData(GsnTypes.RelayRequest calldata relayRequest) internal virtual override view {}
 
-    // solhint-disable-next-line no-empty-blocks
-    function _verifyApprovalData(bytes calldata approvalData) internal virtual override view {}
+    function isTokenSupported(IERC20Metadata token) public view returns (bool) {
+        return tokensSwapData[token].priceDivisor != 0;
```
forshtat, line 179: This means it is impossible to disable any token once it has been added, I think.

packages/paymasters/contracts/PermitERC20UniswapV3Paymaster.sol
```diff
@@ -119,23 +187,33 @@ contract PermitERC20UniswapV3Paymaster is BasePaymaster, ERC2771Recipient {
...
+        IERC20Metadata token = _getTokenFromPaymasterData(relayRequest.relayData.paymasterData);
+        require(isTokenSupported(token),"unsupported token");
+        TokenSwapData memory tokenSwapData = tokensSwapData[token];
+        if (paymasterDataLength != 20) {
+            require(paymasterDataLength >= 24, "must contain \"permit\" and token");
             require(
-                permitMethodSignature == GsnUtils.getMethodSig(relayRequest.relayData.paymasterData),
-                "paymasterData: wrong method sig");
+                tokenSwapData.permitMethodSignature == GsnUtils.getMethodSig(relayRequest.relayData.paymasterData),
+                "wrong \"permit\" method sig");
```
forshtat, line 202: I know I wrote the original code like this, but now I guess it would make much more sense to encode on-chain then to decode-and-compare on-chain now, what do you think?

packages/paymasters/contracts/PermitERC20UniswapV3Paymaster.sol
```diff
@@ -147,33 +225,109 @@ contract PermitERC20UniswapV3Paymaster is BasePaymaster, ERC2771Recipient {
...
-        UniswapV3Helper.swapToEth(
-            address(weth),
-            address(token),
-            ethActualCharge,
-            uniswapPoolFee,
-            uniswap
-        );
-        relayHub.depositFor{value : address(this).balance}(address(this));
+        (address payer, uint256 priceQuote, uint256 tokenPreCharge) = abi.decode(context, (address, uint256, uint256));
+        IERC20Metadata token = _getTokenFromPaymasterData(relayData.paymasterData);
```
forshtat, line 229: Would it be cheaper to encode/decode it with context? Just asking, I don't really care.

packages/paymasters/contracts/PermitERC20UniswapV3Paymaster.sol
```diff
@@ -147,33 +225,109 @@ contract PermitERC20UniswapV3Paymaster is BasePaymaster, ERC2771Recipient {
...
+        uint256 balance = address(this).balance;
+        uint256 amountSwapped = 0;
+        if (balance < depositAmount) {
+            for (uint256 i = 0; i < tokens.length && balance + amountSwapped < depositAmount; i++) {
+                IERC20Metadata tokenIn = tokens[i];
+                uint256 tokenBalance = tokenIn.balanceOf(address(this));
+                if (tokenBalance > 0) {
+                    TokenSwapData memory tokenSwapData = tokensSwapData[tokenIn];
+                    uint256 quote = uint256(tokenSwapData.priceFeed.latestAnswer());
+                    uint256 amountOutMin = _tokenToWei(tokenBalance, tokenSwapData.priceDivisor, quote) * 99 / 100;
```
forshtat, line 258: I assume `99` here means maximum slippage of 1% ? in this case I would suggest making it configureable.

packages/paymasters/contracts/PermitERC20UniswapV3Paymaster.sol
```diff
@@ -147,33 +225,109 @@ contract PermitERC20UniswapV3Paymaster is BasePaymaster, ERC2771Recipient {
...
+    }
+
+    function _refillHubDeposit(uint256 depositAmount) private {
+        uint256 balance = address(this).balance;
+        uint256 amountSwapped = 0;
+        if (balance < depositAmount) {
+            for (uint256 i = 0; i < tokens.length && balance + amountSwapped < depositAmount; i++) {
+                IERC20Metadata tokenIn = tokens[i];
+                uint256 tokenBalance = tokenIn.balanceOf(address(this));
+                if (tokenBalance > 0) {
```
forshtat, line 255: 0 may be a bar way too low. I think it would actually be cheaper to sort tokens by total value or something like that, maybe as an alternative path, or we can get a hint on what to do as another parameter in `paymasterData`. Here you would always start swapping with the same token so it is likely you will be going the opposite way - going from the least valuable to the most valuable one.

packages/paymasters/contracts/PermitERC20UniswapV3Paymaster.sol
```diff
@@ -147,33 +225,109 @@ contract PermitERC20UniswapV3Paymaster is BasePaymaster, ERC2771Recipient {
...
+        }
+        uint256 depositAmount = targetHubBalance - hubBalance + ethActualCharge;
+        _refillHubDeposit(depositAmount);
+    }
+
+    function _refillHubDeposit(uint256 depositAmount) private {
+        uint256 balance = address(this).balance;
+        uint256 amountSwapped = 0;
+        if (balance < depositAmount) {
+            for (uint256 i = 0; i < tokens.length && balance + amountSwapped < depositAmount; i++) {
```
forshtat, line 252: Also a bit concerned about the potentially very, very expensive loop in the `postRelayCall`. 
It can, quite possibly, make the Paymaster inoperable as it may require waaay more gas then available here.
I think for the "forced" swap in users' transaction we limit to a single swap or something like that.

packages/paymasters/contracts/PermitERC20UniswapV3Paymaster.sol
```diff
@@ -147,33 +225,109 @@ contract PermitERC20UniswapV3Paymaster is BasePaymaster, ERC2771Recipient {
...
+                    amountSwapped += amountOut;
+                }
+            }
+            UniswapV3Helper.unwrapWeth(uniswap, amountSwapped);
+        }
+        if (balance + amountSwapped < depositAmount) {
+            emit FundingNeeded();
+            depositAmount = balance + amountSwapped;
+        }
+        relayHub.depositFor{value : depositAmount}(address(this));
```
forshtat, line 276: TokenPaymaster should never hold ether. Can we send `balanceOf(this)`?

package.json
shahafn: I put it in config.yml. moved...

packages/paymasters/contracts/PermitERC20UniswapV3Paymaster.sol
shahafn: This is your own code

packages/paymasters/src/PermitPaymasterUtils.ts
forshtat: In this case I believe they should be defined in the tests or package themselves. Or, we may need to check in a `networks.json` file with supported contracts across networks.

packages/paymasters/contracts/PermitERC20UniswapV3Paymaster.sol
forshtat: I tend to agree with Dror and TypeScript here - optional fields should be in the end. I cannot explain why but this feels like this can make sense.

packages/paymasters/contracts/PermitERC20UniswapV3Paymaster.sol
forshtat: You know what? It is fine.

packages/paymasters/contracts/PermitERC20UniswapV3Paymaster.sol
forshtat, line 252: Please consider extracting the insides of this `for` loop into a separate inner function.

packages/paymasters/src/PermitPaymasterUtils.ts
```diff
@@ -80,6 +101,22 @@ export interface PermitInterfaceEIP2612 {
...
+  tokens: string[]
+  relayHub: string
+  priceFeeds: string[]
+  uniswap: string
+  trustedForwarder: string
+  uniswapPoolFees: number[] | BN[] | string[]
+  gasUsedByPost: number | BN | string
+  permitMethodSignatures: string[]
+  minHubBalance: number | BN | string
+  targetHubBalance: number | BN | string
```
forshtat, line 115: Suggestion: make sure that these values can accept `MAX_INT` to signify special configurations:
(MAX_INT does not necessarily be 2^256 here as this will likely break our math... TBD)

`minHubBalance = MAX_INT` - after each transaction perform a swap to ether
`targetHubBalance = MAX_INT` - when replenishing, do not stop and swap all tokens

packages/paymasters/contracts/PermitERC20UniswapV3Paymaster.sol
shahafn: Fixed

packages/paymasters/contracts/PermitERC20UniswapV3Paymaster.sol
shahafn: Fixed

packages/common/src/ConfigResponse.ts
```diff
@@ -105,6 +106,16 @@ export interface GSNConfig {
...
+  * The address of the token paymaster contract used by TokenPaymasterProvider
+  */
+  tokenPaymasterAddress?: Address
+
+  /**
+  * Fields required by TokenPaymasterProvider for the supported tokens
+  */
+  domainSeparators?: Record<Address, EIP712Domain>
```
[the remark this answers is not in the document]
shahafn: Fixed

packages/common/src/EIP712/TypedRequestData.ts
```diff
@@ -74,7 +74,7 @@ export const GsnDomainSeparatorType = {
   version: '3'
 }
 
-export function getDomainSeparator (verifier: Address, chainId: number): Record<string, unknown> {
+export function getDomainSeparator (verifier: Address, chainId: number): EIP712Domain {
```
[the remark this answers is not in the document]
shahafn: I think that its type is EIP712Domain, it's more clear that way as an exported function

packages/paymasters/contracts/PermitERC20UniswapV3Paymaster.sol
shahafn: Fixed

packages/paymasters/contracts/PermitERC20UniswapV3Paymaster.sol
```diff
@@ -269,11 +275,7 @@ contract PermitERC20UniswapV3Paymaster is BasePaymaster, ERC2771Recipient {
             }
             UniswapV3Helper.unwrapWeth(uniswap, amountSwapped);
         }
-        if (balance + amountSwapped < depositAmount) {
-            emit FundingNeeded();
-            depositAmount = balance + amountSwapped;
-        }
-        relayHub.depositFor{value : depositAmount}(address(this));
+        relayHub.depositFor{value : balance + amountSwapped}(address(this));
```
forshtat, line 278: Or, do the `{value:  address(this).balance}`. This is boolet-proof while `balance + amountSwapped` really depends on `amountSwapped` always being correct, which is an assumption that 3rd party solidity code is precise to a gwei.

packages/paymasters/contracts/PermitERC20UniswapV3Paymaster.sol
```diff
@@ -147,33 +208,110 @@ contract PermitERC20UniswapV3Paymaster is BasePaymaster, ERC2771Recipient {
...
+        uint256 amountSwapped = 0;
+        if (balance < depositAmount) {
+            for (uint256 i = 0; i < tokens.length && balance + amountSwapped < depositAmount; i++) {
+                IERC20Metadata tokenIn = tokens[i];
+                uint256 tokenBalance = tokenIn.balanceOf(address(this));
+                if (tokenBalance > 0) {
+                    uint256 quote = uint256(priceFeeds[tokenIn].latestAnswer());
+                    uint256 divisor = priceDivisors[tokenIn];
+                    uint24 poolFee = uniswapPoolFees[tokenIn];
+                    uint256 amountOutMin = _tokenToWei(tokenBalance, divisor, quote) * 99 / 100;
```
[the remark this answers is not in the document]
shahafn: Now it's configurable

packages/paymasters/contracts/PermitERC20UniswapV3Paymaster.sol
shahafn: The checks don't entirely overlap

packages/paymasters/contracts/PermitERC20UniswapV3Paymaster.sol
```diff
@@ -15,71 +16,75 @@ import "@opengsn/contracts/src/utils/GsnUtils.sol";
...
-    IERC20 public token;
-    IERC20 public weth;
+    using SafeERC20 for IERC20Metadata;
 
-    uint24 public uniswapPoolFee;
-    uint256 public gasUsedByPost;
-    bytes4 public permitMethodSignature;
-    uint256 public permitMethodDataLength;
+    event Received(address indexed sender, uint256 eth);
+    event FundingNeeded();
```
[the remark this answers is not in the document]
shahafn: Removed

packages/paymasters/contracts/PermitERC20UniswapV3Paymaster.sol
```diff
@@ -15,71 +16,75 @@ import "@opengsn/contracts/src/utils/GsnUtils.sol";
...
-    }
 
-    function _msgSender() internal view override(Context, ERC2771Recipient) returns (address sender) {
-        sender = ERC2771Recipient._msgSender();
-    }
+    // Upon reaching minHubBalance, the paymaster will deposit eth to RelayHub to reach targetHubBalance
+    uint256 public minHubBalance;
+    uint256 public targetHubBalance;
+    // Minimum eth amount, above targetHubBalance, to send to the owner
+    uint256 public minWithdrawalAmount;
```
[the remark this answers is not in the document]
shahafn: Why have any uniswap interaction then? just maintain it through a server that does the swapping on behalf of the paymaster. I disagree with it

packages/paymasters/contracts/PermitERC20UniswapV3Paymaster.sol
```diff
@@ -15,71 +16,75 @@ import "@opengsn/contracts/src/utils/GsnUtils.sol";
...
+    uint256 public paymasterFee;
 
-    function _msgData() internal view override(Context, ERC2771Recipient) returns (bytes memory) {
-        return ERC2771Recipient._msgData();
+    struct PaymasterConfig {
+        IERC20 weth;
+        IERC20Metadata[] tokens;
```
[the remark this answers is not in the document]
shahafn: How does all of this matter? Making it `address` type won't do anything other than make me cast it everywhere. Off chain it's passed as an address anyway. I don't understand how that matters.

packages/paymasters/contracts/PermitERC20UniswapV3Paymaster.sol
shahafn: We talked about it, it's a small thing. Dror suggested having it as it is for debugging purposes, it's easier to see strings than bytes if/when something goes wrong

packages/paymasters/contracts/PermitERC20UniswapV3Paymaster.sol
shahafn: Fixed

packages/paymasters/contracts/PermitERC20UniswapV3Paymaster.sol
shahafn: Doesn't make much difference it seems. I can change it if it's important

packages/paymasters/contracts/PermitERC20UniswapV3Paymaster.sol
shahafn: Added a threshold

packages/paymasters/contracts/PermitERC20UniswapV3Paymaster.sol
shahafn: Added a threshold that pretty much solves it

packages/paymasters/contracts/PermitERC20UniswapV3Paymaster.sol
shahafn: Fixed

packages/paymasters/src/PermitPaymasterUtils.ts
shahafn: That's already how it works

packages/paymasters/contracts/PermitERC20UniswapV3Paymaster.sol
shahafn: It was simply cheaper that way

packages/paymasters/src/PermitPaymasterUtils.ts
forshtat: Ok than

packages/paymasters/contracts/PermitERC20UniswapV3Paymaster.sol
```diff
@@ -147,33 +239,116 @@ contract PermitERC20UniswapV3Paymaster is BasePaymaster, ERC2771Recipient {
...
-        UniswapV3Helper.swapToEth(
-            address(weth),
-            address(token),
-            ethActualCharge,
-            uniswapPoolFee,
-            uniswap
-        );
-        relayHub.depositFor{value : address(this).balance}(address(this));
+        (address payer, uint256 priceQuote, uint256 tokenPreCharge) = abi.decode(context, (address, uint256, uint256));
+        IERC20Metadata token = _getTokenFromPaymasterData(relayData.paymasterData);
```
drortirosh, line 243: yes, don't assume you have paymasterData in the postRelayedCall (since in AA there is only context)

packages/paymasters/contracts/PermitERC20UniswapV3Paymaster.sol
```diff
@@ -147,33 +239,116 @@ contract PermitERC20UniswapV3Paymaster is BasePaymaster, ERC2771Recipient {
...
+        (uint256 tokenActualCharge, uint256 ethActualCharge) = _calculateCharge(relayData, gasUseWithoutPost + gasUsedByPost, priceQuote, tokensSwapData[token].priceDivisor);
```
drortirosh, line 244: we assume that gasUsedByPost is constant, regardless of token in use.
I would like to keep it this way (to avoid the complexity of calculating it separately for each token), but we do need to verify that at least for the known tokens, it is close enough (that is, the total cost of of the relayed TX doesn't fluctuate too much with different tokens.)

packages/paymasters/contracts/PermitERC20UniswapV3Paymaster.sol
```diff
@@ -15,71 +16,79 @@ import "@opengsn/contracts/src/utils/GsnUtils.sol";
...
 
-    IChainlinkOracle public priceFeed;
-    ISwapRouter public uniswap;
-    IERC20 public token;
-    IERC20 public weth;
+    struct TokenSwapData {
+        IChainlinkOracle priceFeed;
+        uint24 uniswapPoolFee;
+        // between 0 to 1000, with 2 decimals, that is, 10 = 1%
+        uint8 slippage;
```
drortirosh, line 36: what is the meaning of slippage ? 
it is the name and type of uniswap param, but you don't use it for that purpose at all.

packages/paymasters/contracts/PermitERC20UniswapV3Paymaster.sol
```diff
@@ -147,33 +239,116 @@ contract PermitERC20UniswapV3Paymaster is BasePaymaster, ERC2771Recipient {
...
+    }
+
+    function addSlippage(uint256 amount, uint8 slippage) public pure returns (uint256) {
+        return amount * (1000 - slippage) / 1000;
     }
 
     // as this Paymaster already has a permission from a user to operate the tokens on behalf of the gasless account,
     // it makes this same Paymaster a great recipient of a transaction if its only action is a pure token transfer
-    function transferToken(address target, uint256 value) external {
+    function transferToken(IERC20Metadata token, address target, uint256 value) external {
```
drortirosh, line 317: This helper should be `transferAndCall(IERC20 token, uint value, address target, bytes callData)`
This makes integration with protocols such as "uniswap" simpler (otherwise, we need an explicit "UniswapPaymaster" to add just that)
It allows integration with any protocol that receives tokens, and performs a one-off method call (that is, doesn't depend about the sender of the tokens, which requires the protocol to be GSN-aware)

packages/paymasters/contracts/PermitERC20UniswapV3Paymaster.sol
```diff
@@ -147,33 +239,116 @@ contract PermitERC20UniswapV3Paymaster is BasePaymaster, ERC2771Recipient {
...
-        require(token.transferFrom(_msgSender(), target, value), "transferToken failed");
+        token.safeTransferFrom(_msgSender(), target, value);
     }
 
     receive() external override payable {
         emit Received(msg.sender, msg.value);
     }
+
+    function getGasAndDataLimits() public override pure returns (IPaymaster.GasAndDataLimits memory limits) {
+        return IPaymaster.GasAndDataLimits(
```
drortirosh, line 327: where these values came from?

hardhat.config.ts
```diff
@@ -4,7 +4,7 @@
 module.exports = {
   solidity: '0.8.7',
   networks: {
-    hardhat: { chainId: 1337 },
```
forshtat, line 7: If I may suggest, it would be better to have this `hardhatChainId` calculated outside.

packages/paymasters/contracts/helpers/UniswapV3Helper.sol
```diff
@@ -6,14 +6,15 @@ import "@uniswap/v3-periphery/contracts/interfaces/ISwapRouter.sol";
 import "@uniswap/v3-periphery/contracts/interfaces/IPeripheryPayments.sol";
 
 library UniswapV3Helper {
+    event UniswapReverted(address tokenIn, address tokenOut, uint256 amountIn, uint256 amountOutMin);
     // turn ERC-20 tokens into native unwrapped ETH at market price
-    function swapToEth(
-        address weth,
+    function swapToWeth(
```
forshtat, line 11: The comment here says 'native unwrapped' - should say 'wrapped'

packages/paymasters/src/TokenPaymasterInteractor.ts
```diff
@@ -0,0 +1,51 @@
...
+
+export class TokenPaymasterInteractor {
+  private readonly provider: Web3ProviderBaseInterface
+  private readonly PermitInterfaceDAIToken: Contract<PermitInterfaceDAIInstance>
+  private readonly PermitInterfaceEIP2612Token: Contract<PermitInterfaceEIP2612Instance>
+  private readonly PermitERC20UniswapV3Paymaster: Contract<PermitERC20UniswapV3PaymasterInstance>
+
+  // daiPermittableToken!: PermitInterfaceDAIInstance
+  // eip2612PermittableToken!: PermitInterfaceEIP2612Instance
+  // permitPaymasterInstance!: PermitERC20UniswapV3PaymasterInstance
```
forshtat, line 19: commented-out code

packages/paymasters/src/TokenPaymasterProvider.ts
```diff
@@ -0,0 +1,129 @@
...
+  signAndEncodeEIP2612Permit
+} from './PermitPaymasterUtils'
+import { constants } from '@opengsn/common/dist/Constants'
+import { EIP712Domain, EIP712DomainType, EIP712DomainTypeWithoutVersion } from '@opengsn/common/dist/EIP712/TypedRequestData'
+import { TokenPaymasterInteractor } from './TokenPaymasterInteractor'
+
+export interface TokenPaymasterConfig extends GSNConfig {
+  tokenAddress: Address
+  permitAmount?: string | number | BN
+}
```
forshtat, line 26: Well, this does not make sense to me. First you add `tokenPaymasterAddress` and `tokenPaymasterDomainSeparators` to the `GSNConfig` and then you inherit from `GSNConfig` and add two more fields. This is two bad solutions used at once. 

I think if we want the `TokenPaymasterProvider` to be a first-class Provider then its configuration is part of the GSNConfig. We may need to find a better arrangement later.

packages/paymasters/src/TokenPaymasterProvider.ts
```diff
@@ -0,0 +1,129 @@
...
+
+  constructor (relayClient: RelayClient, provider: Web3ProviderBaseInterface) {
+    super(relayClient)
+    this.tokenPaymasterInteractor = new TokenPaymasterInteractor(provider)
+  }
+
+  static newProvider (input: TokenPaymasterUnresolvedConstructorInput): TokenPaymasterProvider {
+    if (input.config.maxPaymasterDataLength == null) {
+      input.config.maxPaymasterDataLength = 300
+    }
```
forshtat, line 47: This is not how we fill in missing configuration fields. `maxPaymasterDataLength` defaults to `0` which is not `== null` anyways.

packages/paymasters/src/TokenPaymasterProvider.ts
```diff
@@ -0,0 +1,129 @@
...
+  }
+
+  async init (): Promise<this> {
+    await super.init(true)
+    this.relayClient.dependencies.asyncPaymasterData = this._buildPaymasterData.bind(this)
+    // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+    this.paymaster = await this.tokenPaymasterInteractor._createPermitERC20UniswapV3Paymaster(this.relayClient.config.tokenPaymasterAddress!)
+    if (this.config.tokenAddress != null) {
+      await this.useToken(this.config.tokenAddress)
+    }
```
forshtat, line 59: How would you initialize a `TokenPaymasterProvider` without selecting a token? This should not be an option, I think.

packages/paymasters/src/TokenPaymasterProvider.ts
```diff
@@ -0,0 +1,129 @@
...
+    //  Optionally encode permit method,then concatenate token address
+    if (relayRequest.relayData.paymaster !== this.paymaster.address) {
+      throw new Error('Paymaster address mismatch')
+    }
+    const allowance = await this.token.allowance(relayRequest.request.from, relayRequest.relayData.paymaster)
+    let permitMethod = ''
+    if (allowance.eqn(0)) {
+      if (this.config.tokenPaymasterDomainSeparators == null) {
+        throw new Error('TokenPaymasterProvider not initialized. Call init() first')
+      }
```
forshtat, line 73: `init`-check should be on the top of every method anyways.

packages/paymasters/src/TokenPaymasterProvider.ts
```diff
@@ -0,0 +1,129 @@
...
+      const domainSeparator: EIP712Domain = this.config.tokenPaymasterDomainSeparators[this.token.address]
```
forshtat, line 74: `.toLowerCase` the addresses as they sometimes get cHecKsuMmeD when you least expect it

packages/paymasters/src/TokenPaymasterProvider.ts
```diff
@@ -0,0 +1,129 @@
...
+          this.token.address,
+          constants.MAX_UINT256.toString(),
+          constants.MAX_UINT256.toString(),
+          this.web3,
+          domainSeparator,
+          domainSeparator.version == null ? EIP712DomainTypeWithoutVersion : EIP712DomainType
+        )
+      }
+    }
+    // return abi.encodeParameters(['bytes', 'address'], [permitMethod, this.token.address])
```
forshtat, line 97: commented-out code

packages/paymasters/src/TokenPaymasterProvider.ts
```diff
@@ -0,0 +1,129 @@
...
+    return '0x' + removeHexPrefix(this.token.address) + removeHexPrefix(permitMethod)
+  }
+
+  async useToken (tokenAddress: Address): Promise<void> {
```
forshtat, line 101: `useXxx` is a name convention used and abused by the React framework for Hooks and given we target web developers with the library we should not call our methods `useSomthing`.
Call it `setTokenAddress` or something.

packages/common/src/ConfigResponse.ts
```diff
@@ -105,6 +106,16 @@ export interface GSNConfig {
...
 
+  /**
+  * The address of the token paymaster contract used by TokenPaymasterProvider
+  */
+  tokenPaymasterAddress?: Address
+
+  /**
+  * Fields required by TokenPaymasterProvider for the supported tokens
+  */
+  tokenPaymasterDomainSeparators?: Record<Address, EIP712Domain>
```
forshtat, line 117: Does not have to be `?` as the `Record<>` is allowed to be empty, I believe.

packages/paymasters/src/TokenPaymasterProvider.ts
```diff
@@ -0,0 +1,129 @@
...
+    const isSupported = await this.isTokenSupported(tokenAddress)
+    if (!isSupported) {
+      throw new Error(`token ${tokenAddress} not supported`)
+    }
+    this.config.tokenAddress = toChecksumAddress(tokenAddress)
+    if (this.config.tokenPaymasterDomainSeparators == null || this.config.tokenPaymasterDomainSeparators[this.config.tokenAddress] == null) {
+      throw new Error(`Domain separator not found for token ${tokenAddress}`)
+    }
+    const permitSigHash = (await this.paymaster.getTokenSwapData(tokenAddress)).permitMethodSignature
```
forshtat, line 110: can't you `const {permitMethodSignature} = await this.paymaster.getTokenSwapData` ?

packages/paymasters/src/WrapContract.ts
```diff
@@ -0,0 +1,34 @@
+import { Contract, providers } from 'ethers'
+import { Eip1193Bridge } from '@ethersproject/experimental'
+import { ExternalProvider } from '@ethersproject/providers'
+import { Signer } from '@ethersproject/abstract-signer'
+
+import { TokenPaymasterConfig, TokenPaymasterProvider } from './TokenPaymasterProvider'
+import { GSNDependencies, WrapBridge } from '@opengsn/provider/dist'
+
+export async function wrapContract (
```
forshtat, line 9: No, wait, we cannot have top-level global functions with same names in our packages, this is going to be hell. Please rename to `wrapContractTokenPaymaster` or something like that.

packages/relay/src/RelayServer.ts
```diff
@@ -521,10 +521,6 @@ returnValue        | ${viewRelayCallRet.returnValue}
 
     this.chainId = this.contractInteractor.chainId
     this.networkId = this.contractInteractor.getNetworkId()
-    if (this.config.devMode && (this.chainId < 1000 || this.networkId < 1000)) {
-      this.logger.error('Don\'t use real network\'s chainId & networkId while in devMode.')
-      process.exit(-1)
-    }
```
forshtat, line 527: there is this code:
```
if (devMode) {
    if (fs.existsSync(`${workdir}/${TXSTORE_FILENAME}`)) {
      fs.unlinkSync(`${workdir}/${TXSTORE_FILENAME}`)
    }
```

This is the only place I've seen with 'devMode' being used - and this is before the check.

Do we even need this parameter?

packages/paymasters/contracts/PermitERC20UniswapV3Paymaster.sol
shahafn: It's not the name of a uniswap param

packages/paymasters/src/TokenPaymasterProvider.ts
shahafn: This happens before the default but after user input...

packages/paymasters/contracts/PermitERC20UniswapV3Paymaster.sol
```diff
@@ -15,71 +16,79 @@ import "@opengsn/contracts/src/utils/GsnUtils.sol";
...
-
-    function _msgSender() internal view override(Context, ERC2771Recipient) returns (address sender) {
-        sender = ERC2771Recipient._msgSender();
-    }
-
-    function _msgData() internal view override(Context, ERC2771Recipient) returns (bytes memory) {
-        return ERC2771Recipient._msgData();
+    struct PaymasterConfig {
+        IERC20 weth;
+        IERC20Metadata[] tokens;
```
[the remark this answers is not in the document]
shahafn: Fixed

packages/paymasters/contracts/PermitERC20UniswapV3Paymaster.sol
```diff
@@ -119,23 +201,33 @@ contract PermitERC20UniswapV3Paymaster is BasePaymaster, ERC2771Recipient {
...
     internal
     override
-    returns (bytes memory context, bool revertOnRecipientRevert) {
-        (signature);
-        require(approvalData.length == 0, "approvalData: invalid length");
-        if (relayRequest.relayData.paymasterData.length != 0) {
-            require(relayRequest.relayData.paymasterData.length >= 4, "paymastaData: missing method sig");
+    returns (bytes memory, bool) {
+        (signature, approvalData);
+        uint256 paymasterDataLength = relayRequest.relayData.paymasterData.length;
```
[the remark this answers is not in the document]
shahafn: Fixed

packages/paymasters/contracts/PermitERC20UniswapV3Paymaster.sol
shahafn: Fixed

packages/paymasters/contracts/helpers/UniswapV3Helper.sol
shahafn: Fixed

packages/paymasters/src/TokenPaymasterInteractor.ts
shahafn: Removed

packages/paymasters/src/TokenPaymasterProvider.ts
shahafn: Well we have a bit of an issue, since it sits in paymasters package, but the original RelayClient, RelayProvider and GSNConfig sit in other packages, which cannot depend on paymasters. The issue with EIP712Domain is that it's difficult to build it from the token as it may vary between networks, even for the same token. That's why I have it as config param.

packages/paymasters/src/TokenPaymasterProvider.ts
shahafn: Removed

packages/relay/src/RelayServer.ts
shahafn: I think some tests still rely on removing these files, but I believe we can remove it and remove those files in the tests themselves and not from runServer...

packages/paymasters/src/TokenPaymasterProvider.ts
forshtat: Use-case: initialize Provider and select token on the go. Niche but viable.

packages/paymasters/src/TokenPaymasterProvider.ts
shahafn: Fixed

packages/paymasters/src/TokenPaymasterProvider.ts
shahafn: Using checksummed address only

packages/paymasters/src/TokenPaymasterProvider.ts
shahafn: fixed

packages/common/src/ConfigResponse.ts
shahafn: Fixed

packages/paymasters/src/TokenPaymasterProvider.ts
shahafn: Fixed

packages/paymasters/contracts/PermitERC20UniswapV3Paymaster.sol
shahafn: Fixed

packages/paymasters/contracts/PermitERC20UniswapV3Paymaster.sol
```diff
@@ -15,101 +16,163 @@ import "@opengsn/contracts/src/utils/GsnUtils.sol";
...
-        uint256 price = uint256(priceFeed.latestAnswer());
-        uint256 decimals = uint256(priceFeed.decimals());
-        tokenPreCharge = ethMaxCharge * price / (10 ** decimals);
+    returns (uint256 tokenCharge, uint256 ethCharge) {
+        ethCharge = relayHub.calculateCharge(gasUsed, relayData);
+        tokenCharge = addPaymasterFee(_weiToToken(ethCharge, priceDivisor, priceQuote));
     }
 
+    function _tokenToWei(uint256 amount, uint256 divisor, uint256 quote) internal pure returns(uint256) {
+        return 1e18 * amount * quote / divisor;
```
drortirosh, line 164: suggestion: "fold" the divisor into the quote: 
- quote should a "fixed point" value of actualQuote *1e18/divisor
- tokenToWei:  return amount * quote/1e18
- weiToToken: amount*1e18/quote

forshtat (approved):

PR #874, opened 2022-07-31
fix: gsn start command only checks localhost as value of network option

ahnafatef:


forshtat:
@ahnafatef we have decided to remove the `--withNode` option as it removes all Hardhat logs output:
https://github.com/opengsn/gsn/pull/873

PR #876, opened 2022-08-02
OG-893: GSN constructor called with ethers provider should wrap it automatically

forshtat:


backmeupplz:
@forshtat  Any workaround for this until this gets merged and released?

backmeupplz:
Nvm, created https://github.com/opengsn/gsn/pull/877 — this can act as a workaround for now

forshtat:
@backmeupplz you can import `wrapSigner` and `wrapContract` already - just use the full file path; it will take some time to merge the PR and publish package to the NPM.

packages/provider/src/RelayClient.ts
```diff
@@ -110,6 +113,23 @@ export class RelayClient {
     }
     this.rawConstructorInput = rawConstructorInput
     this.logger = rawConstructorInput.overrideDependencies?.logger ?? console
+    this.wrapEthersJsProvider()
+  }
+
+  wrapEthersJsProvider (): void {
+    const provider = this.rawConstructorInput.provider as any
+    if (typeof provider.getSigner === 'function') {
```
drortirosh, line 121: Afaik, ethers had a _provider marker property instead of checking for method

packages/common/src/Utils.ts
```diff
@@ -476,3 +476,34 @@ export async function waitForSuccess<T> (
     }
   })
 }
+
+/**
+ * From https://ethereum.stackexchange.com/a/43649
+ * May be outdated but should do the job
+ */
+export function getProviderImplementation (provider: any): string {
```
drortirosh, line 484: Do we need this? Determine the type of ropc provider is unrelated to the issue of supporting ethers.

packages/provider/src/RelayClient.ts
forshtat: Yes, but not all `ethers providers` have a `getSigner` and we cannot really use a provider that does not have a signer.

packages/common/src/Utils.ts
forshtat: Not really. Just tried to check as many known providers as possible so we only log the warning for ones that are not recognized at all. Can be simplified or even removed.

drortirosh (approved):

PR #880, opened 2022-08-08
OG-897: Remove older version of 'ethereumjs-wallet' dependency

forshtat:


drortirosh:
did you verify this code works with old keys? that is, if a relay is upgraded from previous versions, that it can still use the same previous keys?

drortirosh:
no response here?

PR #885, opened 2022-08-23
OG-908: Cannot catch exception thrown in ContractInteractor

forshtat:
The RPC call may return an object with 'error', 'result', both or none.

Our code would always assume 'result' to be at least non-null, but it is.


packages/common/src/ContractInteractor.ts
```diff
@@ -544,7 +544,7 @@ export class ContractInteractor {
           } else if (errorAsBoolean(err)) {
             reject(err)
           } else {
-            resolve(res.result)
+            resolve(res?.result ?? 'RPC call returned no error and no result')
```
drortirosh, line 547: if there is no result and no error then its an ERROR, not a result...

drortirosh (approved):

PR #886, opened 2022-08-24
OG-898: Make GsnDomainSeparatorType.name field configurable

forshtat:


packages/common/src/ConfigResponse.ts
```diff
@@ -169,4 +169,11 @@ export interface GSNConfig {
    * The number of milliseconds to wait after the first Relay Server responds to the ping before picking a winner.
    */
   waitForSuccessPingGrace: number
+
+  /**
+   * The name of the EIP-712 Domain Separator field. Note that this is usually the name of the requiest the
+   * users will see in MetaMask or other wallets.
+   * Warning! The domain type must be first registered on-chain by calling 'Forwarder::registerDomainSeparator'.
```
drortirosh, line 176: its a NOTE, not a "warning"...

packages/cli/src/CommandsLogic.ts
```diff
@@ -585,7 +586,7 @@ export class CommandsLogic {
     if (deployOptions.deployPaymaster ?? false) {
       pmInstance = await this.deployPaymaster({ ...options }, rInstance.options.address, fInstance, deployOptions.skipConfirmation)
     }
-    await registerForwarderForGsn(fInstance, options)
+    await registerForwarderForGsn(defaultGsnConfig.domainSeparatorName, fInstance, options)
```
drortirosh, line 589: Hm... if any project can register whatever "domain separator" and no contract code is there to validate it, then maybe we should change the forwarder to ignore the "domain separator" altogether....

drortirosh: never mind.
GSN doesn't require the domain-separator to be valid, since the request in an elaborate RelayRequest, which is good enough.

drortirosh (approved):

packages/common/src/ConfigResponse.ts
forshtat: Fixed.

drortirosh (approved):

PR #887, opened 2022-08-25
OG-906: OG-903: Enable the private RelayServer registration and operation

forshtat:
OG-903: Add 'blacklisted recipients' configuration field


packages/relay/src/RegistrationManager.ts
```diff
@@ -397,7 +397,13 @@ export class RegistrationManager {
       transactions = transactions.concat(txHash)
       skipGasEstimationForRegisterRelay = true
     }
-    const registerMethod = await this.contractInteractor.getRegisterRelayMethod(this.hubAddress, this.config.url)
+    let registeredUrl: string
+    if (this.config.privateMode) {
+      registeredUrl = ''
```
drortirosh, line 402: instead of ignoring the config url, maybe private mode should validate the url is not set

forshtat: Well, I was thinking a relay operator who wants to change from private to public would rather flip a flag than remove/add the string. Especially relevant for testing.

drortirosh (approved):

PR #890, opened 2022-08-29
OG-902: RelayServer should support having multiple RPC endpoints

forshtat:


drortirosh:
Instead of implementing that from scratch, there is a long due feature to use ethers instead of web3.
one feature of ethers is a "FallbackProvider, where it uses a "quorum" of connected servers.

forshtat:
> Instead of implementing that from scratch, there is a long due feature to use ethers instead of web3. one feature of ethers is a "FallbackProvider, where it uses a "quorum" of connected servers.

How feasible is it to switch from Web3.js to Ethers.js at this stage? Would you like to do it?

drortirosh:
this solve part of the problem (sendTransaction to multiple providers), but there are some other nice things we can do with multiple providers - e.g. even getting current block number..
might be a better approach to wrap multiple providers by a single provider (copying some of the logic of ethers.FallbackProvider - after all, someone thought what is the right behaviour for such multi-provider provider...

packages/cli/src/GsnTestEnvironment.ts
```diff
@@ -176,6 +176,7 @@ class GsnTestEnvironmentClass {
     const contractInteractor = new ContractInteractor(
       {
         provider: new Web3.providers.HttpProvider(host),
```
drortirosh, line 178: better make `providers[]` parameter, which is an array of all underlying providers. 
the first one is the "primary", and the rest are "alternates"

packages/relay/src/runServer.ts
```diff
@@ -59,23 +60,35 @@ function sanitizeJsonRpcResponse (response?: JsonRpcResponse): JsonRpcResponse |
   return clone
 }
 
+function initializeAlternateProviders (alternateEthereumNodeUrls: string[] = []): Web3ProviderBaseInterface[] {
+  const alternateProviders: Web3ProviderBaseInterface[] = []
+  for (const url of alternateEthereumNodeUrls) {
+    alternateProviders.push(
+      new Web3.providers.HttpProvider(url)
```
drortirosh, line 67: I *think* HttpProvider as a "timeout" parameter. this way, we don't wait indefinitely for each request, but up to a timeout.
(the alternative is to wrap the call to "sendTransaction" with a setTimeout ...)

forshtat:
Will not be added to 3.0

Will be added later, potentially by using Ethers.js in the Relay Server

PR #894, opened 2022-09-07
fix eth_getTransactionByHash: actualTransactionHash same as relayRequ…

akkien:
Fix the error: The api eth_getTransactionByHash of relayProvider return actualTransactionHash same as relayRequestID


forshtat (approved):

forshtat:
Thanks @akkien !

PR #896, opened 2022-09-08
OG-716 OG-926 reducing getBalance() rpc calls and fixing balance readiness checks

shahafn:


packages/relay/src/RelayServer.ts
```diff
@@ -688,12 +692,32 @@ latestBlock timestamp   | ${latestBlock.timestamp}
...
-      this.setReadyState(false)
+    const managerBalanceNotReady = await this._refreshAndCheckBalances()
+    if (managerBalanceNotReady) {
       return []
     }
-    return await this._handleChanges(block, currentBlockTimestamp)
+    return await this._handleChanges(block)
+  }
+
+  async _refreshAndCheckBalances (): Promise<boolean> {
```
forshtat, line 702: There is no reason for this method to return anything at all.

packages/relay/src/RelayServer.ts
```diff
@@ -688,12 +692,32 @@ latestBlock timestamp   | ${latestBlock.timestamp}
...
+    let managerBalanceNotReady = false
+    let workerBalanceNotReady = false
+    if (this.shouldRefreshBalances) {
+      await this.registrationManager.refreshBalance()
+      this.workerBalanceRequired.currentValue = await this.getWorkerBalance(0)
+
+      managerBalanceNotReady = this.registrationManager.balanceRequired.currentValue.lt(toBN(this.config.managerMinBalance.toString()).divn(minBalanceToNotReadyFactor))
+      workerBalanceNotReady = this.workerBalanceRequired.currentValue.lt(toBN(this.config.workerMinBalance.toString()).divn(minBalanceToNotReadyFactor))
+      if (managerBalanceNotReady || workerBalanceNotReady) {
+        this.logger.debug(`${managerBalanceNotReady ? 'manager' : 'worker'} balance too low`)
```
forshtat, line 713: Here, either 'manager', 'worker' or both balances are too low.

packages/relay/src/RelayServer.ts
```diff
@@ -688,12 +692,32 @@ latestBlock timestamp   | ${latestBlock.timestamp}
...
+    const managerBalanceNotReady = await this._refreshAndCheckBalances()
+    if (managerBalanceNotReady) {
       return []
     }
-    return await this._handleChanges(block, currentBlockTimestamp)
+    return await this._handleChanges(block)
+  }
+
+  async _refreshAndCheckBalances (): Promise<boolean> {
+    const minBalanceToNotReadyFactor = 2
```
forshtat, line 703: Arbitrary hard-coded constant.

Also, why would server that is registered and staked be 'not ready' if its manager ran out of ether, if it has a worker that is fully funded? It can still mine a lot of transactions before there is any issue, right?

packages/relay/src/RelayServer.ts
```diff
@@ -688,12 +692,32 @@ latestBlock timestamp   | ${latestBlock.timestamp}
...
+      await this.registrationManager.refreshBalance()
+      this.workerBalanceRequired.currentValue = await this.getWorkerBalance(0)
+
+      managerBalanceNotReady = this.registrationManager.balanceRequired.currentValue.lt(toBN(this.config.managerMinBalance.toString()).divn(minBalanceToNotReadyFactor))
+      workerBalanceNotReady = this.workerBalanceRequired.currentValue.lt(toBN(this.config.workerMinBalance.toString()).divn(minBalanceToNotReadyFactor))
+      if (managerBalanceNotReady || workerBalanceNotReady) {
+        this.logger.debug(`${managerBalanceNotReady ? 'manager' : 'worker'} balance too low`)
+        this.setReadyState(false)
+      }
+      const managerBalanceAwaitsReplenish = this.registrationManager.balanceRequired.currentValue.lt(toBN(this.config.managerMinBalance.toString()))
```
forshtat, line 716: "awaits" assumes there is a pending transaction and we are 'waiting' for it to resolve.

Please rename to 'requires'.

packages/relay/src/TransactionManager.ts
```diff
@@ -132,12 +134,13 @@ data                     | ${transaction.data}
...
   }
 
-  async broadcastTransaction (signedTx: string, verifiedTxId: string, nonce: number): Promise<SignedTransactionDetails> {
+  async broadcastTransaction (signedTx: string, verifiedTxId: string, nonce: number, signer: string): Promise<SignedTransactionDetails> {
     try {
       const transactionHash = await this.contractInteractor.broadcastTransaction(signedTx)
       if (transactionHash.toLowerCase() !== verifiedTxId.toLowerCase()) {
         throw new Error(`txhash mismatch: from receipt: ${transactionHash} from txstore:${verifiedTxId}`)
       }
+      this.emit('TransactionBroadcast')
```
forshtat, line 143: Is this the only reason the `TransactionManager` we have had for years now has to be an `EventEmitter`? 
A single event that has to be handled by a single class?

I don't care that much about this server's codebase as it needs a complete rewrite at this point, but still, I think introducing event-driven code like this is abuse of the concept.

Maybe it is better to accept a single optional `onTransactionListener` variable in the constructor?

packages/relay/src/TransactionManager.ts
```diff
@@ -132,12 +134,13 @@ data                     | ${transaction.data}
     }
   }
 
-  async broadcastTransaction (signedTx: string, verifiedTxId: string, nonce: number): Promise<SignedTransactionDetails> {
+  async broadcastTransaction (signedTx: string, verifiedTxId: string, nonce: number, signer: string): Promise<SignedTransactionDetails> {
```
forshtat, line 137: What is the use of this new `signer` parameter?

forshtat (changes requested):

packages/relay/src/RelayServer.ts
forshtat: Wait, you solved it by removing the log? I like the log. Keep the log.

forshtat (approved):

PR #899, opened 2022-09-11
OG-919: Update issue templates

forshtat:


drortirosh (approved):
approved, though this is a generic template. we could ask for something specific, like browser console logs.
also, can suggest using discord, or lookin at the docs.opengsn.org/faq first..

PR #903, opened 2022-09-16
GSN v3.0.0-beta.2

forshtat:


drortirosh (approved):

drortirosh (commented):
ok, but I still suggest do initial deployment before merging into master (so any modifications, even to the deployment scripts, can be done on this branch.

drortirosh (approved):

PR #904, opened 2022-09-16
OG-932: Fix private preferred relay event lookup

forshtat:


packages/deployer/src/deployUtils.ts
```diff
@@ -161,7 +161,8 @@ export async function printRelayInfo (hre: HardhatRuntimeEnvironment): Promise<v
     maxFeePerGas: 1e12,
     ethereumNodeUrl: network.url
   }, null, 2)))
-  console.log(chalk.white('Relayer register:'))
+  console.log(chalk.white('Register your Relay Server:'))
+  console.log(chalk.grey('Go to https://relays.opengsn.org/ and register your relay server with the web app, or use the following CLI command:'))
```
drortirosh, line 165: nitpic: the idea here is to save US work, of copying a lot of stuff to a file or command-line. just pointing to a URL doesn't give much help (through it doesn't do any harm either.)

drortirosh (approved):

PR #916, opened 2022-11-25
OG-973: Implement default 'ApprovalDataCallback' for the Verifier Tool

forshtat:


packages/provider/src/RelayClient.ts
```diff
@@ -556,6 +561,16 @@ export class RelayClient {
       this.logger.debug(`Reading default client config for chainId ${chainId.toString()}`)
       configFromServer = await this._resolveConfigurationFromServer(chainId, defaultGsnConfig.clientDefaultConfigUrl)
     }
+    if (config.verifierServerApiKey != null && config.verifierServerApiKey.length !== 0) {
```
shahafn, line 564: Do we have to have two different checks for the server api key here, as well as nested checks? maybe clean it up a bit, move that part to a different function, and early return instead of nested checks?

packages/provider/src/RelayClient.ts
```diff
@@ -579,9 +594,13 @@ export class RelayClient {
 
   async _resolveDependencies ({
     provider,
-    config = {},
+    config,
```
shahafn, line 597: Are we sure that we don't depend anywhere on this default value that you removed?

forshtat: Well, it is unlikely as the default value here is literally an empty object. However, I've changed it to accept the fully initialized `GSNConfig` because the `_resolveDependencies` is only called after the `_resolveConfiguration`.
I don't see how this empty config could ever be used here at all.

shahafn (approved):

PR #918, opened 2022-12-05
OG-974: Query VerifierPaymaster address from the Verifier Server

forshtat:


packages/provider/src/RelayClient.ts
```diff
@@ -570,6 +571,9 @@ export class RelayClient {
       }
       config.verifierServerUrl = config.verifierServerUrl ?? DEFAULT_VERIFIER_SERVER_URL
       this.logger.info(`Verifier server API Key is set - setting verifierServerUrl to ${config.verifierServerUrl}`)
+      if (config.paymasterAddress == null) {
```
shahafn, line 574: We really need to clean this up. We have both token paymaster and verifying paymaster, and we have different mechanism to check which one to use. We should probably make something generic for all of our paymasters services configurations (currently two, more in the future).

PR #939, opened 2023-01-26
TokenPayMaster - update err msg

Alexhandru:
The error message of the require inside "_verifyPaymasterData" function should be updated to specifying V3 of Uniswap.


shahafn:
Thank you!

shahafn (approved):

PR #951, opened 2023-03-13
OG-1017: Relay gets stuck if worker balance is insufficient to boost transactions

forshtat:


packages/relay/src/TransactionManager.ts
```diff
@@ -175,7 +207,14 @@ data                     | ${transaction.data}
       })
       this.logger.debug(`sendTransaction: gasLimit from estimate: ${gasLimit}`)
     }
-    await this.validateBalance(txDetails.signer, maxFeePerGas, gasLimit)
+    const signerBalance = await this.contractInteractor.getBalance(txDetails.signer)
+    const {
+      requiredBalance,
+      isSufficient
+    } = await this.validateBalance(txDetails.signer, maxFeePerGas, gasLimit, signerBalance)
+    if (!isSufficient) {
```
shahafn, line 215: Isn't that a duplicate since you already check it inside `validateBalance()`?

packages/relay/src/TransactionManager.ts
```diff
@@ -260,8 +299,16 @@ data                     | ${transaction.data}
...
-    isMaxGasPriceReached: boolean): Promise<SignedTransactionDetails> {
-    await this.validateBalance(tx.from, newMaxFee, tx.gas)
+    isMaxGasPriceReached: boolean): Promise<
+    {
+      signedTransactionDetails?: SignedTransactionDetails
+      balanceRequiredDetails: BalanceRequiredDetails
+    }> {
+    const signerBalance = await this.contractInteractor.getBalance(tx.from)
+    const balanceRequiredDetails = await this.validateBalance(tx.from, newMaxFee, tx.gas, signerBalance)
+    if (!balanceRequiredDetails.isSufficient) {
```
shahafn, line 309: Again duplicate?

packages/relay/src/TransactionManager.ts
forshtat: No. This is the `sendTransaction` method and it throws on insufficient balance.
The `validateBalance` does not throw any more - I don't want to throw in the `resendTransaction` as this is just the boosting - I want to return.

packages/relay/src/TransactionManager.ts
forshtat: Again, here I don't throw but return a required balance. I will not adjust the replenishing amount, but only print an error if I know for sure the target replenishing amount is not enough to "unstuck" this relay.

shahafn (approved):

